PR #1, opened 2020-04-27
add map controls

piducancore:
The title says it all, I added MapControls:

```js
// import
import { MapControls } from "drei"

// and use
<MapControls />
```


drcmda:
nice, thanks!

neftaly:
Is this the same as OrbitControls? I missing something here?

drcmda:
yes, it's a variant of it. threejs/jsm/.../OrbitControls has two exports, OrbitControls and MapControls

neftaly:
Ah, I didn't notice the different import :) Thanks!

PR #5, opened 2020-04-28
Add Sky shader component

sniok:
Adds sky shader from this [example](https://threejs.org/examples/?q=sky#webgl_shaders_sky). [source](https://github.com/mrdoob/three.js/blob/master/examples/webgl_shaders_sky.html)

[Codesandbox](https://codesandbox.io/s/r3f-sky-ufexd)


drcmda:
nice, thanks! its in

PR #7, opened 2020-05-02
Add <GlitchEffect />

talentlessguy:
I took the GlitchEffect example from r3f readme.

* I used `forwardRef`
* I extended all props from `GlitchPass`, `RenderPass` and `EffectComposer`

Example usage:

```jsx
<GlitchEffect>
  <mesh>
    <boxGeometry args={[1, 1, 1]} attach="geometry" />
    <meshBasicMaterial color="red" attach="material" />
  </mesh>
</GlitchEffect>
```

When it will be merged, I'll be able to make a codesandbox with using an example. Or better to make my own with using a component as a part of code?

P.S. I added `preserveSymlinks: true` to tsconfig so pnpm will work too (I use pnpm because of bad internet, altho I removed my lockfile, so it should be fine)


drcmda:
Im unsure how adding children, like meshes and so on, into the effect composer would even work. the composer itself is not a THREE.Object3D.

The other thing i worry about, effects take over the render loop. They cant be orchestrated. This is already a problem with StandardEffects.

But perhaps we could take your idea and make a EffectComposer class that worries and resize and so on, but then could take children which are effectpasses.

talentlessguy:
@drcmda thanks for your feedback!

* I can remove children so it will become a standalone component (should I do?)
* I'll use postprocessing instead

does this sound right?

drcmda:
i think it would make more sense to create a declarative component abstraction of postprocessing.

talentlessguy:
@drcmda, I'm happy to announce that I made it work!

I made the required changes to readme as well, with the new component

Quick preview:
Peek 2020-05-06 01-24

Everything works fine:

* it doesn't take over the render loop
* uses `postprocessing` library

CodeSandbox is [here](https://codesandbox.io/s/drei-glitch-example-h5ni9?file=/src/App.tsx)

code from `Glitch`:

```ts
import { useMemo, useEffect } from 'react'
import { useFrame, useThree } from 'react-three-fiber'
// @ts-ignore
import { GlitchEffect, RenderPass, EffectComposer, EffectPass, NormalPass } from 'postprocessing'
import { HalfFloatType, Vector2 } from 'three'

type GlitchEffectProps = {
  active?: boolean
  ratio?: number
  duration?: Vector2
  delay?: Vector2
  chromaticAberrationOffset?: Vector2
}

const Glitch = (glitchEffectProps: GlitchEffectProps) => {
  const { gl, scene, camera, size } = useThree()

  const composer = useMemo(() => {
    const effectComposer = new EffectComposer(gl, {
      frameBufferType: HalfFloatType,
    })
    effectComposer.addPass(new RenderPass(scene, camera))

    const normalPass = new NormalPass(scene, camera)

    effectComposer.addPass(normalPass)

    const glitchEffect = new GlitchEffect(glitchEffectProps)

    const glitchPass = new EffectPass(camera, glitchEffect)

    glitchPass.renderToScreen = true

    effectComposer.addPass(glitchPass)

    return effectComposer
  }, [camera, gl, scene, glitchEffectProps])

  useEffect(() => composer.setSize(size.width, size.height), [composer, size])
  return useFrame((_, delta) => composer.render(delta), 1)
}

export default Glitch
```

README.md
```diff
@@ -110,15 +111,27 @@ Adds [ambient-occlusion](https://vanruesc.github.io/postprocessing/public/docs/c
...
+  effects={() => [...fx]} // Define your own: ([smaa, ao, bloom]) => [...effects] (default=undefined)
 />
 ```
 
+##### ⚡️ Glitch
+
+Adds [glitch effect](https://vanruesc.github.io/postprocessing/public/docs/class/src/effects/GlitchEffect.js~GlitchEffect.html) from [postprocessing](https://vanruesc.github.io/postprocessing) library.
+
+```jsx
+<Glitch glitchEffectProps={{ delay: new Vector2(1, 1) }} />
```
joaomilho, line 128: I think this is incorrect. Here's a fix:

```suggestion
<Glitch delay={new THREE.Vector2(1, 1)} />
```

drcmda:
@talentlessguy wouldnt something like this be cool?

```jsx
import { EffectComposer, AmbientOcclusion, Bloom, Glitch } from 'react-postprocessing'

<EffectComposer>
  <AmbientOcclusion samples={4} ... />
  <Bloom luminanceThreshold={0.9} ... />
  <Glitch ... />
</EffectComposer>
```

then we could compose effects which would be awesome. i would remove standardeffects from this lib in that case.

if you like i create the repo on the react-spring org and make you shared author. will help you to spread it on twitter etc. could really be a nice project that i imagine lots of people will use.

talentlessguy:
@drcmda oh yes, react wrapper around postprocessing would be nice, however I have zero experience in writing such

talentlessguy:
@drcmda I like the idea! I could try my skills to learn a wrapper around `postprocessing`. I think I need to read `react-three-fiber` source first and then try to do something similar inside of postprocessing wrapper.

I agree on working on it, at least I can try to make something close :heart: 

I think PR should be kept open until I come up with something working?

drcmda:
no, it will be easier, just wrappers, you wont need a reconciler.

talentlessguy:
@drcmda oh, ok, I think I'll still want to try and see what will be in result :)

gsimone:
Can I close this since now react-postprocessing is a thing? @talentlessguy  @drcmda

PR #9, opened 2020-05-09
Fix three-three-fiber peer dependency typo

AustinBrunkhorst:
It appears `three-three-fiber` was erroneously added as a peer dependency, which produces a warning on `yarn install`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "pretty-quick": "^2.0.1",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
-    "react-three-fiber": "^4.1.1",
     "rimraf": "^3.0.2",
     "rollup": "^2.7.2",
     "rollup-plugin-babel": "^4.4.0",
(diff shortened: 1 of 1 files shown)


drcmda:
it does depend on it though? why would it cause a warning?

AustinBrunkhorst:
> it does depend on it though? why would it cause a warning? https://unpkg.com/browse/drei@0.0.32/dist/index.js

It does depend on `react-three-fiber`; however, the peer dependency declares `three-three-fiber` which appears to be a typo.

drcmda:
Oh, didn't see it like that, could you fix the typo instead? In theory react-three-fiber should be a peer dep

AustinBrunkhorst:
Moved it to peer dependencies. I also updated `react` and `react-dom` as it looks like they were declared in dev dependencies.

Should we also move `three` to peer dependencies? This seems like a good candidate.

drcmda:
i think its fine, r3f has a peerdep on three, it should warn if three is not there. i noticed that TS doesnt compile if react and react-dom is removed from devdeps. i have just fixed the typo and published it. thanks! :-)

PR #14, opened 2020-05-17
Adds useHelper hook

gsimone:


drcmda:
is it possible to express args as useHelper(ref, type, arg1, arg2, arg3, ...) ? or was that the reason why you couldnt figure the ts types out initially?

gsimone:
I could probably do it by abusing any, I also think that's a better API

drcmda:
sure, let's do that and someone will fix it later

gsimone:
Done ✅

drcmda:
nice, thanks, btw, i saw your recent edits and prs, would you like to help maintain with shared author rights?

gsimone:
Sure, I'd be glad to help

PR #20, opened 2020-05-19
Enable adding material as <Text>'s child

skuteli:
Thanks for adding <Text> component to drei :)
I think it's more useful to be able to pass children, so we can pass in material and such, like in https://codesandbox.io/embed/troika-3d-text-via-react-three-fiber-ntfx2?fontsize=14

If you think direct import with extension should be used in such cases and drei's helper is meant for material-less simple text, I will drop.

### DIFF
--- a/src/Text.tsx
+++ b/src/Text.tsx
@@ -17,12 +17,13 @@ declare global {
 }
 
 type Props = JSX.IntrinsicElements['mesh'] & {
-  children: string
+  children: React.ReactElement
   color?: ReactThreeFiber.Color
   fontSize?: number
   maxWidth?: number
   lineHeight?: number
   letterSpacing?: number
+  text?: string
   textAlign?: 'left' | 'right' | 'center' | 'justify'
   font?: string
   anchorX?: number | 'left' | 'center' | 'right'
@@ -34,5 +35,9 @@ type Props = JSX.IntrinsicElements['mesh'] & {
 }
 
 export const Text = forwardRef(({ children, ...props }: Props, ref) => {
-  return <textMeshImpl ref={ref} text={children} {...props} />
+  return (
+    <textMeshImpl ref={ref} {...props}>
+      {children}
+    </textMeshImpl>
+  )
 })


drcmda:
yes, it makes sense. only .... `<Text text="text" />` is a bit weird. 🤪 i will run some tests, maybe it's possible to fish out the text nodes from `children` while rendering out react nodes.

drcmda:
i have something that's working, but i think there's something in troika that mutates the original material, it is not reactive anymore. you can only change color through the textMesh class for instance, if you change it in the material nothing happens. i noticed this in th original troika demo, too: https://codesandbox.io/s/r3f-troika-text-2op0v?file=/src/Text.tsx @lojjic

skuteli:
I was thinking about it, and my assumption was that being able to directly nest text nodes in the component is only beneficial when you can put a structured text inside, like say `<span>`'s or `<a>` in html. 
When the text is just a string most API's tend to go with using a prop, for instance MaterialUI will have `{children}` rendered in paragraphs, but `label` prop on inputs for a label string.
For now IMHO it will encourage people trying something like 
```
<Text color="black">
  black
  <Text color="pink">pink</Text>
</Text>
```
and this obviously won't work.

skuteli:
Sorry, actually it works in your approach with children iteration, meaning children are being displayed. 
But as these are in fact meshes, they will be placed on top of another. 
I think making a full paragraph - like experience can be quite a big work.

lojjic:
> but i think there's something in troika that mutates the original material, it is not reactive anymore. you can only change color through the textMesh class for instance, if you change it in the material directly nothing happens. i noticed this in the original troika demo, too, and i think this must also be the case in vanilla three. same applies to all the other props (roughness, etc), they're all dead.

Ah, yes I see what you mean. The TextMesh takes the user-supplied material instance and replaces it with an upgraded material based on its properties at the time. So yes it makes sense that if the child React element reconciler is maintaining a reference to the original material instance and mutating that, rather than asking for the current material, changes would get lost.

I suppose that makes sense in a React environment where there's an implicit assumption that nothing else would ever be modifying the "dom" behind the scenes. This is breaking that assumption.

I'll have to think about this and see if there's a way I can pick up changes from the original material automatically.

lojjic:
Oof this is a hack but it seems to get around the material replacement issue: https://codesandbox.io/s/r3f-troika-text-vjf7o?file=/src/Text.tsx

Basically applying the material's props to Troika's upgraded material instance as a primitive. There's probably a cleaner approach, my facility with React hooks isn't very good yet.

drcmda:
interesting solution, using primitive and just channeling troikas material! and it works, i'll just merge this approach then. and yes, immutability is an important concept that react assumes.

drcmda:
@skuteli ah, i see, that makes sense. but should we rename the component then, due to the `<Text text=""` thing then?

skuteli:
@drcmda well, logically it should just be a `<TextMesh>`, but this makes it kinda pointless to expose in `drei`, as one can just extend troika component...
I think I will just drop this PR and let's keep it as is.
I will work on some more generic component like a <Paragraph> and a <Span> that will allow nesting <TextMeshes> in such a way that you will get them concatenated with separate materials/styling for children. 
The idea is to set props like `position` only on the `<Paragraph>` and `color` or `font` on both `<Paragraph>` and `<Span>`.

```
<Paragraph position={[0,0,0]}>
  <Span color='pink'>Lorem
     <material/>
  </Span>
  <Span color='blue'>Ipsum</Span>
</Pargraph>
```

drcmda:
@skuteli if that topic interests you, there is pretty big need for responsive layout in webgl. if we have a design system that understand flex + the basic text primitives (b, h1...h6, span, etc), then we could start making ui in webgl. there is yoga https://github.com/facebook/yoga , if you would like to try i could give you a spot on github/react-spring and help spreading the word. for sure this is going to have some influence if it works ...

lojjic:
@drcmda @skuteli This isn't the best place to discuss it but since you've mentioned it... https://troika-examples.netlify.app/#flexbox

Unfortunately this is heavily tied to the Troika framework; it would be hard to extract it to a pure Three.js module like I did for TextMesh because it really needs a good predictable component tree lifecycle. But that's obviously something that React/R3F provides, so perhaps someone could adapt my code without too much effort? I'd be happy to assist.

I apologize for the lack of README docs but you could start here: 
https://github.com/protectwise/troika/blob/master/packages/troika-3d-ui/src/flex-layout/FlexNode.js
https://github.com/protectwise/troika/blob/master/packages/troika-3d-ui/src/facade/UIBlock3DFacade.js

drcmda:
@lojjic maybe we should have a chat or call one day, i would like to hear more about troika. before i started r3f i made something similar for the company i work for, never published it, but threejs, mutation, all this imperative stuff made it so inflexible in the end that i gave up, and at that moment react-reconciler was released. of course it never went as deep and capable as troika, i wouldnt dream of creating my own flexbox impl or sdf fonts, troika seems to have lots of very interesting technologies.

lojjic:
@drcmda You're right, that does sound familiar! :)

I would love to discuss this more in detail, but this feels like a poor place for it. Do you have a better forum we could use?

skuteli:
@drcmda. @lojjic  I agree it's not a best place to discuss, but I have no idea where else to reach you...
I picked up @drcmda suggestion with using facebook yoga and this is what I have for now: https://codesandbox.io/s/yoga-r3f-lgl0j
It's WIP of course, lot's of things to improve.
I wanted to include `<Text/>` from `troika` via `drei`, but I have problems measuring a bounding box for it's geometry - @lojjic maybe you can give me some tip on how to approach this?

And yes, of course I'm all about making UI's in webgl. 
We have wasted way too much time on the DOM....

lojjic:
@skuteli Very cool! Yeah, the difficulty is that Yoga requires a `measureText()` implementation that it can call synchronously, and Troika's text processing is all done asynchronously in a web worker.

But, it just occurred to me that I've [already done the integration work](https://github.com/protectwise/troika/blob/master/packages/troika-3d-ui/src/flex-layout/FlexLayoutProcessor.js) required to run Yoga in the same web worker as Troika's text processing where it can do that text measurement. Right now that code sits alongside a bunch of Troika-specific code for managing the flexbox style tree and applying the results to the threejs object tree in the main thread, but there's no reason I couldn't make the actual call to the worker into a standalone module that you could just call from your code. You'd get the added advantage of it being run in a web worker so it doesn't cause frame drops. Does that sound like it would fit your needs?

I've created https://github.com/protectwise/troika/issues/53 to move the flexbox processing to a standalone module. We can discuss it further there if you like.

drcmda:
@skuteli that looks amazing! do you think this can be packed into a lib somehow? you could have a spot on github/react-spring if you want for more exposure. @lojjic and if you'd like to join or have a part - that would be awesome, too! i can help of course.

skuteli:
@drcmda well that would be great:)

I'm a little afraid, but with a bit of polish, fixing memoization, better typings and some meaningful examples I think it may be packed as a library and successfully reused. 
I just need a bit free time to make it happen. 
Anyway being on react-spring will be awesome, challenging, but motivating.

lojjic:
@skuteli @drcmda Did the flexbox layout effort find a home?

I've created https://github.com/protectwise/troika/tree/master/packages/troika-flex-layout which should help with that effort, if that's still being worked on. This is the Yoga processing part of troika-3d-ui, including text measurement and execution in a web worker, as a standalone utility.

skuteli:
@lojjic TBH I just didn't have time to work on it, but this is going to change soon. 

I didn"t know that you also use `yoga` for calculating layout in `troika` as well. 
I need to think on some reasonable approach that will not duplicate the code and carefully read what you did.

I hope I'll be able to make some progress next week. 
The goal as I understand it:  <Flex> component for `r3f`  enabling flexbox layout for it's children, compatible with your's text implementation, so that one can put <Flex><Text></Flex> and get expected layout.

lojjic:
@skuteli Yeah! I'd mentioned it in a previous comment above but it's easy to miss things buried in places like this. :)

Tell you what, whenever you get back to it, loop me in and I can talk you through my experience implementing this, since I've been through it before. That utility package I linked to should save you a lot of time and let you focus on the style tree management and rendering aspects.

skuteli:
Thanks, I surely will.
I guess most of the work is actually done in your code, i just need to understand it well and provide a component with react API and some meaningful examples.
Any preferred form of contact like a chat or smth?
I'm skuteli@gmail.com on google

drcmda:
@lojjic @skuteli just to let you know, the idea of using flex for 3d content is amazing, we're going to roll with it, currently putting a repo on react-spring. i found a declarative pattern that looks stable to me, and it quite effortlessly would also allow for nested flex, so real layouts are possible: https://codesandbox.io/s/yoga-r3f-forked-24838 (made after Oscars demo above). if you want to join you would be very welcome!

skuteli:
@drcmda 
This is very good, I especially like the use of context, basically I see it as cleaner implementation of what I did initially. I also like the "two components" approach, I was thinking in this direction as well. 
I think of this approach as leveraging React's lifecycle and three.js methods for providing measurement.
Happy to see some of my code in use:) 

What I tried to solve - and as for now failed - was marrying it with troika's text, which uses actual yoga's way of measuring - that is providing a method on the yoga node. Basically @lojjic 's approach is coherent and clean, but I couldn't figure out how to make it work with `useLayoutEffect` - and I didn't want to lose it, because it abstracts away all the measuring of meshes with whatever transformations three supports nicely.
The problem is we will have now two separate instances of yoga, one in service worker via troika and one in the main thread. 

I will be thinking about solving this but this is a hard problem. 

I reviewed a number of libraries using yoga in the meantime: 
- https://github.com/facebook/react-native
- https://github.com/react-figma/react-figma
- https://github.com/diegomura/react-pdf
- https://github.com/vadimdemedes/ink

The thing is, they all work either by providing their own components with yoga implementations that wrap the environment, or they do this on the reconciler level. 
The first is not good  for r3f as it is a great advantage of the library that it just exposes three, not wraps it.
The second - reconciler level - I'm not really sure how to do this and if  it is even a good idea.

drcmda:
you mean troika, the bare sdf text also uses yoga? i thought that was an add-on @lojjic ? in either case, the bundler will only use yoga once, but it's gonna be constructed two times then. generally i'd prefer yoga to be an abstraction on top, i wasn't aware that presentational-type components use it for measuring, too.

lojjic:
Oskar is referring to the [troika-flex-layout](https://github.com/protectwise/troika/tree/master/packages/troika-flex-layout) package I created for his use. It's not that far off from how you're using Yoga directly, but with two main differences:

- It integrates with troika-three-text for measuring intrinsic sizes of text nodes, which is pretty crucial as soon as you start involving text and want things to adapt to it.
- It performs the layout execution asynchronously in a web worker. I've found this critical to avoid frame drops, especially once text measurement is involved. Since I do a lot of WebXR work that's very important to me.

I've done a quick fork trying to use troika-flex-layout in place of direct Yoga calls: https://codesandbox.io/s/yoga-r3f-forked-bjz8m
-- the main change is just building a tree of style objects rather than a tree of Yoga nodes, and handling the different result format after the async call. There's obviously a lot of visual lag currently, but I think that's mostly due to a lot of extra renders happening and can probably be (mostly) optimized away.

I do have some other thoughts on the API, but I can keep those for discussion in the new repo you're setting up. 😃 It may boil down to different use cases; my uses are mostly for building [2D UI with some 3D objects sprinkled in](https://troika-examples.netlify.app/#flexbox), whereas you seem to be focusing on 3D objects.

jmatsushita:
> I think that's mostly due to a lot of extra renders happening and can probably be (mostly) optimized away.

Well that one was a `wat?!`: [useEffect hook with array dependency that has a default value causes render loop](https://github.com/facebook/react/issues/18123) so this sandbox fixes it https://codesandbox.io/s/yoga-r3f-fixed-hnw7n

I also have some questions about r3f troika text and yoga integration, so I'm starting a new issue since this thread might be hard to find.

Issue #18, opened 2020-05-18
Strange behaviour with AO and instancing


giulioz:
I've noticed a strange behaviour when you combine AO from StandardEffects and InstancedGeometry.

If you have any other objects on the scene you can see a "ghost" of the single instance on the center.

Try enabling and disabling the flag on the start of this sandbox:
https://codesandbox.io/s/r3f-instanced-colors-cvbdq?file=/src/index.js

giulioz:
Update:
not sure if related or not, but if you edit the flag with hot reload you may get this error:

![image](https://user-images.githubusercontent.com/1353142/82265256-dc4d7880-9966-11ea-8538-55b5b18fca55.png)

giulioz:
Rendering directly the normal pass gives this result:
![image](https://user-images.githubusercontent.com/1353142/82265378-20407d80-9967-11ea-8653-bccb8e4be53f.png)

It seems like the normal pass it totally ignoring instancing…

giulioz:
Seems like it may be a THREE related issue. It's happening every time you set an overrideMaterial with instancing. Closing for now.

PR #22, opened 2020-05-19
Added AlternativeNormalPass option for StandardEffects

giulioz:
As shown in Issue #18, when you combine instancing and SSAO the rendering breaks, since it uses internally `scene.overrideMaterial`. It's not possible to use a single material for instanced and non instanced objects, since the program attributes are different.

I've noticed that using this approach it's pretty difficult to make a fix, so I've found an hack: you can render a Normal Map, even with instancing, only if you use separate MeshNormalMaterials, traversing the scene tree and temporarily exchanging the materials.

I acknowledge that this can be a corner case or an hack, but it may be useful for someone…


drcmda:
Could you put this one into react-postprocessing instead? I'll remove standardeffects when it's out

giulioz:
I'll open an issue, since it's more complicated that that unfortunately…

PR #27, opened 2020-05-26
Forward transforms to wrapped group

### LINKED ISSUE #21: How to use controlled TransformControls?
I tried something like this:

```jsx
<TransformControls position={props.position} ref={ref}>
	<mesh position={props.position}>
		{/* geometry and material omitted for brevity */}
	</mesh>
</TransformControls>
```

and fire a callback like so:

```js
ref.current.addEventListener('dragging-changed', function(e) {
  props.onChange(ref.current.object.position);
});
```

but when the position is updated and the view re-rendered, the mesh and the UI helper will jump away for the same distance of the last drag. My gut feeling is that there are some internal offset stored by the control which breaks this controlled pattern.

kevinresol:
Fixes #21.

When used, only set transforms on `<TransformControls/>` but not in the wrapped object, e.g.

```jsx
<TransformControls position={props.position}>
	<mesh>
		{/* geometry and material omitted for brevity */}
	</mesh>
</TransformControls>
```

### DIFF
--- a/src/TransformControls.tsx
+++ b/src/TransformControls.tsx
@@ -19,7 +19,7 @@ declare global {
 }
 
 export const TransformControls = forwardRef(
-  ({ children, ...props }: { children: React.ReactElement<Object3D> } & TransformControls, ref) => {
+  ({ children, position, rotation, scale, ...props }: { children: React.ReactElement<Object3D> } & TransformControls, ref) => {
     const controls = useRef<TransformControlsImpl>()
     const group = useRef<Group>()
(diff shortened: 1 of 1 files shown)


drcmda:
i think {...props} should be on the group anyway, then we dont need to destructure at all. could you remove the three transforms and just move ...props to the group? 

what should be destructured and put onto the controls are probably these: https://github.com/mrdoob/three.js/blob/dev/examples/jsm/controls/TransformControls.js#L54-L67

PR #30, opened 2020-05-26
Fixes error in shapes

gsimone:
Shapes don't work atm, since the lowercase b makes the name "boxbufferGeometry" instead of the correct camel name


drcmda:
damn :-S

thanks a lot!

PR #33, opened 2020-06-02
enable optional classname on Stats to allow custom styling

beeglebug:
I recently needed to use Stats, but couldn't have it in the top left corner because of existing UI elements, so I made this modification to allow a className to be added to the container so I could target with css.

The only downside is having to use `important` in the css to actually move the damn thing because of the inline styles, but it still seems to be the simplest solution to add a touch of customisation without a load of custom props.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,7 @@ Adds [stats](https://github.com/mrdoob/stats.js/) to document.body. It takes ove
 ```jsx
 <Stats
   showPanel={0}             // Start-up panel (default=0)
+  className='stats'         // Optional className to add to the stats container dom element
   {...props}                // All stats.js props are valid
 />
 ```
--- a/src/Stats.tsx
+++ b/src/Stats.tsx
@@ -13,7 +13,7 @@ export function Stats({ showPanel = 0, className }: Props): null {
   useEffect(() => {
     stats.showPanel(showPanel)
     document.body.appendChild(stats.dom)
-    stats.dom.classList.add(className)
+    if (className) stats.dom.classList.add(className)
     return () => document.body.removeChild(stats.dom)
   }, [])
   return useFrame((state) => {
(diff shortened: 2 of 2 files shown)


unphased:
Do you think it would be feasible to allow it to be inserted somewhere other than the main body, because I'd like to attach it to the canvas in my app, but its position is going to vary a lot. 

Can definitely use workarounds with transform and things like this, but allowing insertion in an arbitrary location seems more elegant.

beeglebug:
I don't see why a 'domElement' prop to override document.body would hurt,
it would match other Three APIs like Orbit controls etc

unphased:
I'm trying to create a React component that renders a dummy container, uses the original stats.js code, and with effect hooks pins it onto the container. That way we can use it like a proper component, rather than only being able to specify a dom node  that has to already exist as a prop.

drcmda:
sure, make a pr, we could add that prop.

unphased:
I tried a few things and they definitely did not work, but I'm pretty new to the newfangled React and functional components. 

It's not clear to me that it's even supported that a dom element can be specified as a prop. Partly because on initial render, the render is happening without any DOM elements initialized yet... but there are other issues.

beeglebug:
hmmm yeah, it might be a little awkward because you'll need a ref to the dom node, but Stats needs to be inside the Canvas to use useFrame, so you may need to do a bit of juggling to get the ref down from your 2d layer into the 3d bit

gsimone:
```jsx
const node = useRef(document.createElement('div'))

  useEffect(() => {

    node.current.id = 'test'
    document.body.appendChild(node.current)

    return () => document.body.removeChild(node.current)
  }, [])

...

<Stats parent={node} />
```

I changed Stats like this:

```jsx
...

export function Stats({ showPanel = 0, className, parent }: Props): null {
  const [stats] = useState(() => new (StatsImpl as any)())
  useEffect(() => {

    const node = (parent && parent.current) || document.body

    stats.showPanel(showPanel)
    node?.appendChild(stats.dom)

    if (className) stats.dom.classList.add(className)

    return () => node?.removeChild(stats.dom)

  }, [parent])

  ...
}

```

gsimone:
You can see the code on this branch https://github.com/gsimone/drei/tree/custom-stats-parent

PR #47, opened 2020-06-23
[WIP] Deprecate <HTML>

### LINKED ISSUE #43: <HTML> is not PascalCase
`<HTML>` primitives throw the following linting error with default standard/semistandard(/eslint?) config:

```
error  Imported JSX component HTML must be in PascalCase  react/jsx-pascal-case
```

Is it possible to rename `<HTML>` back to `<Dom>` (as was in r3f) or similar?

neftaly:
Closes #43 

I don't know what the idiomatic way of doing this is, sorry! It's just my best guess.


neftaly:
^ I also don't know how to write TS :wink:

src/Html.tsx
```diff
@@ -133,3 +133,8 @@ export const HTML = React.forwardRef(
     return <group {...props} ref={group} />
   }
 )
+
+export const HTML = (props: HtmlProps, ref: React.Ref<HTMLDivElement>) => {
```
neftaly, line 137: Should this have type signatures? If so, should it use the React.forwardRef HOF?

drcmda: type signatures? im not so good with ts, what's that? i mean, it should just behave like the original and have the same types-

neftaly: Sorry, this is my first time writing any TS! I mean, should it look more like this:
```
export const HTML = React.forwardRef(
  (props: HtmlProps, ref: React.Ref<HTMLDivElement>) => {
    console.warn('danger');
    return <Html {...props} ref={ref} />;
  }
);
```
or can I do something like this:
```
const HTML = (...args) => {
  console.warn('danger');
  return <Html {...args} />;
}
```

gsimone: The first one looks good to me, I would consider putting the warning in a useEffect(() => ..., [])

neftaly: I'm also worried `<HTML>` isn't supposed have its own `ref`, in case it's pointing to the wrong place. This is probably a non-issue, my knowledge of the React API is out-of-date.

drcmda:
is this ready to go in?

neftaly:
I think so, `<HTML>` warns as expected, and `<Html>` imports fine.

PR #58, opened 2020-07-13
Add prop for Stats to allow specifying where to add it in the DOM

unphased:
This is useful when the Stats widget is suitable for placing with a canvas that is not a full screen canvas. 

Unsure if the proper type is `HTMLElement` or `Node`, but I have confirmed that this is functional in my app.


unphased:
Huh, realized that the conflict is due to the same thing being already implemented on master.

PR #66, opened 2020-07-23
Added <DeviceOrientationControls/>

ChrisDesigns:
This will provide a component, storybook example, and a README.md update. DeviceOrientationControls is one of five missing controls. It could be useful for those wanting to provide mobile interactions and could reduce redundancy for self-implementations of this control.


gsimone:
Thank you @ChrisDesigns!

PR #69, opened 2020-07-23
Stop the audio on PositionalAudio unmount

johnmarinelli:
Stop audio when the component unmounts.  Addresses issue here: https://github.com/react-spring/drei/issues/68


drcmda:
Oh makes sense! I'm on a little camping trip will publish later 🙂

PR #73, opened 2020-07-26
Add Reflector component

flo-pereira:
Adds Reflector component.

I recreated the threejs exemple to illustrate:

> https://codesandbox.io/s/jolly-benz-pmk9j

> Original: https://threejs.org/examples/#webgl_mirror


gsimone:
Hey, looking great, can you add a story to the storybook and an entry in the readme? The storybook will be previewable here https://deploy-preview-73--drei-storybook.netlify.app/

flo-pereira:
done 👍

drcmda:
this is a nice idea! the only thing i dont understand is the cloneElement just to add attach. the problem with cloning is that it breaks composition:

```jsx
function Foo = () => <planeBufferGeometry args={[2, 5]} />

<Reflector>
  <Foo />
</Reflector>
```

I think if we explain that attach is needed it would be more versatile.

btw attach will fall be automated in v5 anyway, you don't have to write out for materials and geometries.

flo-pereira:
You're right @drcmda, it's fixed ! 👍

PR #76, opened 2020-08-01
Added knobs for sky story

giulioz:
I started adding some knobs to the storybook, starting from Sky component.
This way is easier to understand what the props are for.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,14 +1,24 @@
 {
   "env": {
     "browser": true,
+    "shared-node-browser": true,
     "es6": true,
     "node": true
   },
-  "parser": "babel-eslint",
-  "extends": ["eslint:recommended", "plugin:import/errors", "plugin:import/warnings", "prettier"],
-  "plugins": ["react-hooks", "react"],
+  "extends": [
+    "prettier",
+    "prettier/react",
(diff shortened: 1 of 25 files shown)


.storybook/stories/Sky.stories.js
```diff
@@ -8,13 +10,16 @@ import { Plane } from '../../src/shapes'
...
+  decorators: [withKnobs, (storyFn) => <Setup> {storyFn()}</Setup>],
 }
 
 function SkyScene() {
   return (
     <>
-      <Sky sunPosition={[0, Math.PI, -1]} />
+      <Sky
+        distance={number('distance', 450000)}
+        sunPosition={[number('sunPosX', 0), number('sunPosY', Math.PI), number('sunPosZ', -1)]}
```
gsimone, line 21: Can you use range for the inputs? See the story for MeshWobbleMaterial

giulioz: What could be a reasonable range for a position vector?

giulioz: https://github.com/mrdoob/three.js/blob/master/examples/webgl_shaders_sky.html#L61

In THREE.js official samples is calculated from euler angles, for example, but I don't think this can be a good option, since they are not the real props.

PR #77, opened 2020-08-02
added DeviceOrientationControls to index

ChrisDesigns:
Seems that DeviceOrientationControls is missing and it makes it more complex to use


gsimone:
My bad, didn't catch it

PR #78, opened 2020-08-03
adds a warning on standardEffects mount

gsimone:
This adds a console.warn when mounting StandardEffects, since we plan on deprecating it in favour of react-postprocessing

When merging this, release v0.0.071

### DIFF
--- a/src/StandardEffects.tsx
+++ b/src/StandardEffects.tsx
@@ -107,5 +107,8 @@ export function StandardEffects({
   }, [camera, gl, scene, smaa, ao, bloom, edgeDetection, bloomOpacity])
 
   useEffect(() => void composer.setSize(size.width, size.height), [composer, size])
+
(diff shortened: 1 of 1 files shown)


stephencorwin:
@gsimone I thought having a standard effects component was really nice. Deviations would often occur, but I know that for me -- it was helpful when just getting started with SSAO, bloom, etc.

Is there any harm in having `<StandardEffects />` simply compose `react-postprocessing` for a small subset of effects? This might also prevent the need for a breaking change.

drcmda:
@stephencorwin essentially calling 

```jsx
<EffectComposer><SSAO /><Bloom /></EffectComposer>
```

would get you the same thing. but it would be less magic and users would be pushed a little to look up what these weird terms actually do. before react-pp this would have been an imperative 100 LOC block that's why i thought it would be great to have something quick for demoing stuff. but with react-pp it's not only quick, it's also extensible and versatile.

PR #87, opened 2020-08-17
Added rounded box component and story

giulioz:


drcmda:
could you move it to shapes.tsx? then they're all together.

giulioz:
Yes but `shapes.tsx` contains generated shapes from existing THREE components, I thought it would be a better idea to have special cases in different files, since this component is quite different from others. But no problem to have them in the same file anyways. What do you think?

drcmda:
i think that's fine, it's a shape and a super useful one at that. would be a good fit to put it where the others are. it's all just a flat bundle anyway and everything gets pumped into one big distro file. but this way we have less files on the dev end.

drcmda:
folder would also make sense, maybe even more than files. because we already categorize on the GH front but still we have a bunch of files. if you want, make /shapes and untangle the three shapes, then roundedbox also gets its own file. would you like that better?

gsimone:
Maybe consider changing the name of radius0? It feels weird

drcmda:
dude arent you on holidays or something? be gone and enjoy! 😄

src/shapes/RoundedBox.tsx
```diff
@@ -0,0 +1,49 @@
...
+  shape.absarc(width - radius * 2, height - radius * 2, eps, Math.PI / 2, 0, true)
+  shape.absarc(width - radius * 2, eps, eps, 0, -Math.PI / 2, true)
+  return shape
+}
+
+type Props = {
+  width: number
+  height: number
+  depth: number
+  radius0: number
```
stephencorwin, line 21: `{ width = 1, height = 1, depth = 1, radius0 = 0.1, smoothness = 4, children, ...rest },`

I would be in favor of having sane defaults for these dimensions and make them optional.
Then a user could just `<RoundedBox />`  and get a default box that is `1m` cubed.

Also, +1 to `radius0` needing a rename. It does look a bit awkward ;)

giulioz: Done

PR #100, opened 2020-08-26
Text sync invalidate

### LINKED ISSUE #99: invalidateFrameLoop with drei Text
Hi there, I'm trying to use invalidateFrameLoop with drei Text but as you can see here https://codesandbox.io/s/r3f-invalidate-frameloop-drei-text-x5ndk?file=/src/index.js the Text doesn't display on the first frame. 

Could there be something going on because [`sync()` is asynchronous](https://github.com/protectwise/troika/blob/master/packages/troika-three-text/src/Text.js#L259) and `useLayoutEffect()` expects a synchronous function?

I tried with `troika-three-text` directly https://codesandbox.io/s/troika-3d-text-via-react-three-fiber-invalidateframeloop-c5i7p but the problem is here still, so it might be something more subtle with the asynchronous interaction of react and troika? Is that a use case for `<Suspense>`?

giulioz:
Closes #99

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "drei",
-  "version": "1.3.0",
+  "version": "1.3.1",
   "description": "useful add-ons for react-three-fiber",
   "keywords": [
     "react",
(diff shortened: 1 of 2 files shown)


giulioz:
Not to my knowledge

PR #104, opened 2020-09-02
useProgress

drcmda:
#### ⚡️ `useProgress()` [](https://codesandbox.io/s/cranky-newton-k7f9x)

A convenience hook that wraps `THREE.DefaultLoadingManager`'s progress status.

```jsx
function Loader() {
  const { active, progress, errors, item, loaded, total } = useProgress()
  return <Html center>{progress} % loaded</Html>
}

<Suspense fallback={<Loader />}>
  <GLTFModel />
  <CubeMapEnv />
  <TexturedPlane />
</Suspense>
```

If you don't want your progress component to re-render on all changes you can be specific as to what you need, for instance if the component is supposed to collect errors only:

```jsx
const errors = useProgress(state => state.errors)
```

👉 Note that your loading component does not have to be a suspense fallback. You can use it anywhere, even in your dom tree, for instance for overlays.


gsimone:
I think mentioning that it's zustand under the hood and adding a link would be beneficial (ie. you can do anything you could with a zustand store)

gsimone:
I think you forgot to add a story for the new hook in the storybook

PR #106, opened 2020-09-02
making drei treeshakable

### LINKED ISSUE #72: DeviceOrientationControls is missing invalidate on change
We added DeviceOrientationControls in #66 but THREE.DeviceOrientationControls doesn't dispatch a change event:
https://github.com/mrdoob/three.js/blob/master/examples/jsm/controls/DeviceOrientationControls.js

Same for FlyControls ( being added in #64 )
https://github.com/mrdoob/three.js/blob/master/examples/jsm/controls/FlyControls.js

Compare to OrbitControls 
https://github.com/mrdoob/three.js/blob/b5c272cf408cb33153190fa715d81581bd95ee47/examples/jsm/controls/OrbitControls.js#L266

If we want to keep consistency between all controls, we would need to either open an issue or PR on three.

drcmda:
turns out it can't tree-shake, probably because of the extends which are called in global space.

hence, all extends have to go :-S

- [x] DeviceOrientationControls
- [x] FlyControls
- [x] Line
- [x] MapControls
- [x] MeshDistortMaterial
- [x] MeshWobbleMaterial
- [x] OrbitControls
- [x] Reflector
- [x] Sky
- [x] Stars
- [x] Text
- [x] TrackballControls
- [x] TransformControls

#### Other

- [x] Add invalidation to all controls ( needs three > 120, we are on 119 )
I added a `?.` on all addEventListeners to keep compatibility with an three version where controls don't have event listeners. Closes #72 

#### Bonus 

- [x] Added props for sky uniforms 

#### Todo

- [x] Fix types on Wobble
- [x] Fix types on Distort
- [x] Add dispose={null} to all primitives that replaced extends

### DIFF
--- a/.storybook/stories/MeshDistortMaterial.stories.js
+++ b/.storybook/stories/MeshDistortMaterial.stories.js
@@ -1,10 +1,11 @@
-import React from 'react'
+import React, { useRef } from 'react'
 
 import { withKnobs, number } from '@storybook/addon-knobs'
 
 import { Setup } from '../Setup'
 import { MeshDistortMaterial } from '../../src/shaders/MeshDistortMaterial'
 import { Icosahedron } from '../../src/shapes'
+import { useFrame } from 'react-three-fiber'
 
 export default {
   title: 'Shaders/MeshDistortMaterial',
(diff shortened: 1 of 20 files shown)


drcmda:
one thing remains, what about dispose? in v5 primitive will have dispose={null} by default because it shouldnt go about mutating/destroying objects it doesn't own. i think doing `dispose={undefined}` will allow it to dispose again, ... though rn im unsure how that even works.

gsimone:
> one thing remains, what about dispose? in v5 primitives will have dispose={null} by default because it shouldnt go about mutating/destroying objects it doesn't own. doing `dispose={undefined}` will allow it to dispose again, since the renderer specifically looks for `null` to decide if it should bail out. i think we should have that on all primitives that have replaced extended elements.

Added to the todo list 👍

PR #108, opened 2020-09-02
Use function declarations in useTextureLoader

CharlieHess:
Think this fixes the awkwardness when using TypeScript with these loaders.


CharlieHess:
jk, looks like we'd need to fix [`useLoader`](https://github.com/react-spring/react-three-fiber/blob/master/src/hooks.ts#L99-L104) first

PR #112, opened 2020-09-05
Loader / self contained overlay component

drcmda:
I figure that this is something i need in every single project, but im also too lazy to style/animate/transition on my own everytime. I think it would be nice having useProgress for self made loaders, and simple, readymade Loader as something you can just drop in. 

Could someone pls help me with the storybook?

---

#### ⚡️ `<Loader/>`

A quick and easy loading overlay component that you can drop on top of your canvas. It will show an animated loadingbar and a percentage.

```jsx
<Canvas>
  <Suspense fallback={null}>
    <AsyncModels />
  </Suspense>
</Canvas>
<Loader />
```

You can override styles, too.

```jsx
<Loader 
  containerStyles={...container} // Flex laout
  innerStyles={...inner} // inner container
  barStyles={...bar} // the loading bar
  dataStyles={...data} // text
>
```


gsimone:
This could belong in a ‘prototyping’ category, since its something one would need to remake, since it’s so opinionated (colors, styling, markup, ecc )

drcmda:
Makes sense, prototyping could be nice for pick up components that are light on the design side but that can be used for rough drafts. I could think of a couple. I'll add the category.

PR #115, opened 2020-09-07
use-assets hooks

emmelleppi:
Hello! I report here what has already been done in the [library](https://github.com/emmelleppi/use-r3f-assets) that I have published in these days.
Since there is a new prototyping section, it makes sense to have assets-related hooks.

#### 🛠️ `useNormalTexture`  `(needs suspense)`

Loads normal textures from this repository: https://github.com/emmelleppi/normal-maps

```jsx
const [normalMap, url] = useNormalTexture(
  1, // index of the normal texture - https://github.com/emmelleppi/normal-maps/blob/master/normals.json
  // second argument is texture attributes
  {
    offset: [0, 0],
    repeat: [normRepeat, normRepeat],
    anisotropy: 8
  }
)

return (
  ...
  <meshStandardMaterial normalMap={normalMap} />
  ...
)

```

#### ⚒️ `useMatcapTexture` `(needs suspense)`

Loads matcap textures from this repository: https://github.com/emmelleppi/matcaps

(It is a fork of this repository: https://github.com/nidorx/matcaps)

```jsx
const [matcap, url] = useMatcapTexture(
 0, // index of the matcap texture https://github.com/emmelleppi/matcaps/blob/master/matcap-list.json
 1024 // size of the texture ( 64, 128, 256, 512, 1024 )
)

return (
 ...
 <meshMatcapMaterial matcap={matcap} />
 ...
)
```

👉 You can also use the exact name of the matcap texture, like so:

```jsx
const [matcap] = useMatcapTexture("3E2335_D36A1B_8E4A2E_2842A5");
```

👉 Use the `url` to download the texture when you are ready for production!

Last thing: right now the hooks fetch the data from my repos. It would make sense that those repos would be forked by react-spring, once done we can change the url with the new ones ✌🏼


drcmda:
great! whats still missing is referencing them in index.tsx and the github docs

emmelleppi:
ok, it should be fine 👍

Issue #103, opened 2020-09-01
Request: show example of dashed lines using Line abstraction


RobRendell:
I'm having no luck getting the drei Line abstraction to draw with dashed lines.  I'd love for the example codeSandBox to demo the effect in action, so I can see what I'm doing wrong.

Looking at the source, I can see that drei's version is calling computeLineDistances(), so setting `dashed` to true and `dashScale`, `dashSize` and `gapSize` to sensible values should give the dashed line effect.

(Of course, what those sensible values might be isn't obvious - THREE's doco for the LineDashedMaterial is pretty unhelpful.)

In any case, I'm unable to get the effect to show, either in my own code or by editing the Line example codeSandBox.

giulioz:
It's because it defaults to LineMaterial, not to LineDashedMaterial. I'll try to give the option to change the materials in a next version.

RobRendell:
@giulioz Ah, interesting!  Thanks! So it's a flaw in the current implementation, not in my usage of it :)

If I may suggest something, it seems to me that the existing `dashed` boolean prop that the Line component already takes should select between `LineMaterial` and `LineDashedMaterial`.

RobRendell:
Actually, doing some more reading about this, the `LineMaterial` in Drei's Line abstraction is not the base `THREE.LineMaterial`, it's the one in `three/examples/jsm/lines/LineMaterial`, and it does support `dashed`, as long as you do the hacks outlined here: https://github.com/mrdoob/three.js/blob/22ed6755399fa180ede84bf18ff6cea0ad66f6c0/examples/webgl_lines_fat.html#L236-L239

I've raised a PR implementing this change, and an update to the Line storyboard which shows it in action.

RobRendell:
With the above PR merged, there is now a working example of a dashed Line abstraction in the Storyboard: https://drei.react-spring.io/?path=/story/abstractions-line--vertex-colors-line

Thanks!  Closing this issue.

jonybekov:
Try this
```typescript
import { Line } from "@react-three/drei"

function App() {
   return (
     <Canvas>
           <Line
                lineWidth={0.5}
                worldUnits
                **depthWrite
                dashed**
                dashScale={10}
                points={points}
                color={color}
              />     
     </Canvas>
                    
}
```

PR #116, opened 2020-09-08
Support the `dashed` prop in Line abstraction (#103)

RobRendell:
This PR adds support for the `dashed` prop for Drei's Line abstraction, using the advice from the below comment on the Three `webgl_lines_fat` example.

https://github.com/mrdoob/three.js/blob/22ed6755399fa180ede84bf18ff6cea0ad66f6c0/examples/webgl_lines_fat.html#L236-L239

### DIFF
--- a/.storybook/stories/Line.stories.js
+++ b/.storybook/stories/Line.stories.js
@@ -2,7 +2,7 @@ import React from 'react'
 import { GeometryUtils } from 'three/examples/jsm/utils/GeometryUtils'
 import { Vector3 } from 'three'
 
-import { withKnobs, number, color } from '@storybook/addon-knobs'
+import { withKnobs, number, color, boolean } from '@storybook/addon-knobs'
 
 import { Setup } from '../Setup'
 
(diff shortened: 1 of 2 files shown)


drcmda:
great! could you make that useeffect a layouteffect?

drcmda:
thinking of it, doesnt this work?

```jsx
      <primitive dispose={null} object={lineGeometry} attach="geometry" />
        vertexColors={Boolean(vertexColors)}
        resolution={resolution}
        linewidth={lineWidth}
        dashed={dashed}
        defines-USE_DASH={dashed ? "" : undefined}
        {...rest}
      />
```

RobRendell:
Neat - I didn't even know that useLayoutEffect existed.  I'll update the PR.

No, just changing the prop isn't sufficient, because it won't set `lineMaterial.needsUpdate` to `true`.  I expect it would work if `dashed` was constant, but the line material wouldn't update if you changed `dashed` while the `Line` was mounted.

(Also, I don't know what's happening inside the shader, but apparently setting `lineMaterial.defines.USE_DASH` to `undefined` is not sufficient - it actually needs to be deleted)

drcmda:
oh, ok then lets use uLE. to explain, this effect fires after react has rendered, but before threejs had the chance to paint/render it to the screen. so you get references to all created objects, but you can still change stuff before actual render. only caveat is that it must block render, so not so suitable for expensive calculations etc.

Issue #121, opened 2020-09-10
Bug: `dispose` is read-only when using a base material for <Text />


johnmarinelli:
codesandbox: https://codesandbox.io/s/bug-dispose-is-readonly-fg6ib?file=/src/index.js

johnmarinelli:
fixed here: https://github.com/protectwise/troika/commit/4d4bfbc5d4d730eb0098d33beb1c3c562037fddf, closing issue

PR #122, opened 2020-09-10
Fix Bug #121

johnmarinelli:
I'm not sure if you meant to set dispose={null} for any particular reason, but I just removed it.  not sure if this has further implications, if it does, let me know

### DIFF
--- a/src/abstractions/Text.tsx
+++ b/src/abstractions/Text.tsx
@@ -39,7 +39,6 @@ export const Text = forwardRef(({ anchorX = 'center', anchorY = 'middle', childr
         if (baseMtl) {
           n.push(
             <primitive
-              dispose={null}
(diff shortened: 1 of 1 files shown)


gsimone:
Tagging paul for this @drcmda, I'm still not too confident about the state of dispose

lojjic:
The fact that troika is making the material's `dispose` method non-writable is an oversight, and I can fix that. I think that would clear this up, though it's still worth having Paul weigh in on whether it's desirable here.

lojjic:
troika-three-text v0.32.0 should fix the error when trying to overwrite the dispose method.

johnmarinelli:
@lojjic yep looks like it.  i will close this PR then.  Thanks!

PR #128, opened 2020-09-12
Adds useDetectGPU

gsimone:
This is a simple wrapper, with a few ergonomics changes.


gsimone:
> What is the difference between:
> `tier: string | undefined`
> and
> `tier?: string`
> 
> I feel like if something is meant to be optional, we should be using the `?` notation from TS.

```jsx
import { useMemo } from 'react'
import { getGPUTier, IGetGPUTier } from 'detect-gpu'

export function useDetectGPU(
  props: IGetGPUTier
): {
  isDesktop: boolean
  isMobile: boolean
  tier: string
  type: string
} {
  const GPUTier = useMemo(() => {
    const GPUTier = getGPUTier(props)

    return {
      isDesktop: GPUTier.tier.indexOf('DESKTOP') > -1,
      isMobile: GPUTier.tier.indexOf('MOBILE') > -1,
      tier: GPUTier.tier.split('_').pop()!,
      type: GPUTier.type,
    }
  }, [props])

  return GPUTier
}
```

I actually went a different route and used `.pop()!` since the tier string should always be defined anway. What do you think?

PR #130, opened 2020-09-13
Implement loader storybook.

### LINKED ISSUE #118: Chore: Add story for Loader

LinuxIsCool:
#118 Pretty simple implementation. I followed https://codesandbox.io/s/cranky-newton-k7f9x as a reference. I added the thor and midgard serpent scene into .storybook/public. Alternatively, we could simply use suzanne, but I figure it doesn't hurt to have a scene example in the repository for other use cases as well. 

Closes #118


gsimone:
Hey! Thanks for contributing. The PR is almost perfect, I'll add a couple of things to fix later today 👍

.storybook/stories/useLoader.stories.js
```diff
@@ -0,0 +1,43 @@
...
+export default {
+  title: 'Prototyping/useLoader',
+  component: UseLoader,
+}
+
+function Suzanne() {
+  const { nodes } = useGLTFLoader('suzanne.glb', true)
+  return <mesh geometry={nodes.Suzanne.geometry} />
+}
+
```
gsimone, line 17: This can be removed since it's not being used anymore

.storybook/stories/useLoader.stories.js
```diff
@@ -0,0 +1,43 @@
...
+function UseLoader() {
+  return (
+    <>
+      <Loader />
+      <Canvas concurrent camera={{ position: [0, 15, 1000], fov: 70 }}>
+        <fog attach="fog" args={[0xfff0ea, 10, 60]} />
+        <ambientLight intensity={6} />
+        <Suspense fallback={null}>
+          <ZoomIn />
+          {/* <Suzanne /> */}
```
gsimone, line 32: This can be removed

.storybook/stories/useLoader.stories.js
```diff
@@ -0,0 +1,43 @@
...
+
+function ZoomIn() {
+  const vec = new Vector3(0, 10, 30)
+  return useFrame(({ camera }) => camera.position.lerp(vec, 0.1))
+}
+
+function UseLoader() {
+  return (
+    <>
+      <Loader />
```
gsimone, line 26: This should be put as a fallback for the <Suspense> block, check this example:

https://codesandbox.io/s/r3f-whats-new-in-drei-9ziry

The way you proposed, the loader would always be visible, even after we loaded everything

gsimone (changes requested):
Here's a couple of changes, let me know if you can't get to them, I can fix them later! 

Thanks again for contributing 🚀

.storybook/stories/useLoader.stories.js
LinuxIsCool: Thanks for the feedback gsimone. I made the changes as you suggested, but I do not see the loader when I refresh the scene now. Can you try it and let me know? Thanks!

PR #131, opened 2020-09-15
Added useFBXLoader wrapper similar to useGLTFLoader.

emileswain:
Updated storybook
Updated README.md


.storybook/stories/useFBXLoader.stories.js
```diff
@@ -0,0 +1,41 @@
+import React, { Suspense } from 'react'
+
+import { Setup } from '../Setup'
+
+import {useFBXLoader} from "../../src/loaders/useFBXLoader";
```
gsimone, line 5: Looks like there are some style issues here, can you make sure the code is ran through prettier?

.storybook/stories/useFBXLoader.stories.js
```diff
@@ -0,0 +1,41 @@
...
+export default {
+  title: 'Loaders/FBX',
+  component: useFBXLoader,
+  decorators: [(storyFn) => <Setup cameraPosition={[0, 0, 5]}>{storyFn()}</Setup>],
+}
+
+function SuzanneFBX() {
+  let  fbx = useFBXLoader('suzanne/suzanne.fbx')
+
+  const envMap = useCubeTextureLoader(['px.png', 'nx.png', 'py.png', 'ny.png', 'pz.png', 'nz.png'], { path: 'cube/' })
```
gsimone, line 18: I would remove this envMap to keep the story simple

.storybook/stories/useFBXLoader.stories.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const mat = fbx.children[0].material;
+  mat.envMap = envMap;
+  mat.reflectivity = 0.6
+  mat.normalScale = new Vector2(0.1,0.1)
+  fbx.children[0].material = mat;
+  // console.log(fbx);
+
+  return <primitive object={fbx} dispose={null} />
```
gsimone, line 27: I would also remove this material changes because we usually want to avoid side-effects in the body of the component

You are also re-assigning an object to itself, which shouldn't be necessary

gsimone (changes requested):

gsimone:
I review and asked for some minor changes, let me know if you can't do them, I can take care of it later. Thank you so much for the contrib!

.storybook/stories/useFBXLoader.stories.js
emileswain: Sure, i tend to prefere more helpfull examples, but i'll remove this. Apparently you can't edit fbx.children[0].material.envMap = envMap; Its read only. Hence you have to reassign it.

.storybook/stories/useFBXLoader.stories.js
emileswain: sure will do

emileswain:
> I review and asked for some minor changes, let me know if you can't do them, I can take care of it later. Thank you so much for the contrib!

One thing i did ponder was if to return the <primative> by default, with perhaps Destructuring [primative, fbx] or something like that.

.storybook/stories/useFBXLoader.stories.js
gsimone: It's a good point about the helpfulness 🤔

I'm going to think about a way to keep this in a way that feels better

emileswain: i've removed it just looking to update the pullrequest.

gsimone (approved):

PR #135, opened 2020-09-17
Adding ability to specify a target three object to attach html to in a prop

unphased:
I'm working with some three.js code which is not R3F compatible. However it is straightforward to obtain a handle to an object, so these are some changes I added to enable that. 

I also changed a few other things to satisfy lint rules from my project (such as `() => void (side effect)` -> `() => { side effect; }`)

### DIFF
--- a/src/misc/Html.tsx
+++ b/src/misc/Html.tsx
@@ -31,7 +31,9 @@ function objectScale(el: Object3D, camera: Camera) {
     const dist = objectPos.distanceTo(cameraPos)
     return 1 / (2 * Math.tan(vFOV / 2) * dist)
   }
-  if (camera instanceof OrthographicCamera) return camera.zoom
+  if (camera instanceof OrthographicCamera) {
+    return camera.zoom
+  }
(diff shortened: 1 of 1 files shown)


drcmda:
tbh it seems kind of hacky that way. so you're injecting a reference to an outside object in order to trick it. but it will still load the entire r3f into your project.

my take is that we need a common ground, which react provides in order to write re-usable components. that assumes that everything adheres to the same rules.

it would be better if you fork this file and adapt it to your own needs.

unphased:
Yep, definitely hacky.

We're stuck with using https://github.com/gkjohnson/urdf-loaders for now so I cannot think of a better way to deal with this. I just copied this forked file over to PR it and get your opinion on it. Thanks for the feedback!

giulioz:
Are you sure that the library is completely incompatible with r3f? it seems a model loader to me…

gsimone:
Closing this since it's inactive since Sep, please let me know if you want to keep working on it

PR #137, opened 2020-09-21
ContextBridge

stephencorwin:
Related: https://github.com/pmndrs/jotai/pull/57

This may not work specifically with how `jotai` uses contexts. However, this is a more general use-case Context Bridge that is built to support the standard `React.createContext` api.

Currently, folks have troubles whenever they need to consume context from within a `<Canvas />`. They must go through a few hoops to forward the context into the canvas. This is particularly cumbersome when dealing with multiple contexts -- like most applications use.

This bridge works by reducing an array of contexts that are passed as a prop so that they become nested within each other. The output mimics the typical pattern of:
```tsx
<Context1.Provider>
  <Context2.Provider>
    <Context3.Provider>
      // ....
    </Context3.Provider>
  </Context2.Provider>
</Context1.Provider>
```

In the gif below, you can see that I am changing a stateful value that is being forwarded as a context into the `<Canvas />`. I am also consuming a theme context for the colors.
context-bridge
The output in this case mimics this:
```tsx
return (
  <Canvas>
    <ThemeContext.Provider value={React.useContext(ThemeContext)}>
      <GreetingContext.Provider value={React.useContext(GreetingContext)}>
        <Scene />
      </GreetingContext.Provider>
    </ThemeContext.Provider>
  </Canvas>
)
```
but is written in a more composable/flattened way. This is particularly useful when there are 2 or more contexts being forwarded. Contexts are wrapped/parented in the order that they are passed in.
```tsx
return (
  <Canvas>
    <ContextBridge contexts={[
      {context: ThemeContext, value: React.useContext(ThemeContext)},
      {context: GreetingContext, value: React.useContext(GreetingContext)},
    ]}>
      <Scene />
    </ContextBridge>
  </Canvas>
)
```

### DIFF
--- a/.storybook/stories/ContextBridge.stories.js
+++ b/.storybook/stories/ContextBridge.stories.js
@@ -1,0 +1,81 @@
+import React from 'react'
+import {Canvas} from 'react-three-fiber'
+import { withKnobs, text } from '@storybook/addon-knobs'
+
+import { OrbitControls } from '../../src/controls/OrbitControls'
+import { Box } from '../../src/shapes/generated'
+import { ContextBridge } from '../../src/misc/ContextBridge'
+import { Text } from '../../src/abstractions/Text'
+
+export default {
+  title: 'Misc/ContextBridge',
+  component: ContextBridge,
+  decorators: [
+    (storyFn) => storyFn(),
+    withKnobs,
+  ],
+}
+
+const GreetingContext = React.createContext({})
+const ThemeContext = React.createContext({})
+
+function Scene() {
+  // we can now use the context within the canvas via the regular hook
+  const theme = React.useContext(ThemeContext)
+  const greeting = React.useContext(GreetingContext)
+  return <>
+    <Box
+      position-x={-4}
+      args={[3, 2]}
+      material-color={theme.colors.red}
+    />
+    <Box
+      position-x={0}
+      args={[3, 2]}
+      material-color={theme.colors.green}
+    />
+    <Box
+      position-x={4}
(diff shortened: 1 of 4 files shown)


stephencorwin:
This is really just a bit of sugar to make things a bit more readable. However, I think that by simply having a storybook story for the forwarding context pattern, it will help reinforce how to use it.

gsimone (approved):

stephencorwin:
I'm going to follow-up with another PR to enhance the story a bit since I heard that knobs is being discontinued. Merging for now though since the implementation won't change.

drcmda:
i dont understand this bit

```jsx
<ContextBridge contexts={[
      {context: ThemeContext, value: React.useContext(ThemeContext)},
      {context: GreetingContext, value: React.useContext(GreetingContext)},
    ]}>
```

object notation forces you to remember api, but here there's also duplication. a bridge is an unfortunate thing that really the react should solve, but until then i think we should reduce the boilerplate the most we possibly can. 

this for instance could work:

```jsx
import { useBridgeContext } from "drei"

const Bridge = useBridgeContext(ThemeContext, GreetingContext)

<Bridge>
  ...
```

dai-shi:
> ```ts
> const Bridge = useContextBridge(ThemeContext, GreetingContext)
> ```

AFAIU, this only works as expected, if `Bridge` returned by the hook is referentially stable when it rerenders. Otherwise, react will re-mount the entire tree. (It's possible to do so in drei, but not in jotai.)

PR #138, opened 2020-09-21
Detach Transform Controls to clean up old event listeners

notrabs:
- Removes the unused `Props` type. (It is already derived directly from the three type)
- Fixes events being triggered by removed TransformControl children, which appeared for me in a conditionally rendered TransformControl after toggling it several times.

Excerpt of the application code triggering the errors:
```
{isSelected && (
  <TransformControls ref={controlsRef}>
    <group ref={innerGroupRef} />
  </TransformControls>
)}
```


notrabs:
Nvm. It now randomly detaches sometimes :/

PR #141, opened 2020-09-22
Sky component prop changes

jbixon13:
- Add inclination & azimuth props to be referenced in phi & theta constants as in [THREE.js example](https://github.com/mrdoob/three.js/blob/696d7836d1fc56c4702a475e6991c4adef7357f4/examples/webgl_shaders_sky.html#L58-L63)

- Split `sunPosition` prop into `sunPositionX`, `sunPositionY`, `sunPositionZ` props as suggested by @giulioz 
	- Currently commented out references to sunPosition prop, will remove if this is the approach decided upon.

- Changed second Story to include knobs for `inclination` and `azimuth`


src/shaders/Sky.tsx
```diff
@@ -43,7 +61,9 @@ export const Sky = forwardRef<Sky>(
         material-uniforms-mieCoefficient-value={mieCoefficient}
         material-uniforms-mieDirectionalG-value={mieDirectionalG}
         material-uniforms-rayleigh-value={rayleigh}
-        material-uniforms-sunPosition-value={sunPosition}
+        material-uniforms-sunPosition-value={[sunPositionX, sunPositionY, sunPositionZ]}
```
giulioz, line 64: Can this work without modifiyng the shader?

src/shaders/Sky.tsx
```diff
@@ -1,11 +1,16 @@
 import React, { forwardRef, useMemo } from 'react'
-import { ReactThreeFiber, extend } from 'react-three-fiber'
+import { ReactThreeFiber } from 'react-three-fiber'
 import { Sky as SkyImpl } from 'three/examples/jsm/objects/Sky'
 import { Vector3 } from 'three'
 
 export type Sky = {
   distance?: number
-  sunPosition?: ReactThreeFiber.Vector3
+  // sunPosition?: ReactThreeFiber.Vector3
```
giulioz, line 8: why splitting it?

jbixon13: I split it based on your message in the Discord channel: 

"to be honest I would separate the three angles in three props"

Let me know if you envisioned something else; this implementation didn't make much sense to me since I couldn't see a scenario in which an end-user would want to declare the three angles separately.

src/shaders/Sky.tsx
jbixon13: Do you mean by referencing a single sunPosition prop as done previously but that combines sunPosition X, Y, Z at an earlier point?

src/shaders/Sky.tsx
giulioz: Yeah I meant three different angles, not three different positions tbh, otherwise it's better using a `ReactThreeFiber.Vector3`.

Either way it may be a bad idea to keep both angles and position, which should have priority?

src/shaders/Sky.tsx
giulioz: You are setting `inclination` and `azimuth` material uniforms, which does not exists in the shader

src/shaders/Sky.tsx
jbixon13: As someone who recently tried it for the first time it would make much more sense to me to just expose `inclination` and `azimuth` to the end user and to handle the angle definitions internally without exposing that logic. 

As long as it mimics how the angles are defined in the THREE example, that should be enough to let the end-user do what they want while retaining parity with the source example.

giulioz: Ok so we could remove sunPosition entirely 😛

src/shaders/Sky.tsx
```diff
@@ -1,11 +1,16 @@
 import React, { forwardRef, useMemo } from 'react'
-import { ReactThreeFiber, extend } from 'react-three-fiber'
+import { ReactThreeFiber } from 'react-three-fiber'
 import { Sky as SkyImpl } from 'three/examples/jsm/objects/Sky'
 import { Vector3 } from 'three'
 
 export type Sky = {
```
giulioz, line 6: Wait isn't this entirely useless?

src/shaders/Sky.tsx
jbixon13: Awesome, I'll have that changed later today

src/shaders/Sky.tsx
giulioz: I mean, defining the component props and JSX declaration, we are using a `<primitive>` after all

src/shaders/Sky.tsx
jbixon13: Ah I must have misunderstood how that works, I'll have that changed as well

src/shaders/Sky.tsx
giulioz: @gsimone am I insane? lol

jbixon13: I don't think I'm at all qualified to answer that haha, I've just started playing around with THREE and r3f in the last week. What specifically are you referring to by "this"? The fact that I removed extend? Or importing `ReactThreeFiber`at all?

jbixon13:
@giulioz @gsimone are there any adjustments needed in order to merge?

geonsonroldan:
No

giulioz:
Really sorry! Merging right now

PR #148, opened 2020-09-30
Replace star exports with named exports to fix tree-shaking

### LINKED ISSUE #147: Tree-shaking not working
I've been trying to tree-shake ThreeJS in my project, since I have a bundle that weighs 1.7 MB... I've tried using [three-minifier](https://github.com/yushijinhun/three-minifier), as well as creating my own export file as suggested in a [react-three-fiber recipe](https://github.com/pmndrs/react-three-fiber/blob/master/recipes.md#reducing-bundle-size), but neither solution worked.

Then, I tried removing `drei`, which I use for the `Html` component, and **my bundle dropped by 800 kB**!

I see that a previous attempt was made at fixing this in #106, but it did not seem to work (I'm on `drei` v1.5.6). 

I think the issue is caused by the `export *` syntax used in https://github.com/pmndrs/drei/blob/master/src/index.tsx (cf. this [webpack issue](https://github.com/webpack/webpack/issues/2847)). Fortunately, it looks like it's been fixed in webpack v5.

It may be worth replacing `export *` with named exports, since broad adoption of webpack v5 in the ecosystem may take a while.

axelboc:
Fix #147

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,53 +1,75 @@
-export * from './abstractions/Billboard'
-export * from './abstractions/Detailed'
-export * from './abstractions/Line'
-export * from './abstractions/PositionalAudio'
-export * from './abstractions/Text'
-export * from './abstractions/Environment'
+export { Billboard } from './abstractions/Billboard'
+export type { BillboardProps } from './abstractions/Billboard'
+export { Detailed } from './abstractions/Detailed'
+export { Line } from './abstractions/Line'
+export { PositionalAudio } from './abstractions/PositionalAudio'
+export { Text } from './abstractions/Text'
+export { Environment } from './abstractions/Environment'
 
-export * from './cameras/OrthographicCamera'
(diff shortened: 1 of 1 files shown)


gsimone:
Did you verify that this fixes tree-shaking? Can you attach something? ( a bundle analyzer screenshot or equivalent )

axelboc:
I tried to test it with `npm link` but I ended up with two versions of ThreeJS in my bundle because of the way `npm link` works... And I couln't put an alias in webpack because it would have interfered with `three-minifer`... Not sure what else to try. 😓

axelboc:
I was able to work around NPM's symlinking behaviour by running `npm pack` in drei's `dist` folder and installing the resulting tarball in my project.

Unfortunately, I'm not seeing any improvement, drei still adds the same amount of weight to my bundle (roughly 500 kB). My guess is my changes are helping, but they're not enough.

Issue #89, opened 2020-08-18
Distort and Wobble material constructor problem


acolbourn:
Hi,

I discovered a bug that should be a quick fix.  In the main MeshWobbleMaterial example (link: https://codesandbox.io/s/r3f-sky-g5373), with versions past 0.0.67 the following error occurs:

Class constructor WobbleMaterialImpl cannot be invoked without 'new'

This error can be easily reproduced in that codesandbox by switching drei to version 0.0.68

I believe this is probably caused by a change in the constructor when MeshDistortMaterial was added as that has the same error and I think was possibly added at that version.

Thanks and let me know if you need any more info.

robksawyer:
I'm running into the same issue.

See https://codesandbox.io/s/bold-http-i0lmf?file=/src/App.js

stephencorwin:
@acolbourn 
It seems to be an issue that only occurs when using these materials in combination with the `<Text />` component.
https://codesandbox.io/s/r3f-drei-meshwobblematerial-forked-2r443?file=/src/App.js

This is running 1.2.0 and by removing the material from `<Text />`, we still have the wobble when clicking on the `<Donut />` or `<Image />`.

stephencorwin:
Possibly related: https://github.com/react-spring/drei/pull/20

lojjic:
This appears to be a shortcoming of how `troika-three-text` extends the material; it assumes the base material's class was a classic JS constructor function, not a ES2015 class.

I've created https://github.com/protectwise/troika/issues/79 to see if I can fix this on my side.

lojjic:
This should be fixed by the dependency update in #150

johnmarinelli:
this is fixed as of drei@2.1.0, this issue can be closed

acolbourn:
Awesome, thank you!

PR #150, opened 2020-10-02
Update to troika-three-text 0.33 and remove the forked r3f-troika

lojjic:
Included in this update:
- The derived material now automatically picks up changes to the base material; the hacky code for manually forwarding props is no longer needed and has been removed.
- Using materials defined as ES6 classes no longer errors; see #89.
- Text and all its dependencies are now tree-shakeable.


drcmda:
incredible! thanks!

PR #152, opened 2020-10-02
Issue #120: Add story for PointerLockControls

johnmarinelli:
i noticed that the TrackballControls story is actually using `<OrbitControls/>`.  I think this is because `<Setup/>` needs `controls={false}` for any *Controls story, otherwise OrbitControls will override any other *Controls.

### DIFF
--- a/.storybook/stories/PointerLockControls.stories.js
+++ b/.storybook/stories/PointerLockControls.stories.js
@@ -1,0 +1,51 @@
+import React, { useMemo } from 'react'
+
+import { PointerLockControls } from '../../src/PointerLockControls'
+import { Setup } from '../Setup'
+import { Icosahedron } from '../../src/shapes'
+
+export default {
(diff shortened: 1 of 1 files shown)


.storybook/stories/PointerLockControls.stories.js
```diff
@@ -0,0 +1,51 @@
+import React, { useMemo } from 'react'
+
+import { PointerLockControls } from '../../src/PointerLockControls'
+import { Setup } from '../Setup'
+import { Icosahedron } from '../../src/shapes'
+
+export default {
+  title: 'Controls/PointerLockControls',
+  component: PointerLockControlsScene,
+  decorators: [(storyFn) => <Setup cameraPosition={[0, 0, 10]} controls={false}>{storyFn()}</Setup>],
```
johnmarinelli, line 10: notice `controls={false}` - i think all *Controls stories need this option

Issue #176, opened 2020-11-11
Loading multiple textures with useTexture


Code47X:
I'm using the useTexture hook to load in a collection of textures. Currently I've been loading each texture individually like so, which seems to work fine:

```
const MetalicCube = () => {
  const colorMap = useTexture('/Textures/Metal007_2K_Color.jpg');
  const normalMap = useTexture('/Textures/Metal007_2K_Normal.jpg');
  const metalnessMap = useTexture('/Textures/Metal007_2K_Metalness.jpg');
  const roughnessMap = useTexture('Textures/Metal007_2K_Roughness.jpg');
  const displacementMap = useTexture('Textures/Metal007_2K_Displacement.jpg');

  return (<div />)
}
```
I've checked out the useTexture examples, and TypeScript shows it can **return** an array of textures, but I'm not quite following the syntax. It seems like useTexture will only accept a single string as an argument. TS won't accept an array of strings as an argument. I've tried pointing it to the folder containing the textures, but no luck.

From Example:
`const [texture1, texture2] = useTexture([texture1, texture2])`

Just trying to figure out if there is a better way to load all the textures, or if using useTexture individually for each is the correct way. Thanks.

drcmda:
yes it should work like that. useLoader does for sure. could you look into it and fix it? maybe the types are off.

Code47X:
Checked it out, and I believe the typing is slightly off. It looks like the conditional type would always evaluate to string. I can throw up a quick PR if this is an acceptable fix.

Current:
`useTexture(url: string extends any[] ? string[] : string): Texture | Texture[];`

Fixed for me:
`useTexture(url: string | string[]): Texture | Texture[];`

joshuaellis:
@drcmda when using `useLoader` I get the following error:
```shell
Type 'Texture' must have a '[Symbol.iterator]()' method that returns an iterator.
```

For this code:
```js
 const [matcap1, matcap2] = useLoader(TextureLoader, ['matcap-1.png', 'matcap-2.png'])
```

Using @Code47X's TS example in `useTexture` get's the same error.

Code47X:
@joshuaellis Not sure if this helps you at all, but when I was messing with source code for `useLoader` I had to pass in a generic `useLoader<Texture[]>(TextureLoader, [t1.png, t2.png])` in order to make it work with multiple textures.

joshuaellis:
[...]

PR #177, opened 2020-11-11
useTextures hook for loading multiple textures

Code47X:
Related to #176 

Currently when using TypeScript it doesn't seem possible to load multiple textures with useTexture(), as it won't allow an array of strings as an argument.

Initially I tried to fix `useTexture()` to work with both `string | string[]`, however it uses `useLoader()` under the hood. `useLoader()` works fine, but in order to give an array of strings it requires you pass a generic like so... `useLoader<Texture[]>(TextureLoader, urls)` Since it's a hook I couldn't find a way to conditionally call `useLoader() / useLoader<Texture[]>()` solely within `useTexture()`.

So in the end I just made another function called `useTextures`, identical to `useTexture`, but under the hood it calls `useLoader<Texture[]>`, which allows it to accept an array of strings. Maybe someone with more TS experience can figure out a more elegant solution, but I thought I'd throw up the PR if there was any interest.

### DIFF
--- a/src/useTexture.tsx
+++ b/src/useTexture.tsx
@@ -1,8 +1,13 @@
 import { Texture, TextureLoader } from 'three'
 import { useLoader } from 'react-three-fiber'
 
-export function useTexture(url: string extends any[] ? string[] : string): Texture | Texture[] {
+export function useTexture(url: string): Texture {
   return useLoader(TextureLoader, url)
 }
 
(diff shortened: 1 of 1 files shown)


Code47X:
Closing this PR for now, not happy with this implementation. I'll push up a cleaner fix when I have some free time.

PR #184, opened 2020-11-20
updated dependencies, fixed typescript imports error esmoduleinterop,…

JustFly1984:
… fixed performance, added eslint-plugin-react-perf, fixed eslint warnings and errors


JustFly1984:
@drcmda can you please look at this PR?

gsimone:
There's too much going on in the PR, I'd be happy to review if you could separate each task you took on and documented the what and the why.

I'm also not a fan of using React.useMemo every time you need to pass an object as props

JustFly1984:
There is not actually a lot is going on. As I've described in commit message, there is only plugin-react-perf.

The use of useMemo is required. 2 main changes - fix in imports - in every react file, and performance fix according to Eslint- to optimize performance. The component to which you pass new object/array will be re rendered each time parent re renders, and it hits performance badly.

gsimone:
Also, what's the reason for the change in the imports?

JustFly1984:
I have a project, where I can't use dependencies which has built with `esModuleInterop: true`, cos then I type-check the project including dependencies, I have a bunch of errors with recomendation to switch `esModuleInterop: true` in my project 
https://github.com/apollographql/apollo-server/issues/1977#issuecomment-662946590 this is a link with explanation, though you can read an issue from beginning.
There is a lot of issues like that in different libraries too, but slowly it get fixed.
BTW for node library it still could be ok, but for frontend libraries it is a bummer.

Thank you for your response, I really appreciate it.

PS: the log of errors in typescript from drei:

Screen Shot 2020-11-28 at 12 11 51 AM

JustFly1984:
@gsimone I've done the same fixed to react-three-fiber https://github.com/pmndrs/react-three-fiber/pull/823 and it was merged, and released new version, which resolved all the typescript errors. I see errors only from drei.

.storybook/stories/useCamera.stories.js
```diff
@@ -12,14 +12,14 @@ export default {
...
+  const virtualCam = React.useRef()
+  const ref = React.useRef()
 
-  const [hover, set] = useState(null)
+  const [hover, set] = React.useState(null)
 
   const { gl, scene, camera } = useThree()
 
-  const virtualScene = useMemo(() => new THREE.Scene(), [])
+  const virtualScene = React.seMemo(() => new THREE.Scene(), [])
```
gsimone, line 22: typo

gsimone:
So, I thought about it and I won't merge the PR as is, please limit the changes to the interop fixes, the aggressive use of the perf optimizations is unwarranted in the examples/stories and the specifics are better discussed/worked on in a dedicated PR.

JustFly1984:
@gsimone sorry had been busy with life. I've created new PR https://github.com/pmndrs/drei/pull/198

gsimone:
Closing in favor of #198

PR #185, opened 2020-11-23
Fix types for useAspect

tsumo:
- string literal type for 'type' argument
- 3-number tuple as return type (for compatibility with `<mesh>` scale prop)

### DIFF
--- a/src/useAspect.tsx
+++ b/src/useAspect.tsx
@@ -1,6 +1,6 @@
 import { useThree } from 'react-three-fiber'
 
-export function useAspect(type: string, width: number, height: number, factor: number = 1) {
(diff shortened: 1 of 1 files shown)


src/useAspect.tsx
```diff
@@ -1,6 +1,6 @@
 import { useThree } from 'react-three-fiber'
 
-export function useAspect(type: string, width: number, height: number, factor: number = 1) {
+export function useAspect(type: 'cover', width: number, height: number, factor: number = 1): [number, number, number] {
```
JustFly1984, line 3: if there is only one option for the `type`, should not it be omitted, and used as value?

tsumo: I guess it's a future-proofing for an api, because it should support other css-like modes, though nobody wrote that support yet.

PR #188, opened 2020-12-01
Bump detect-gpu to pick up the SSR fix

### LINKED ISSUE #143: detect-gpu & Vercel: "window is not defined"
Hello! I'm using r3f and drei in my current Next.js-based project which is hosted on Vercel. There is an issue with latest versions of drei. It fails to build with the following error: "ReferenceError: window is not defined". The probem is in "node_modules/detect-gpu" package.

CharlieHess:
`detect-gpu` is currently unusable in SSR setups like Next. `3.0` fixes that: https://github.com/TimvanScherpenzeel/detect-gpu/pull/51. This PR is just upgrading the version drei uses.

Fixes #143.

### DIFF
--- a/.storybook/stories/useDetectGPU.stories.js
+++ b/.storybook/stories/useDetectGPU.stories.js
@@ -14,9 +14,13 @@ export default {
 function Simple() {
   const GPUTier = useDetectGPU()
 
-  return (
+  return GPUTier ? (
     <Text>
-      {GPUTier.tier} {GPUTier.type}
+      Tier {GPUTier.tier.toString()} {GPUTier.type}
+    </Text>
+  ) : (
+    <Text>
+      Detecting GPU …
     </Text>
   )
 }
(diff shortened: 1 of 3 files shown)


gsimone:
The upgrade isn't free, as you can see from the failed deploy. Do you want to take a look?

CharlieHess:
yeah I see now @gsimone; it's async so the only thing I think we can do is return null while we're waiting to hear back. Also felt like we should just return the exact `TierResult` from `detect-gpu` vs manipulating it a bit

gsimone:
Give me a 👍 if I can merge. IMO we can have it like this until someone can find the time to give it a proper go with the new API

CharlieHess:
yep, just needed to fix the storybook. we're all set now 👍

gsimone:
Would you mind opening an issue so we can remember to improve this?

Issue #172, opened 2020-10-28
Add new edgesplit and curve modifiers


gsimone:
New modifiers have been added to three in r122 

https://threejs.org/examples/?q=modifie#webgl_modifier_curve

Can be made similar to the other existing hooks such as https://drei.react-spring.io/?path=/story/modifiers-usesimplification--use-simplification-scene-st

https://github.com/pmndrs/drei/blob/master/src/useTessellation.tsx
https://github.com/pmndrs/drei/blob/master/src/useSimplification.tsx
https://github.com/pmndrs/drei/blob/master/src/useSubdivision.tsx

joshuaellis:
Happy to look at this 👍🏼

gsimone:
Awesome, dm me on twitter/discord or email or whatever if you need support

joshuaellis:
So the `CurveModifier` is a class called `Flow` or `InstancedFlow` it doesn't have modify function, it looks like you have to add the class to the scene – [example](https://github.com/mrdoob/three.js/blob/master/examples/webgl_modifier_curve.html). Because of this, I'm wondering if it wants to be a component like `Box` where the child component is the mesh you want to travel. 

But im unfamiliar with how you regularly handle these things in the library so it'd be good to hear your thoughts.

gsimone:
So, a bunch of thoughts:

Our component should accept:
- a mesh that will be curved
- a curve 

Will return:
- the flow instance accessible via ref

The mesh is not added to the scene, instead it's done later via flow.object3D
https://github.com/mrdoob/three.js/blob/master/examples/webgl_modifier_curve.html#L127

Updating the curve should result in an update to the curved object
https://github.com/mrdoob/three.js/blob/master/examples/webgl_modifier_curve.html#L148

So the component approach would make sense, need to think about an API

joshuaellis:
My initial thought was something like this
```tsx
<Flow curve={new THREE.CatmullRomCurve3(...args)}>
  <Octahedron args={[3, 2]}>
    <meshNormalMaterial attach="material" flatShading color="hotpink" />
  </Octahedron>
</Flow>
```

I imagine this would work with `InstancedFlow` too?

Since this component requires 2 "elements" (the curve & mesh) we could however do a similar approach to how mesh works where you can pass two children, and it understands how to handle the geometry and material, but I don't know enough to know if that's feasible.

joshuaellis:
Hi @gsimone any thoughts on the above implimentation suggestion?

[...]

PR #196, opened 2020-12-08
Feature/drei 172 - useEdge

joshuaellis:
This is one half of #172, useEdge is a hook to be used. However CurveModifier is a bit more complicated and requires a new component to be made. I'm in the process of doing so but in the meantime wanted to share the work i've done for useEdge in case there's any feedback.

I've bumped `three` to 0.122.0 because that's the version that introduced `EdgeSplitModifier` & `CurveModifier`.


src/useEdge.tsx
```diff
@@ -0,0 +1,35 @@
+import * as THREE from 'three'
+import { useEffect, useRef } from 'react'
+import { EdgeSplitModifier } from 'three/examples/jsm/modifiers/EdgeSplitModifier'
+
+export function useEdge(cutOffAngle) {
```
gsimone, line 5: Could you change the name useEdgeSplit to keep consistency with the others?

.storybook/stories/useEdge.stories.js
```diff
@@ -0,0 +1,37 @@
...
+    
+    const meshRef = useEdge(cutoffEdge * Math.PI / 180)
+  
+    return <mesh ref={meshRef} material={materials['Material.001']} geometry={nodes.Suzanne.geometry} />
+  }
+
+function UseEdgeScene() {
+  return (
+    <Suspense fallback={null}>
+      <SuzanneModel />
```
gsimone, line 29: This model doesn't show the modifier well, probably because of the geometry, I think you should look for something that has curves, maybe even a Tube geometry, but really, anything as long as the effect is a bit more obvious

gsimone (changes requested):
Also missing a README entry.
Everything else lgtm

joshuaellis:
Added readme & i've copied the model from the three.js example. Also renamed the component.

gsimone:
Merged 🥂

PR #151, opened 2020-10-02
Issue #127: Add a custom material story for Text

johnmarinelli:
Closes #127

PR #198, opened 2020-12-12
fix imports to support esModuleInterop:false

JustFly1984:
Here is my fresh approach to esModuleInterop: false;

I had to replace json files with ts.

I've met an issue with importing json files without esModuleInterop: true, due to rollup issues.

```
0 is not exported by src/helpers/matcap-assets.json
```

`import * as matcapList from 'src/helpers/matcap-assets.json'` works with webpack, but rollup fails.

trying to `import matcapList = require('src/helpers/matcap-assets.json') causes babel errors.

I've tried to update rollup plugins to new @rollup/plugin npm packages, but it did not solved issue.


gsimone (changes requested):
Just a couple of minor changes, otherwise we should be good to go

I'll merge this after #151 #203 and #199

PR #199, opened 2020-12-12
Feature/drei172 – CurveModifier

joshuaellis:
Second part to #172 adding CurveModifier.

Still to add `InstancedFlow` was wondering if you had a preference on either having a seperate component e.g. `InstancedCurveModifier` or we could modifier the current implimentation to check if `original.current` is `instanceof InstancedMesh`. 

I've bumped `three` to version `0.123.0` as the Flow class couldn't handle two materials - `meshNormalMaterial` being one of them.

Edit: Also updated some other files to coincide with the typescript updates `three` had, those were in `useSubdivider` and storybook `setup`.


gsimone:
Hey Joshua, sorry for not getting back at you sooner, I've been sick most of the week, here's some feedback:

- the API feels solid, I thought about it a bit and component seems like the best idea, since it's going to abstract away the fact that it's a whole new object
- the example might be better if you use a geometry that gets deformed, right now it's hard to understand what's happening, and people might think it's just animating along the curve

Otherwise, if you fix the conflicts I'll gladly merge and release a minor 👍

joshuaellis:
Hi @gsimone

Sorry to hear that, hope youre feeling better.

RE. example, sure, maybe I should just get it closer to what THREE have done [here](https://threejs.org/examples/?q=modif#webgl_modifier_curve) and use Text?

joshuaellis:
Updated the example, think this is ready to merge 🚀

gsimone:
Would you mind adding link to the README table of contents? (missing for useEdgeSplit too, I believe). If not I can do it this evening 👍

joshuaellis:
@gsimone no worries, i've updated it.

PR #203, opened 2020-12-18
Add cubemap presets to Environment component

mattrossman:
Resolves #202

This includes 9 preset cubmaps from the [Three.js examples](https://github.com/mrdoob/three.js/tree/master/examples/textures/cube). Preset options largely follow the corresponding folder name, though some have been shortened (e.g. "castle" instead of "SwedishRoyalCastle" and "sky" instead of "skyboxsun25deg"). If an invalid preset is provided, the error message will list all valid preset names. `preset` simply re-assigns `path` and `files` at the top of the component, so it will override them if set.

I also added a story for the Environment component that lets you preview the different presets.

⚠️ There seems to be a bug with this story where the environment stops updating after trying out a few presets. Also, the environment disappears if you go to a different story and then come back. This doesn't seem to stem from my modifications, because the component logs the correct values for `files` and `path` on each render. Open to suggestions here.

PR #209, opened 2020-12-26
Document Environment presets

mattrossman:
Follow-up to #203, this simply adds a link to the new Environment storybook entry and explains what each prop does.

I also renamed `cubemap-assets.ts` to `environment-assets.ts` since the files are technically equirectangular HDRIs, not cube maps.


README.md
```diff
@@ -324,10 +324,17 @@ Adds a `<Plane />` that always faces the camera.
 
 #### Environment
```
gsimone, line 325: Could you also add the link to the story?

mattrossman: The link is there already, it's on line 327

gsimone: How did I miss it? 👁️

PR #216, opened 2021-01-06
adds useFBO and ScreenQuad

dbismut:
Adds useFBO and ScreenQuad (the gist is stolen from https://github.com/gsimone/ombra).

As mentioned in the updated ReadMe, useFBO creates a `THREE.WebGLRenderTarget` or `THREE.WebGLMultisampleRenderTarget`.

```jsx
const target = useFBO(
  // width: 500,  height: 500, 
  // width and height are optional and defaulted to the viewport size
  // multiplied by the renderer pixel ratio, and recalculated whenever the
  // viewport size changes.
  {
    multisample: true, // if the renderer supports webGL2, it will return a WebGLMultisampleRenderTarget
    stencilBuffer: false // you can pass any options supported by THREE.WebGLRenderTarget
  } 
)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,8 @@ A box buffer geometry with rounded corners, done with extrusion.
 </RoundedBox>
 ```
 
+#### ScreenQuad
+
 ## Abstractions
 
 #### Text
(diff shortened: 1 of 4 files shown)


gsimone:
Please add stories to the storybook.

gsimone:
ScreenQuad is missing any kind of explaination in the README, I suggest copying from my repo

gsimone:
Both ScreenQuad and useFbo are missing a link in the table of contents

PR #217, opened 2021-01-07
Adds or fixes docs and stories for useFBO and ScreenQuad

dbismut:
Also includes minor fix for `<ScreenQuad />`.

@gsimone let me know what we should do about ScreenQuad as it's inspired by [this article](https://medium.com/@luruke/simple-postprocessing-in-three-js-91936ecadfb7) and is designed to feature a post-processing shading rather than be rendered as a plane.


gsimone:
I think you should merge master on this, there are build errors that should be fixed on that branch

gsimone:
@dbismut I fixed the merge conflicts, can you check it our real quick so we can merge?

PR #222, opened 2021-01-08
Feat/react-native-exports

### LINKED ISSUE #19: Text with react-native TypeError: element.setAttribute is not a function
Hi there,

When using the Text component in a react-native app (with expo-gl) I get the following error:
```
TypeError: element.setAttribute is not a function. (In 'element.setAttribute(eventName, 'return;')', 'element.setAttribute' is undefined)

isEventSupported
    react-dom.development.js:2484:18
<anonymous>
    react-dom.development.js:3828:15
<anonymous>
    react-dom.development.js:25213
loadModuleImplementation
    require.js:322:6
<global>
    index.js:36:7
loadModuleImplementation
    require.js:322:6
<unknown>
    AppEntry.bundle?platform=ios&dev=true&minify=false&hot=false:338646:45
loadModuleImplementation
    require.js:322:6
<unknown>
    AppEntry.bundle?platform=ios&dev=true&minify=false&hot=false:338533:26
loadModuleImplementation
    require.js:322:6
<unknown>
    AppEntry.bundle?platform=ios&dev=true&minify=false&hot=false:124908:53
loadModuleImplementation
    require.js:322:6
<unknown>
    AppEntry.bundle?platform=ios&dev=true&minify=false&hot=false:124791:51
loadModuleImplementation
    require.js:322:6
<global>
    AppEntry.js:5
loadModuleImplementation
    require.js:322:6
guardedLoadModule
    require.js:201:45
global code
    AppEntry.bundle?platform=ios&dev=true&minify=false&hot=false:409896:4
```

Simply importing the package seem to trigger the error.

### LINKED ISSUE #182: TypeError: element.setAttribute is not a function.
Hello drei-sters,

I have been struggling with drei a lot when it comes to trying to make it run on device (Android / Expo). Running it on web-browser works but device always returns the Exception error:

`TypeError: element.setAttribute is not a function. (In 'element.setAttribute(eventName, 'return;')', 'element.setAttribute' is undefined)`
This refers to line 321 of \node_modules\metro\src\lib\polyfills\require.js
This line is:

```
factory(

            global,
            metroRequire
            metroImportDefault,

```
etc

I am using the default code:

```
import React, { useRef, useState } from "react";
import { StyleSheet, View } from "react-native";
import { Canvas, useFrame } from "react-three-fiber";
import { Sky } from '@react-three/drei'; // this causes the error


function Box(props) {
  // This reference will give us direct access to the mesh
  const mesh = useRef();

  // Set up state for the hovered and active state
  const [hovered, setHover] = useState(false);
  const [active, setActive] = useState(false);

  // Rotate mesh every frame, this is outside of React without overhead
  useFrame(() => {
    if (mesh && mesh.current) {
      mesh.current.rotation.x = mesh.current.rotation.y += 0.01;
    }
  });

  return (
    <mesh
      {...props}
      ref={mesh}
      scale={active ? [1.5, 1.5, 1.5] : [1, 1, 1]}
      onClick={(e) => setActive(!active)}
      onPointerOver={(e) => setHover(true)}
      onPointerOut={(e) => setHover(false)}
    >
      <boxBufferGeometry attach="geometry" args={[1, 1, 1]} />
      <meshStandardMaterial
        attach="material"
        color={hovered ? "blue" : "red"}
      />
    </mesh>
  );
}

  return (
    <View style={styles.container}>
      <Canvas>
        <ambientLight />
        <Sky />
        <pointLight position={[10, 10, 10]} />
        <Box position={[-1.2, 0, 0]} />
        <Box position={[1.2, 0, 0]} />
      </Canvas>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: "black",
  },
});
```

import { Sky } from '@react-three/drei' is the part that causes the error, and any drei reference causes this. If I remove that and the reference to it on export default function App(), the project works perfectly on device. The project does work on browser even with drei.

I kindly appreciate advice or just a solid confirmation that drei does not work on device.

edit: changed the title.

### LINKED ISSUE #212: import drei error on React Native please fix this bug
While i import OrbitControls like : "import { OrbitControls } from 'drei' " i getting this error : TypeError: element.setAttribute is not a function. (In 'element.setAttribute(eventName, 'return;')', 'element.setAttribute' is undefined)

### LINKED ISSUE #221: react-native exports causing crashes because of internal imports
## What
There are currently three issues relating to imports in `react-native`
* #212 
* #182
* #19

## Why
This [comment](https://github.com/pmndrs/drei/issues/19#issuecomment-733558823) highlights the issue – `Html` and `Loader` components because they use either `react-dom` or `react-spring/web`. 

## Solution (potential)
Have an ad-hoc exporter for native, kind of like `react-spring` where `Html` would not be included. We *could* create a `react-native` version of `Loader` if we want.

joshuaellis:
Resolves #221 
Resolves #212 
Resolves #182 
Resolves #19 

Looked at what `react-spring` does and used that as a basis for this PR to fix those issues. Created a `native` that doesn't include `html` and `loader` and a `web` that includes all. `index` imports `web`


src/web.ts
```diff
@@ -0,0 +1,68 @@
...
+export * from './core/ContactShadows'
+
+// Shapes
+export * from './core/shapes'
+export * from './core/RoundedBox'
+export * from './core/ScreenQuad'
+
+// Prototyping
+export * from './core/useMatcapTexture'
+export * from './core/useNormalTexture'
```
joshuaellis, line 68: This could be a folder with the following structure
- web
- - index.ts
- - loader.tsx
- - html.tsx

We'd then remove `loader` and `html` from `core` and core could have a reliable `index` file.

joshuaellis:
Last thought would be - should `@react-spring/web` be included as a dependency. Native users don't need it... Could be a peer? Also, I will update readme after we're happy with implementations.

gsimone:
@drcmda what do you think? 

@joshuaellis I converted the PR to draft so that we don't merge it until everyone is ok 👍

I think this would break the stories, if paul is ok with the change, we need to also fix the imports ( right now they import from each module file, we should import from index to catch errors early)

joshuaellis:
> I converted the PR to draft so that we don't merge it until everyone is ok

sure that makes sense.

Yes it would break the stories, I think I was meant to suggest that we merge this into a beta, to ensure the exports work 100%. I'm working on #223 as well and this is a draft, these could compliment well together and both could be merged into a beta.

joshuaellis:
I've had a bit more of a think about it and I do think it's important to put `Loader` & `Html` in the web folder, I think for anyone coming to devlop it's clear that in the event that components can only be used by a particular target (although we all want to reduce that) that we put them in said folder.

gsimone:
Do you want to do that with a breaking change or keep supporting all imports from index?

joshuaellis:
It wouldn't be breaking (i think, feel free to correct me) as the root: `import {component} from '@react-three/drei'` use's `/web` as it's reference:

```ts
import { Loader } from '@react-three/drei' // works
import { Loader } from '@react-three/drei/web' // works
import { Loader } from '@react-three/drei/native' // does not work
```

This would be the same for `Html` too. Mearly just pushing native users to use the specific `native` folder, meanwhile (what I assume is most people) can carry on as normal.

gsimone:
Looks good to me, we'll add a paragraph for Native users in the README 👍

joshuaellis:
Awesome, will do & i'll resolve merge conflicts too.

joshuaellis:
Conflicts fixed, readme updated, checks are passing 🙌🏼 v2.2.0 here we come!

gsimone:
@drcmda can you give it a 👍?

joshuaellis:
@gsimone he's happy to merge into beta, left a message on [discord](https://discord.com/channels/740090768164651008/741751532592038022/800401752624594964)

PR #223, opened 2021-01-09
[Feature] move stories to TS

### LINKED ISSUE #219: Type for useTexture
The following snippet is from [@react-three/drei#useTexture](https://www.npmjs.com/package/@react-three/drei#usetexture).

```
const texture = useTexture(url)

const [texture1, texture2] = useTexture([texture1, texture2])
```

According to the NPM document, `useTexture` accepts two kinds of input, which is `string | string[]`. However, on version `2.2.15`, the type for `useTexture` is `useTexture(url: string): Texture | Texture[]`, which is causing a constraint of inputting `string[]` and outputting only `Texture` or `Texture[]`.

For example, the following code is reporting a type error.

```TypeScript
const texture = useTexture('/asset/stacy.jpg');

return (
    <group>
        <primitive obejct={object}
        <skinnedMesh {...props}>
            <meshStandardMaterial map={texture} />
            // texture doesn't match `map` property because it only accepts `Texture`, and the type for texture is `Texture | Texture[]`.
        </skinnedMesh>
    </group>
       )
```

In conclusion, `useTexture` should be generic, output `Texture` when the input is `string`, and output `Texture[]` when the input is `string[]`.

#### Package Version

- typescript: 4.1.3
- three: 0.124.0
- @react-three/drei: 2.2.15
- react-three-fiber: 5.3.11

### LINKED ISSUE #220: [Idea] Stories should be written in TS
I just saw #219 and realised if the stories were written in TS then this issue probably wouldn't have happened. I'm happy to work on this if you want to make the move, it's a bit of a chore, but in the long run it'd probably be useful (especially when making new components).

### LINKED ISSUE #226: Contact shadows example is broken
Contact shadow example on storybook doesn't work. I am seeing no shadows.

https://drei.react-spring.io/?path=/story/shaders-contactshadows--contact-shadow-st

Tested on Chrome for macOS.

### Screenshot
Screenshot 2021-01-10 at 22 17 10

joshuaellis:
## What

Move the storybook stories files to Typescript.

## Why

This will help us catch errors such as the import wasn't put in the `index` file. It'll also help us check our TS errors when using the library practically since there are no tests.

This will probably end up closing the following issues 
- resolves #220 
- resolves #219 
- resolves #226 

## Other
I think this would compliment #222 well, so ideally this could be merged into a beta too and both launched together. This is still WIP, hence the draft, but thought it would be useful to track for the wider team since it's a large change and there might be some suggestions to the TS writing of the stories.


src/Reflector.tsx
```diff
@@ -16,7 +16,7 @@ declare global {
   }
 }
 
-type Props = ReflectorOptions & { children: React.ReactElement<any> }
+type Props = ReflectorOptions & Omit<JSX.IntrinsicElements['mesh'], 'args'> & { children: React.ReactElement<any> }
```
joshuaellis, line 19: The THREE class `Reflector` extends `Mesh` so the story failed, although I find the TS for R3F confusing (becuase im no master) so this might be wrong?

src/MeshWobbleMaterial.tsx
```diff
@@ -4,7 +4,8 @@ import { useFrame } from 'react-three-fiber'
 
 type WobbleMaterialType = JSX.IntrinsicElements['meshStandardMaterial'] & {
   time?: number
-  factor: number
+  factor?: number
+  speed?: number
```
joshuaellis, line 8: the Ref example was proof that these args are optional in the props.

src/MeshDistortMaterial.tsx
```diff
@@ -7,8 +7,8 @@ import distort from './helpers/glsl/distort.vert.glsl'
 
 type DistortMaterialType = JSX.IntrinsicElements['meshPhysicalMaterial'] & {
   time?: number
-  distort: number
-  radius: number
+  distort?: number
+  radius?: number
```
joshuaellis, line 11: the Ref example was proof that these args are optional in the props.

joshuaellis:
will resolve #226

joshuaellis:
will resolve #176

.storybook/stories/useGLTF.stories.tsx
```diff
@@ -1,19 +1,20 @@
...
 export default {
   title: 'Loaders/GLTF',
   component: useGLTF,
-  decorators: [(storyFn) => <Setup cameraPosition={[0, 0, 5]}>{storyFn()}</Setup>],
+  decorators: [(storyFn) => <Setup cameraPosition={new Vector3(0, 0, 5)}>{storyFn()}</Setup>],
 }
 
 function Suzanne() {
-  const { nodes, materials } = useGLTF('suzanne.glb', true)
+  const { scene } = useGLTF('suzanne.glb', true)
```
joshuaellis, line 15: `nodes` and `materials` are not valid keys of `THREE.GLTF` so this would fail.

I can't seem to find any core documentation on this, but if people agree we could impliment an extension to `THREE.GLTF` that would create the type with those keys. ***Very*** quick proposal:

```ts
interface GLTF extends THREE.GLTF {
  nodes: [key: string]: THREE.Object3D
  materials: [key: string]: THREE.Material
}
```

joshuaellis:
Hi @gsimone,

So there are a few changes left: 

`useGLTF` needs this [PR](https://github.com/pmndrs/react-three-fiber/pull/930) to merged to R3F although im not sure it's the best solution so I put in a temp fix for now.

`useCamera` is throwing TS errors in relation to this issue https://github.com/pmndrs/react-three-fiber/issues/925#issuecomment-758625005 which i'm going to start looking at now.

But I thought it might be good to get the review process started whilst those to issues are going on.

gsimone:
Fixed a couple of minor things in useNormalTexture and useMatcapTexture, and removed js and jsx linting, since we are not going to have those anymore (and it was giving me error when running the linting script 🤷 )

joshuaellis:
>Fixed a couple of minor things in useNormalTexture and useMatcapTexture

Ah yeah, thank you! hopefully we won't need to declare `Texture` anymore as it should be fixed in https://github.com/pmndrs/react-three-fiber/pull/930

joshuaellis:
This is ready for merge, codebox CI is complaining because of pkg being private? Had to make tweeks to src files as I was going in regards to the TS impl, just to make you aware.

PR #228, opened 2021-01-14
useAnimations Story + TS fix

Code47X:
I recently added a PR to gltfjsx https://github.com/pmndrs/gltfjsx/pull/64 that typecast the group ref in order to use `useAnimations` without TS errors. I was thinking a better solution would be to just add `THREE.Group` as a valid ref option for `useAnimations`. This would allow gltfjsx to pass in the group ref without the need to typecast it first. If there is any interest in this PR and it gets merged I would be happy to go back and remove the typecast introduced in my previous gltfjsx PR.

Current gltfjsx output:
image


joshuaellis:
Group is basically the same as Object3D as it extends the class. Doesn’t this make this PR redundant? 

Would you mind making a basic sandbox just so I can look at this? If we do need to make a change I want to make sure it’s the right change.

joshuaellis:
Basically what I'm wondering is should the type *extend* `Object3D`? Because `AnimationMixer.clipAction`'s second arg is is an `Object3D`, so you can pass it a `Mesh` 🤔

Code47X:
@joshuaellis Hmm, that's a good point... I hadn't considered that making the type extend `Object3D`, but maybe that's the way to go. What would be the syntax to type a React ref like that? `React.MutableRefObject<something extends Object3D>`

joshuaellis:
First thoughts are:

```ts

type Api = {
  ref: React.MutableRefObject<Object3D | undefined>
  clips: AnimationClip[]
  mixer: AnimationMixer
  names: string[]
  actions: {
    [key: string]: AnimationAction
  }
}

export function useAnimations<Clip extends AnimationClip[], Ref extends Object3D>(
  clips: Clip,
  root?: React.MutableRefObject<Ref>
): Api
```

Hopefully this will infer types, the point of all this is so we dont have to actually type.... But we don't have a story to show `useAnimations`... There's a "little tokyo" model out there that I know has some animation tracks, it'd be good to make a story and test this TS implimentation before merging. 

Therefore, i'm going to move this PR to a draft & assign to both of us. Although I would still like to see the issue in action before we go down this rabbit hole of making generics work.

Code47X:
Looking at it again, I think the solution may be as simple as adding `| undefined` to the root ref. I'm playing with it locally and it's allowing me to pass in a Group ref without typecasting it, and working. I think that was what was causing the TS error in https://github.com/pmndrs/gltfjsx/issues/60

`function useAnimations(clips: AnimationClip[], root?: React.MutableRefObject<Object3D | undefined>): Api;`

Code47X:
https://codesandbox.io/s/strange-dream-mjpdw?file=/src/YBot.tsx

Things work properly, but TS complains about the ref type.

joshuaellis (changes requested):
Looks like it works from the sandbox, can you make a story for this please? We're missing one and you've practically made it in the sandbox so that would be really helpful.

Code47X:
Sure thing, I'll add in a story for `useAnimations` a little bit later today.

Just to be clear about the codesandbox, it only works in there without the change proposed in this PR because strict mode is disabled by default (so ts errors don't prevent compilation). In a normal project the ts error would prevent compilation.

joshuaellis:
thanks @Code47X. Also, please rebase from `master` we fixed the sandbox ci this morning 👍🏼

Code47X:
https://user-images.githubusercontent.com/5169367/104779494-d0f45c00-5744-11eb-8266-d3e6469a363f.mp4

I probably had too much fun on this story. Ready for review.

.storybook/stories/useAnimations.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+    mixamorigHips: THREE.Bone
+  }
+  materials: {
+    YB_Body: THREE.MeshStandardMaterial
+    YB_Joints: THREE.MeshStandardMaterial
+  }
+}
+
+type ActionName = 'Dance' | 'Idle' | 'Strut'
+type GLTFActions = Record<ActionName, THREE.AnimationAction>
```
joshuaellis, line 29: This doesn't look like it's been used?

.storybook/stories/useAnimations.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  React.useEffect(() => {
+    actions[selectedAction].reset().fadeIn(blendDuration).play()
+    return () => void actions[selectedAction].fadeOut(blendDuration)
+  }, [actions, selectedAction, blendDuration])
+
+  return (
+    <group ref={group} {...props} dispose={null}>
+      <group rotation={[Math.PI / 2, 0, 0]} scale={[0.01, 0.01, 0.01]}>
+        <primitive object={nodes.mixamorigHips} />
```
joshuaellis, line 57: This is really great to see & you've made the story look really lovely. I think if I was a first time user (which many are to `drei`, this story is a bit complicated to disect what part we're trying to show. 

I suggest that we abstract the model bits:
```tsx
<group rotation={[Math.PI / 2, 0, 0]} scale={[0.01, 0.01, 0.01]}>
    <primitive object={nodes.mixamorigHips} />
    <skinnedMesh geometry={nodes.YB_Body.geometry} skeleton={nodes.YB_Body.skeleton}>
      <meshMatcapMaterial attach="material" matcap={matcapBody} skinning={true} />
    </skinnedMesh>
    <skinnedMesh geometry={nodes.YB_Joints.geometry} skeleton={nodes.YB_Joints.skeleton}>
      <meshMatcapMaterial attach="material" matcap={matcapJoints} skinning={true} />
    </skinnedMesh>
  </group>
```

In to a new component called YBot & then have the old `YBot` could be called `YBotAnimationRig` or something of the sort. I'd just like to make it really simple and clear to someone at first glance, but I don't want to throw away the finesse you've added to the story 👍🏼

src/useAnimations.tsx
```diff
@@ -12,7 +12,7 @@ type Api = {
   }
 }
 
-export function useAnimations(clips: AnimationClip[], root?: React.MutableRefObject<Object3D>) {
+export function useAnimations(clips: AnimationClip[], root?: React.MutableRefObject<Object3D | undefined>) {
```
joshuaellis, line 15: I've had a play with this & I don't believe this is the full solution. This solution means I can pass
```ts
const group = useRef<THREE.Group>()

useAnimations(actions, group)
```

Only providing I **use** the ref in the application. Because if we don't, we're passing a ref that has `ref.current === undefined` which passes the check on l17 & throws errors at L31/2. Also some people will define `null` as the initial value of their ref so this would therefore fail typechecking:
```ts
const group = useRef<THREE.Group>(null)

useAnimations(actions, group)
```

So this needs a bit more thought.... I'll have a play & thought myself today - two heads better than one 💡

joshuaellis (changes requested):
Changes are in line with where they need to be, of course happy to discuss any further if you don't agree with them! 😄

.storybook/stories/useAnimations.stories.tsx
Code47X: Yeah you're right. I copied over a lot of the output from gltfjsx. I can remove these unused to clean up the story a bit.

.storybook/stories/useAnimations.stories.tsx
Code47X: No problem, thanks for the review. First time dealing with stories and to be honest I wasn't exactly sure what their use case was exactly, but that makes perfect sense. I'll try to emphasize the `useAnimations` a little more and abstract the model itself.

src/useAnimations.tsx
Code47X: Good catch, interested in any other solutions you see.

.storybook/stories/useAnimations.stories.tsx
joshuaellis: I think the best view to have on this storybook is to show people the capabilities of the abstractions/helpers etc. and **how** to use them / write their code.

joshuaellis: But it doesn't need to be every feature displayed, it's a top level kind of thing.

src/useAnimations.tsx
joshuaellis: hey @Code47X, ive done some investigation in to how `useRef` works with TS, take this example:

```ts
const groupRef = useRef<THREE.Group>()
```

This equates to `groupRef = THREE.Group | undefined` hence why your solution worked, however if I pass `null` like this:
```ts
const groupRef = useRef<THREE.Group>(null)
```
It'll probably throw an error, because `null` is not `THREE.Group | undefined` so we'd have to pass `<THREE.Group | null>` to the `useRef` generic. So i think the best solution is just adding `null` to the type you created! 

Then we want to check where we use something like `root.current` we're checking if it's defined // not null. To avoid any errors. Sorry if i just over explained things you already knew, thought it'd be good as a record, regardless.

Code47X: @joshuaellis Yup, makes sense to me. I'll throw in null to the ref type and make sure the ref has a `current` value before its used in the hook. I plan to make the changes we discussed sometime tonight! Thanks!

Code47X:
Hey @joshuaellis, I fixed up the null typing in `useAnimations` and cleaned up the storybook a little. I think `useAnimations` is where it needs to be now, everything seemed to be working. If necessary I can reopen a new PR to consolidate all the commits into one. (I was lazy initially and was working off the master branch of my fork, having to force push it everytime to rebase to pull latest changes :sweat_smile:).

joshuaellis (approved):

joshuaellis:
Great work @Code47X! It'll be on the beta release soon, there's a larger change to the repo structure coming (splitting native & web up), and I'd quite like to squash some bugs too.

Code47X:
@joshuaellis Wooo! Thanks for the help :fire:

PR #229, opened 2021-01-14
[fix] remove private from pkg.json

joshuaellis:
delete `private` instead of setting to false to fix merge checks...

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@react-three/drei",
   "version": "2.2.15",
-  "private": true,
   "description": "useful add-ons for react-three-fiber",
   "keywords": [
     "react",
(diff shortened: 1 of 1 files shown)


joshuaellis:
@gsimone the sandbox CI breaking all the time was bugging me, but I don't want to just merge the fix without understanding:
* how come the pkg.json for the project has `private:true`?
* what is the CI supposed to actually do?
* Do we want to keep it? not sure how long it's been broken for.

drcmda:
The private flag prevents you from publishing from root, because if you did the release is broken. You can only publish from dist so that end users can import directly without having to include the dist folder. Imo this pr would be kinda dangerous.

gsimone:
Need to change the publishDirectory config for CSB, meant to fix it but life attacked.

gsimone:
Fixed in #230

joshuaellis:
@drcmda  this is why we make PRs to review things 👌🏻 All I saw was an error in the logs.

@gsimone  🙏🏻

drcmda:
> this is why we make PRs to review things 👌🏻 

😅 that must have been a secret hint in there

Issue #235, opened 2021-01-15
[Feature] Center children wrapper


joshuaellis:
### Describe the feature you'd like:

Suggestion on [discord](https://discord.com/channels/740090768164651008/741751532592038022/799644579031679026). A component that would center it's children, this idea mainly came about from the need of constantly centering your models that are imported.

### Suggested implementation:

Here's what was proposed on the discord server:

```js
function Center({ children, ...props }) {
  const outer = useRef()
  const inner = useRef()
  useLayoutEffect(() => {
    const box = new THREE.Box3()
    box.setFromObject(inner.current)
    outer.current.position.copy(box.getSize().multiplyScalar(-.5))
  }, [children])
  return (
    <group ref={outer}>
      <group ref={inner}>{children}</group>
    </group>
  )
}
```

This would need to be written in Typescript however with an accompanying story & readme update.

PR #237, opened 2021-01-16
<Center />

drcmda:
reference: https://github.com/pmndrs/drei/issues/235

still misses a storybook (help wanted)

ps. accidentally contains another fix for environment, just an argument that wasn't needed.


joshuaellis:
I've added a center story, also there was a ref typecast as any, it should probably be `Group` so i've amended that, but if i've misunderstood, do say so.

joshuaellis:
resolves #235

Issue #242, opened 2021-01-19
In softShadows the `frustum` attribute is being called `frustrum` 


gsimone:
We should probably just alias it to the correct name

joshuaellis:
Is this on `three`?

gsimone:
https://github.com/pmndrs/drei/blob/93d842ad43c32dd3d5ad67e560c400801dfe8a99/src/softShadows.tsx#L12

PR #243, opened 2021-01-19
[fix] #242 Frustum has incorrect spelling in softshadows

joshuaellis:
I've added the correct spelling to the type & set as the default param in `pcss`. It'll try to use the user's `frustrum` spelling & if that's undefined (for new users) then it'll use the correct spelling as deafult.

Also the readme has been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -484,7 +484,7 @@ Injects [percent closer soft shadows (pcss)](https://threejs.org/examples/?q=pcs
 
 ```jsx
 softShadows({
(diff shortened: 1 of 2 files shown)


joshuaellis:
Not sure if that's what you had in mind in terms of aliasing?

gsimone (approved):

gsimone:
Perfect 👍  We'll deprecate next major

PR #245, opened 2021-01-19
useCameraShake

Code47X:
Is there was any interest in a camera shake system for drei? I made one recently based on this awesome talk: 
https://www.youtube.com/watch?v=tu-Qe66AvtY

It uses a `trauma` based system to modify the shake, as described in the talk. It's ideal for things like games where you often add bursts of varying intensity on top of a player controlled camera, but by modifying the config you can also use it in more general cases. It adds one dependency, `open-simplex-noise`, which is to generate the gradient noise used to generate the camera shake. Right now it only implements rotational shake for 3D scenes, but it could be easily extended to handle translational shake for 2D as well.

There is a storybook demo to play with the system. It has a little UI overlay to help visualize the 'trauma' and 'shake' values. I'm just making this system for fun, so even if no interest for drei, I'd appreciate any feedback on how to make this better!

(Note - The UI overlay in the storybook might be broken. It disappeared for me a couple times after a page refresh)


joshuaellis:
I've got a few questions and a few opinions: 

* IMO this is a heavy API return, have you looked at `useImperitiveRef`? This might be able to help tidy it up. 
* Also, it'd be good if you could explain the config, why do people need to set the seed etc. I can make assumptions, but if this is merged, it needs to be added to readme.
* `ShakeConfigPartial` can be written as `export type ShakeConfigPartial = Partial<ShakeConfig>` but why do both need to be exported? 
* Why are you returning `setConfig` if you wanted to change the config, you would pass it as the argument to the hook, no?
* Also does `setDefaultCamera` need to be called on every frame? (I'm not sure if it does or not but it seems odd?)
* I'm not sure if using refs as dependants of your `useEffect` is reliable enough to guarentee it fires, you might need to use `useCallback` as a ref style thing.

That should be enough to get the ball rolling 😄

gsimone:
A little feedback on the story:
- Toggle Decay should be a boolean field
- I'd add some of the settings ( let's say excluding the seeds) as knobs
- I understand how the lines are helpful, but they add a lot of noise to the story
- same with the scene, I'd make it way simpler, with wireframes or basic materials, just to keep the story super short

Although, since this is a pretty complex thing, I think we could make a more complete codesandbox example to add to the doc, with use-tweaks to show full capabilities.

The camera shake itself looks fantastic! Have you looked into directional shakes also? ( https://www.youtube.com/watch?v=fn3hIPLbSn8 )

Code47X:
@gsimone @joshuaellis Thanks for the input! Doing some refactoring today!

Code47X:
I did a little refactoring, here's my current thought process on things:

`useCameraShake` takes 2 arguments. A `controlledCamera` ref, and an optional `shakeConfig`. This second argument can be a partial, and its just used to overwrite any of the the default setting.

It returns a `ShakeController` object, with these control functions that can be destructued out:
```
  setTrauma: (val: number) => void
  getTrauma: () => number
  addTrauma: (val: number) => void
  clearTrauma: () => void
  updateConfig: (cfg: ShakeConfigPartial) => void
```
I made the config stateful, I don't think there is a reason it needed to be a ref. It's not updated in the three render loop. This also lets us regenerate the noise functions if one of the configSeeds is changed on a config update.

The `updateConfig` function updates the internal config. This can again be called with just a partial if you just need to do things like toggle decay. This just makes it easier to work with `shakeController`, as it allows us to pass it child components and still easily make changes. The hook's argument is only used to overwrites the initial defaults we work with.

I removed the UI overlay from the storybook, and I'll plan to make a codesandbox and some docs to demonstrate more of its usage. I added `@storybook/addon-controls` to control the shakeConfig. If you don't want to include the controls addon I can remove it, but it seems like controls will be supported going forward. I used addon-knobs for the buttons that add/clear trauma, as those allowed me to modify things without re-rendering the story. Should I slim down the story further?

`setDefaultCamera` is called every frame in `useCameraShake` to swap between the `shakeCam` and `controlledCam`. This seems to be safe, since its assigning the camera from 2 ref objects, it only re-renders when swaping from one to the other.

Any thoughts? Does this seem like a usable structure?

Code47X:
Can we change this to a draft pr? I'm working on a codesandbox, still some issues to work out.

joshuaellis:
I've been thinking about this feature a bit more, I'm trying to figure out if it's too specialised for `drei`, I can see the use case in general, it'd be great for games. I'm not saying we shouldn't include, I'm just wondering if:

a. there's already a game orientated repo for r3f abstractions//helpers
b. if not, maybe we should categorise this under "games" until we build a big enough repertoire and look to move out?

I'm just trying to think long term vision & what this repo really is for? Maybe I've got the wrong idea & actually this could be useful in regular scenes too? What do you think @gsimone & @drcmda?

gsimone:
This is useful in more situations than just games, for example in making animations, since drei is meant to be a catch-all thingie, I think we are ok merging all kind of stuff. If a pattern emerges and we get a lot of game-specific stuff we'll see 🤷‍♂️

joshuaellis:
Thanks @gsimone.

@Code47X I won't review again until you're ready, so just request me when it's time 👍🏼

Code47X:
Just posting the initial sandbox. I'll continue to add to it.

https://codesandbox.io/s/adoring-herschel-j6oob?file=/src/App.tsx

Code47X:
@joshuaellis Hey I think the useCameraShake is in a pretty stable spot right now. Do you have any thoughts/opinions on it?

I want to add in the option for translational shake as well (for 2D/Orthographic cameras). Any thoughts on how I should handle passing multiple config objects? Translational shake needs things like x/y/z-frequency, x/y/z-max instead of yaw-pitch-roll. I could throw all the options into one big shakeConfig, but that seems like it might be a little bloated.

.storybook/stories/useCameraShake.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+  maxRoll: 0.1,
+  yawFrequency: 4,
+  pitchFrequency: 4,
+  rollFrequency: 4,
+  yawNoiseSeed: 10,
+  pitchNoiseSeed: 20,
+  rollNoiseSeed: 30,
+}
+
+export const UseAnimationsSt = ({ ...args }) => <UseCameraShakeDemo cfg={args} />
```
joshuaellis, line 113: this name needs change from `UseAnimationSt` it should probably follow the naming convention in the others:
```js
 export const UseCameraShakeSt = ({ ...args }) => <UseCameraShakeScene cfg={args} /> 
```

package.json
```diff
@@ -52,6 +52,7 @@
     "glsl-noise": "^0.0.0",
     "lodash.omit": "^4.5.0",
     "lodash.pick": "^4.4.0",
+    "open-simplex-noise": "^2.4.0",
```
joshuaellis, line 55: I've been thinking about this, can we not use glsl-noise for this? I'd personally like to avoid adding dep's where we can...?

.storybook/stories/useCameraShake.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+    (storyFn) => (
+      <Setup cameraPosition={new THREE.Vector3(0, 0, 5)} controls={false}>
+        {storyFn()}
+      </Setup>
+    ),
+    withKnobs,
+  ],
+}
+
+function StoryBookKnobs({ shaker }: { shaker: ShakeController }) {
```
joshuaellis, line 23: I'm not familiar with Storybook controls, but does it have to be rendered as a component? Couldn't it just be a func?

src/useCameraShake.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export function useCameraShake(
+  controlledCam: React.MutableRefObject<Camera | undefined>,
+  shakeConfig?: ShakeConfigPartial,
+  initialTrauma?: number
+) {
+  const { setDefaultCamera } = useThree()
+  const [config, setConfig] = React.useState<ShakeConfig>({ ...defaultConfig, ...shakeConfig })
+  const shakyCam = React.useRef<Camera>(new THREE.PerspectiveCamera())
```
joshuaellis, line 50: This should probably be a memo?

src/useCameraShake.tsx
```diff
@@ -0,0 +1,125 @@
...
+      shakyCam.current.rotation.x += pitch
+      shakyCam.current.rotation.y += yaw
+      shakyCam.current.rotation.z += roll
+
+      if (config.decay) {
+        trauma.current -= config.decayRate * delta
+        constrainTrauma()
+      }
+
+      setDefaultCamera(shakyCam.current)
```
joshuaellis, line 117: I think paul has mentioned before that `setDefaultCamera` shouldn't be called in useFrame.

src/useCameraShake.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const clearTrauma = () => {
+    trauma.current = 0
+  }
+
+  const updateConfig = (cfg: ShakeConfigPartial) => {
+    setConfig({ ...config, ...cfg })
+  }
+
+  React.useEffect(() => {
```
joshuaellis, line 85: I think these effects should be memo'd too.

src/useCameraShake.tsx
```diff
@@ -0,0 +1,125 @@
...
+  initialTrauma?: number
+) {
+  const { setDefaultCamera } = useThree()
+  const [config, setConfig] = React.useState<ShakeConfig>({ ...defaultConfig, ...shakeConfig })
+  const shakyCam = React.useRef<Camera>(new THREE.PerspectiveCamera())
+  const trauma = React.useRef<number>(initialTrauma ? initialTrauma : 0) // range [0-1]
+
+  const yawNoise = React.useRef<Noise2D>(makeNoise2D(config.yawNoiseSeed))
+  const pitchNoise = React.useRef<Noise2D>(makeNoise2D(config.pitchNoiseSeed))
+  const rollNoise = React.useRef<Noise2D>(makeNoise2D(config.rollNoiseSeed))
```
joshuaellis, line 55: Yeah, if you change the effects to memo = less code

.storybook/stories/useCameraShake.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+import * as THREE from 'three'
+import { Camera, useFrame } from 'react-three-fiber'
+import { withKnobs, button } from '@storybook/addon-knobs'
+
+import { Setup } from '../Setup'
+
+import { useCameraShake, ShakeConfigPartial, ShakeController, PerspectiveCamera } from '../../src'
+
+export default {
+  title: 'Misc/useCameraShake',
```
joshuaellis, line 11: This should probably say `Cameras/useCameraShake`

.storybook/stories/useCameraShake.stories.tsx
Code47X: Ah I missed that, thanks.

joshuaellis:
I've added some thoughts, I need to have a play with the code to figure out the `as never` TS issue, so i'll get back to you on that one, i'm sure it can be solved 🔍

package.json
Code47X: I saw that glsl-noise was included in the deps, and was actually wondering that myself. I think it should be possible, I'll play around with it and see.

.storybook/stories/useCameraShake.stories.tsx
Code47X: I think your right, could just be a func.

src/useCameraShake.tsx
Code47X: Hmm ok, I wasn't aware of that. It seemed like it was only causing a re-render when the cameras were changing, but there is probably more going on there than I thought. I had it setting a state variable there initially, which would then call setDefaultCamera outside the renderloop, when the state changed. But that resulted in a double render - once when the state updated and once when the camera updated. I'll have to look into this part more.

.storybook/stories/useCameraShake.stories.tsx
Code47X: Haha ok... I just swapped it from Camera to Misc, wasn't sure where it belonged.

joshuaellis: No worries, IMO we should put it in Camera, cause it's camera based ⭐

package.json
Code47X: @joshuaellis I think that glsl-noise package something specific to shaders? I only see it used in a .glsl file.

joshuaellis: That would make sense given the repo name. Have you looked at `simplex-noise`? Looking the repo for `open-simplex-noise` and reading more about it? It seems that the non-open version is faster? Thoughts?

Code47X: I'm open to either. I don't know how relevant it is (or if its even true) but I thought there is a patent on `simplex-noise`? `open-simplex-noise` seems to be maintained still where the simplex-noise package hasn't been updated in a while - maybe there is just no need to. No strong opinions either way, I'll try them both out.

src/useCameraShake.tsx
```diff
@@ -0,0 +1,109 @@
...
+  clearTrauma: () => void
+}
+
+export function useCameraShake(
+  controlledCam: React.MutableRefObject<Camera | undefined>,
+  shakeConfig?: ShakeConfigPartial,
+  initialTrauma?: number
+) {
+  const { setDefaultCamera } = useThree()
+  const shakyCam = React.useMemo(() => new THREE.PerspectiveCamera(), [])
```
joshuaellis, line 48: If you're changing the default camera to perspective, does that mean that this hook can be only used with `PerspectiveCamera`? Because this is why theres a TS error on L86. The copy source has to be a `PerspectiveCamera` so the hook would probably break with `OrthographicCamera`.

gsimone: Ideally here you should use the default camera from `useThree`.

joshuaellis: I had a look at the code & did some suggestions @gsimone you can see them here – https://github.com/Code47X/drei/pull/1

Code47X:
@joshuaellis @gsimone Here is a working version based on the refactor. Here are some of my thoughts:

First, if we only want this hook to work on the default camera, we can leave out the `baseCamera` argument. With this argument though, the hook will work with any defined cameras we pass in, instead of only the default camera.

If this hook just returns the `shakyCam` ref, it will be up to the user to render from it to get the effect, since the default camera renders by default :brain:. This is effectively what `setDefaultCamera()` was doing before, just handled internally by the hook.

image

Do we want the code for manually controlling the camera shake in drei, or should I keep that part in a separate Repo. I commented it out on this last commit.

Manual control codesandbox for reference:
https://codesandbox.io/s/adoring-herschel-j6oob?file=/src/App.tsx

joshuaellis:
I think we need part of the API to `get` & `set`. I think you can remove `clear` & `add` imo. But without the API funcs it's like a one-time gig, I can see this being used in games where if you're hit multiple times you might wanna keep increasing the trauma...

I still wonder if we can make it work more ootb... 🤔

.storybook/stories/useCameraShake.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    </>
+  )
+}
+
+function UseCameraShakeDemo({ cfg }) {
+  const baseCamRef = React.useRef<Camera>()
+  const shakeCamRef = useCameraShake(baseCamRef.current, cfg)
+
+  useFrame(({ gl, scene }) => {
+    gl.render(scene, shakeCamRef.current)
```
joshuaellis, line 49: lets move this into the hook.

src/useCameraShake.tsx
```diff
@@ -0,0 +1,95 @@
...
+  maxRoll: 0.1,
+  yawFrequency: 1,
+  pitchFrequency: 1,
+  rollFrequency: 1,
+  yawNoiseSeed: 10,
+  pitchNoiseSeed: 20,
+  rollNoiseSeed: 30,
+}
+
+export interface ShakeController {
```
joshuaellis, line 23: lets trim this to a basic `set` & `get` you can doo `add` and `clear` with `set` & `get`.

src/useCameraShake.tsx
```diff
@@ -0,0 +1,95 @@
...
+  clearTrauma: () => void
+}
+
+type ShakeCamera = Camera & ShakeController
+
+export function useCameraShake(
+  baseCamera: Camera | undefined,
+  shakeConfig?: ShakeConfigPartial,
+  initialTrauma: number = 1
+) {
```
joshuaellis, line 36: can you add the return val please. & lets return ShakeController

src/useCameraShake.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export interface ShakeController {
+  setTrauma: (val: number) => void
+  getTrauma: () => number
+  addTrauma: (val: number) => void
+  clearTrauma: () => void
+}
+
+type ShakeCamera = Camera & ShakeController
```
joshuaellis, line 30: this is no longer being used.

src/useCameraShake.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export function useCameraShake(
+  baseCamera: Camera | undefined,
+  shakeConfig?: ShakeConfigPartial,
+  initialTrauma: number = 1
+) {
+  const cameraRef = React.useRef<Camera>(new THREE.PerspectiveCamera())
```
joshuaellis, line 37: Because of what we're doing, this should be `THREE.Camera` not the r3f Camera type.

src/useCameraShake.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const pitchNoise = React.useMemo(() => makeNoise2D(config.pitchNoiseSeed), [config.pitchNoiseSeed])
+  const rollNoise = React.useMemo(() => makeNoise2D(config.rollNoiseSeed), [config.rollNoiseSeed])
+
+  const constrainTrauma = () => {
+    if (trauma.current < 0 || trauma.current > 1) {
+      trauma.current = trauma.current < 0 ? 0 : 1
+    }
+  }
+
+  // React.useImperativeHandle(
```
joshuaellis, line 50: This can be turned into our return funcs.

joshuaellis (changes requested):
Hello, this is really great, I think it's come a long way and it's almost there. If we can make these changes, it'll work ootb for people & we're exposing a simple API that has a lot of flex. Please also add this hook to the readme ⭐ 🌟.

.storybook/stories/useCameraShake.stories.tsx
Code47X: @joshuaellis I made this change in the latest commit. I'm not sure it belongs inside the hook though. Since were giving it a render priority we are disabling automatic rendering. It feels like this hook shouldn't dictate the render priority. Just my take though, I'll leave it up to you.

joshuaellis: Hm. The hook doesn’t work without it being order 1 does it?

Code47X: Don't believe so, because it will render the scene out of the default camera (which isn't actually shaking) without the render priority.

joshuaellis: So maybe it's okay to highjack the render-order for this effect assuming there's an option to remove it's hierarchy? WDYT @gsimone

joshuaellis: we do this in `useCamera` I've decided it's a-okay lol.

joshuaellis:
@Code47X Looks all good, can you just add a description to the readme please?

src/core/useCameraShake.tsx
```diff
@@ -0,0 +1,82 @@
...
+      cameraRef.current.rotateX(pitch)
+      cameraRef.current.rotateY(yaw)
+      cameraRef.current.rotateZ(roll)
+
+      if (config.decay) {
+        trauma.current -= config.decayRate * delta
+        constrainTrauma()
+      }
+
+      gl.render(scene, cameraRef.current)
```
gsimone, line 75: I propose changing this to `setDefaultCamera` instead

```jsx
  useFrame(({ clock, setDefaultCamera }, delta) => {
    if (baseCamera === undefined) return

    if (trauma.current > 0) {
      cameraRef.current.copy(baseCamera)

      const shake = Math.pow(trauma.current, 2)
      const yaw = config.maxYaw * shake * yawNoise(clock.elapsedTime * config.yawFrequency, 1)
      const pitch = config.maxPitch * shake * pitchNoise(clock.elapsedTime * config.pitchFrequency, 1)
      const roll = config.maxRoll * shake * rollNoise(clock.elapsedTime * config.rollFrequency, 1)

      cameraRef.current.rotateX(pitch)
      cameraRef.current.rotateY(yaw)
      cameraRef.current.rotateZ(roll)

      if (config.decay) {
        trauma.current -= config.decayRate * delta
        constrainTrauma()
      }

      // 👇👇
      setDefaultCamera(cameraRef.current)
    } else {
      setDefaultCamera(baseCamera)
    }
  })
```

this would swap the camera instead of rendering, leaving the render to the user
and since cameraRef is a 1:1 copy it should support any kind of camera and postprocessing setups

joshuaellis: We had this, @drcmda said you shouldn’t use set default camera on every frame.

soz for formatting, on phone.

.storybook/stories/useCameraShake.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const baseCamRef = React.useRef<Camera>()
+  useCameraShake(baseCamRef.current, cfg)
+
+  return (
+    <>
+      <PerspectiveCamera makeDefault ref={baseCamRef} position={[0, 0, 20]} />
+      <React.Suspense fallback={null}>
+        <Scene />
+      </React.Suspense>
+    </>
```
gsimone, line 54:
```suggestion
  const { camera } = useThree()
  useCameraShake(camera, cfg)

  return (
    <>
      <OrbitControls />
      <React.Suspense fallback={null}>
        <Scene />
      </React.Suspense>
    </>
```

1. show how the default camera can be used
2. add OrbitControls to show that it doesn't conflict

Not sure about point 1, honestly, let me know what you think :D

src/core/useCameraShake.tsx
joshuaellis: Do you think whatever we decide now we should implement on useCamera too then?

gsimone: What about setting it only when it needs to change?

Code47X: This was maybe an overoptimization to begin with... I was only setting the default camera back to the original when there is no shake being applied (trauma = 0) to avoid copying the baseCamera's values (since nothing would be modified). We could just setDefaultCamera once on the initial render, then just make sure cameraRef copys baseCamera every frame.

Code47X:
I changed the latest commit to use `setDefaultCamera`. The storybook was updated to a working example with `OrbitControls` - its a little disorienting lol. I could only get it working calling `setDefaultCamera` inside the `useFrame` loop... I'm not seeing any issues, but if either of y'all can figure out a way to extract it out, feel free.

Only thing to note if using this hook, make sure not to have other camera's with the `makeDefault` prop on them.
`<PerspectiveCamera makeDefault />`

drcmda:
just saw this: https://twitter.com/pschroen/status/1357909701028175872

i guess i just needed a convincing example, im in 😅 please lets get this in

package.json
```diff
@@ -53,6 +53,7 @@
...
```
joshuaellis, line 56: `three` has a a simplex noise function - `three/examples/jsm/math/SimplexNoise` can we not use that?

README.md
```diff
@@ -167,6 +167,12 @@ A responsive [THREE.PerspectiveCamera](https://threejs.org/docs/index.html#api/e
 
 A responsive [THREE.OrthographicCamera](https://threejs.org/docs/index.html#api/en/cameras/OrthographicCamera) that can set itself as the default.
 
+#### useCameraShake
+
+[![](https://img.shields.io/badge/-codesandbox-blue)](https://codesandbox.io/s/simplex-camera-shake-j6oob)
+
+A hook for applying a configurable camera shake effect.
```
joshuaellis, line 174: this should have an example with the config roughly explaing config opts, a bit like `useAspect`

joshuaellis (changes requested):
I'd like to discuss using `three` simplex noise over adding a new package. and update the readme

package.json
Code47X: Ah cool I wasn't aware of that, should be able to swap to that.

joshuaellis: only if you think it'll work? I haven't looked at it in depth and tbf it isn't documented in THREE at all (surprised? no).

src/core/useCameraShake.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const setTrauma = (val: number) => {
+    trauma.current = val
+    constrainTrauma()
+  }
+
+  useFrame(({ clock }, delta) => {
+    if (baseCamera === undefined) return
+
+    if (trauma.current > 0) {
+      origin.current = origin.current.copy(baseCamera.rotation)
```
Code47X, line 62: You need to copy over the positional values each frame as well in case the camera is moving. Instead of creating 2 `Three.euler` refs, I think it's a little simpler to just copy over the camera object itself - plus that ensures we get a 1:1 copy.

joshuaellis: So the camera can dolly?

package.json
Code47X: Assuming it's some type of gradient noise function, should be easy enough to swap in. I can look into it a bit later today.

src/core/useCameraShake.tsx
Code47X: Yup, or if something like `<OrbitControls />` is moving the camera.

src/core/useCameraShake.tsx
```diff
@@ -0,0 +1,82 @@
...
+      const roll = config.maxRoll * shake * rollNoise(clock.elapsedTime * config.rollFrequency, 1) + origin.current.z
+
+      if (config.decay) {
+        trauma.current -= config.decayRate * delta
+        constrainTrauma()
+      }
+
+      baseCamera.rotation.x = THREE.MathUtils.lerp(origin.current.x, pitch, config.lerpValue)
+      baseCamera.rotation.y = THREE.MathUtils.lerp(origin.current.y, yaw, config.lerpValue)
+      baseCamera.rotation.z = THREE.MathUtils.lerp(origin.current.z, roll, config.lerpValue)
```
Code47X, line 77: I think this is an incorrect use of lerp for this case. Lerping to 80% of the value each frame makes the animation/configuration framerate dependent. The framerate difference is already taken into account in the noise functions, so the camera should be set to those value exactly to remain framerate independent.

Don't mean to be picky :sweat_smile: , I just work on a high refresh rate monitor and notice these things often.

joshuaellis: no it's a valid point, it's not there yet. To give you some context, we want to remove `setDefaultCamera` and we don't want to hijack the rendering.

Code47X: Cool agreed. The main issue comes down to assigning directly to the `baseCamera` again. In theory we want to keep the `baseCamera` pure, (allowing it to be modified externally). Applying rotations to baseCamera inside useCameraShake will cause it to compound on itself. I don't know if that's possible without setDefaultCamera... but I'll keep thinking about it.

joshuaellis:
Right so everyone is on the same page, this current code has two stories, one without OrbitControls and one with and right now, they both work. `setDefaultCamera` did not work because you change the camera, you pass the new camera and you get the new rotation values with the offset we just added, so it throws it out.

The issue is we're currently using `gl.render` and we're highjacking the render order by enforcing this component is no.1 which we're currently saying is not good.

This component currently works by getting where the defaultCamera of the scene is and adding a a little offset each frame. However it's important to note that every frame we get the **original** rotation set of the camera, so we're not adding offset ontop of each other. Imo the tricky thing is getting it to work with OrbitControls and without, because OrbitControls also controls rotation and position of the camera.

Code47X:
@joshuaellis I thought we were going with `setDefaultCamera` to avoid touching the render priority. We can pull it out of the useFrame loop and into the hook body - or do you not want to use `setDefaultCamera` at all?

joshuaellis:
When I converted it back to use `setDefaultCamera` the example without OrbitControls did not work. I think the general concensus is that this *should* be possible without changing cameras.

Can you help me understand why we're changing the rotation instead of the position?

Code47X:
I see what your saying now about setDefaultCamera and OrbitControls, so maybe render priority is correct. We are only changing the rotation inside the hook - because it only implements rotational camera shake currently. If we want to move/rotate the camera (not shake), we update the baseCamera itself - outside the hook.

README.md
Code47X: Added a simple example, just explaining the main config options. I tried to keep it pretty breif, wasn't sure how in depth I should go in the readme. I didn't go into some of the more advanced examples, but there is more info explaining those in the sandbox. More/Less in the readme, just lmk.

Code47X:
@joshuaellis I swapped the noise generation over to use the threejs simplex-noise and removed open-simplex-noise. I just looked at it briefly, but it seems to work well. The seed values currently don't do anything currently. I couldn't find much into on this... based on the .d.ts file it seems like the constructor takes an optional `r` object. I'm guessing that probably has the options for seed values, but I honestly have no idea what that `r` object is.

joshuaellis:
Looking at the source code for `Simplex` r is an object that should contain a `random` method & it defaults to Math. So it doesn't look like you can pass a seed. Is that an issue?

Code47X:
Ah ok cool. So I think its just assigning a random seed value each instantiation. Should'd be an issue, we can just remove the seed values from the config then. The only reason I exposed them in the config was for flexibility, if someone wasted to keep the roll/pitch/yaw in sync on the same cycle - but honestly I can't think of any use cases where that would be useful.

drcmda:
this is shaping up to be one of my favourite drei primitives, good work you all! https://t4l0f.csb.app :-)

src/core/useCameraShake.tsx
joshuaellis: > We could just setDefaultCamera once on the initial render, then just make sure cameraRef copys baseCamera every frame.

@Code47X  did you try this?

gsimone:
Something I thought about tonight instead of sleeping, could you add a story based on an impulse-like shake? 

So maybe just click on a ball, and get a screenshake

Code47X:
Here's the latest. CameraShake is now a component, it has 4 props - config, intensity, ref, additive.

- **config** - the camera shake config
- **intensity** - the intensity of the camera shake (formerly called trauma). This only sets the initial intensity. Use the ShakeController to modify in real time.
- **ref** - returns the ShakeController for modifying shake realtime - adding impulses etc.
- **additive** - The camera shake rotations will be additive. Use this if the camera's rotation values are set every frame. Allows camerashake to work with orbitcontrols.

src/core/useCameraShake.tsx
Code47X: I did, this solution works as well.

joshuaellis:
Nice! I’ll give it a good look over tomorrow 👌🏻

gsimone:
I made some minor changes to the story & removed the hold hook & related story. Readme should be updated for the new API (and explain the quirks for OrbitControls). Otherwise we're good

src/core/CameraShake.tsx
```diff
@@ -0,0 +1,80 @@
...
+export interface ShakeController {
+  getIntensity: () => number
+  setIntensity: (val: number) => void
+}
+
+export interface CameraShakeProps {
+  config?: ShakeConfigPartial
+  intensity?: number
+  additive?: boolean
+  children?: React.ReactElement
```
Code47X, line 28: Reminder to remove this. No longer needs children

drcmda:
@Code47X instead of `<CameraShake config={...} />` could we use direct component props? `<CameraShake maxRoll={1} ...` etc. the config prop would cause a memo invalidate every render and "config" props feel a bit weird in react.

Code47X:
@drcmda Yeah no problem we can use direct props. I thought having a separate config prop would keep things a bit more organized for when (I'd like to add transnational shake for use with 2D/Orthographic Cameras). But I'm cool with either.

We don't need the hook anymore - I just left it in this last commit for comparison purposes. We can remove one or the other before merge.

drcmda:
yeah that way it would be more inline with the other primitives. if users want a config object they can always memoize one and spread. but the downsides of course would be that most users wont use usememo and hence it would always invalidate since config={{ ... }} is a new object on every render.

joshuaellis (approved):
I've just removed the memo'd config for merging defaults with props, we can have defaults in the destructuring of props! I've also updated the readme for the component.

This has come a long way, I think it's a really great job @Code47X! Thanks to all involved, i'm happy to merge this into master now. @gsimone you happy?

gsimone:
LGTM

PR #246, opened 2021-01-20
Add MeshoptDecoder support to useGLTF

### LINKED ISSUE #241: Support EXT_meshopt_compression in useGLTF

I'd like to support the `EXT_meshopt_compression` glTF extension in the `useGLTF` hook, which enables further file compression beyond what Draco can achieve. See: https://www.npmjs.com/package/gltfpack

For example, with Draco alone I can shrink a 600 KB model down to 430 KB. With `gltfpack` using the `-c` flag I can shrink it all the way down to **130 KB**.

### Suggested implementation:

Supporting this involves:
```js
import { MeshoptDecoder } from 'three/examples/jsm/libs/meshopt_decoder.module'

// In the draco function
loader.setMeshoptDecoder(MeshoptDecoder)
```

I'm not sure if there's any disadvantage to always leaving this on, but if needed we could add another parameter after the draco path to allow users to customize the behavior.

mattrossman:
Resolves #241

I changed the `draco` factory into a more generic `extensions` factory so users can selectively disable one or the other feature. I'm defaulting this extension to on, since I'm not aware of any adverse effects and we already default draco on.

I don't think it's worth adding a story for this, but you can test it by loading a model which has been compressed via `gltfpack -c`, 

Example: [suzanne_meshopt.zip](https://github.com/pmndrs/drei/files/5839525/suzanne_meshopt.zip)


joshuaellis:
Thanks @mattrossman, I'd quite like to merge this into beta so i've changed the target and resolved a couple of conflicts for you.

drcmda:
looking good. i think it only actually pulls these binaries if there's a need for it, at least for draco that was always the case.

mattrossman:
@drcmda I'm not sure this dynamically pulls the binaries, since I used a regular import statement to pull in the decoder (it looks like the WASM is just a big string in that file). If that behavior would be preferred, I could switch it to a lazy `import()` call.

PR #247, opened 2021-01-20
feat: #181 – adding points to line

### LINKED ISSUE #181: Adding points in Line
Hi, firstly thanks for the project!

I am trying to draw a line with width where the path changes based on state. The number of points to be drawn can increase or decrease as state changes.

It seems that the geometry in `Line` is not updated when the number of points change. I realise this might be related to the way three works and might be desirable from a performance perspective, but if I use the basic r3f `line` element, the line is updated (presumably by re-creating the geometry) as points are added.

I tried hacking around with `onUpdate`, similar to how it works with `line`, but it didn't work.

I created codesandbox showing the issue: https://codesandbox.io/s/r3f-line-adding-points-y7utc?file=/src/index.js

The `Line` is created with 5 points initially, so you can add up to five points with the Add Point button, but after that new points don't appear.

Any help/workaround much appreciated!

joshuaellis:
### Why

You can't continously add points to a lineGeometry as seen in [this sandbox](https://codesandbox.io/s/r3f-line-adding-points-y7utc?file=/src/index.js)

### Solution

Create new geometry in an effect every time points is changed as suggested by @jugglingcats, because we don't realistically know how many points the user will need we probably shouldn't make a bufferGeometry big enough which is suggested [here](https://stackoverflow.com/questions/31399856/drawing-a-line-with-three-js-dynamically/31411794#31411794). But i'll be honest, i'm not sure if theres some kind of cost to this in the long term, I assume as soon as we replace the geometry the old one is disposed of correctly?

### Related Issues
* resolves #181

### Live example
https://codesandbox.io/s/drei-181-proposal-vjdtz?file=/src/index.js

### DIFF
--- a/src/core/Line.tsx
+++ b/src/core/Line.tsx
@@ -21,14 +21,19 @@ export const Line = React.forwardRef<Line2, Props>(function Line(
   ref
 ) {
   const [line2] = React.useState(() => new Line2())
-  const [lineGeometry] = React.useState(() => new LineGeometry())
   const [lineMaterial] = React.useState(() => new LineMaterial())
   const [resolution] = React.useState(() => new THREE.Vector2(512, 512))
+
   React.useEffect(() => {
-    lineGeometry.setPositions(points.flat())
-    if (vertexColors) lineGeometry.setColors(vertexColors.flat())
+    const lineGeom = new LineGeometry()
+    lineGeom.setPositions(points.flat())
+    if (vertexColors) {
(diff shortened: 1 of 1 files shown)


joshuaellis:
@gsimone I've looked into it a bit more, I don't think we have to dispose of the geometry before creating a new one?

joshuaellis:
Paul gave some solid advice & said to use `useMemo` over `useEffect`, so i've refactored it & I think it's good to be merged.

Issue #257, opened 2021-01-25
Request: Preload


drcmda:
When the webgl renderer first encounters an object, it will compile its materials and upload textures to the gpu. In scale this can tear into the framerate for a bit. Threejs has the ability to pre-compile, this component would allow us to do this declaratively, but it would also work in complex async scenarios, which would be fantastic.

Proposal:

#### Preload

The WebGLRenderer will compile materials only when they hit the frustrum. This can cause jank. This component precompiles the scene using [gl.compile](https://threejs.org/docs/index.html#api/en/renderers/WebGLRenderer.compile) which makes sure that your app is responsive from the get go. 

```jsx
<Canvas>
  <Scene />
  <Preload all />
```

By default gl.compile will only preload visible objects, if you supply the `all` prop, it will circumvent that. With the `scene` and `camera` props you could also use it in portals.

If you have async models you can drop it into the same suspense boundary *in concurrent mode*.

```jsx
<Canvas concurrent>
  <Suspense fallback={null}>
    <Model />
    <Preload all />
```

gsimone:
I don't understand the implication of this. Do !visible objects not get their shaders compiled?

drcmda:
webglrenderer.compile will only compile things that are visible. it's quite the problem, the moment something appears on screen threejs will churn through the compilation stage which bites into the framerate.

joshuaellis:
Why wouldn't this be part of r3f? It feels like a renderer thing?

```js
<Canvas preload />
```

drcmda:
because of suspense. it wouldnt be able to get the scene in a useEffect. imo this can only work by letting the user plug that in. the above - im not yet sure if that's the way yet, just putting it out there.

drcmda:
here's an example: https://codesandbox.io/s/cell-fracture-forked-h9pv8?file=/src/App.js

before:

<img width="611" alt="Screenshot 2021-01-25 at 22 29 57" src="https://user-images.githubusercontent.com/2223602/105768589-dfb8eb00-5f5c-11eb-95b6-7d3f4b12448f.png">

after:

<img width="756" alt="Screenshot 2021-01-25 at 22 30 42" src="https://user-images.githubusercontent.com/2223602/105768657-fb23f600-5f5c-11eb-9eb9-c63bd3311ec7.png">

drcmda:
btw that was the reason for todays twitter spat. the guy ran that demo and thought react creates jank. svelte man and vue fella smelled blood.

[...]

PR #258, opened 2021-01-25
Preload

drcmda:
for reference: https://github.com/pmndrs/drei/issues/257

#### Preload

The WebGLRenderer will compile materials only when they hit the frustrum. This can cause jank. This component precompiles the scene using [gl.compile](https://threejs.org/docs/index.html#api/en/renderers/WebGLRenderer.compile) which makes sure that your app is responsive from the get go. 

```jsx
<Canvas>
  <Scene />
  <Preload all />
```

By default gl.compile will only preload visible objects, if you supply the `all` prop, it will circumvent that. With the `scene` and `camera` props you could also use it in portals.

If you have async models you can drop it into the same suspense boundary *in concurrent mode*.

```jsx
<Canvas concurrent>
  <Suspense fallback={null}>
    <Model />
    <Preload all />
```


README.md
```diff
@@ -659,6 +660,19 @@ Calculates a boundary box and centers its children accordingly.
 </Center>
 ```
 
+#### Preload
+
+Precompiles the scene using [gl.compile](https://threejs.org/docs/index.html#api/en/renderers/WebGLRenderer.compile), this makes sure that your scene is responsive from the get go. If you have async models you must drop it into the same suspense boundary *in concurrent mode*. If you don't then simply put it into the canvas. By default gl.compile will only preload visible objects, if you supply the `all` prop, it will circumvent that.
```
joshuaellis, line 665: So does this only work in concurrent mode?

joshuaellis: And if I have multiple models, do I need to put it in each suspense block?

drcmda: nope. works everywhere.

```jsx
<Canvas>
  <mesh>
    <boxBufferGeometry />
   </mesh>
   <Preload />
 </Canvas>
```

now it will precompile the material and block render until it's done. this is common in threejs, people use gl.compile all the time. we just never had a declarative means to do it but you could have always just called it.

threejs people WILL face trouble with async stuff though because they have no means to manage it. this is the great part, preload will work here, too. but it must be concurrent, or else contents dont fire in lockstep - this now is a react thing and wouldn't be the place to explain. react has three renderers: legacy, blocking, cm. legacy works differently in suspense.

drcmda: added more explanations

PR #259, opened 2021-01-26
[RC] v3.0.0

### LINKED ISSUE #181: Adding points in Line

### LINKED ISSUE #221: react-native exports causing crashes because of internal imports

### LINKED ISSUE #233: Blob is not defined (nextjs - ssr)
- `three` version: 124
- `react-three-fiber` version: 5.3.12
- `@react-three/drei` version: 2.2.16

### Problem description:

with the new update I can see `Cloud.tsx` imports `CloudImage` which at some point calls `new Blob` I don't think `Blob` is defined in `node`.

### Suggested solution:

we could use `cross-blob` which is probably like `isomorphic-fetch`. Alt we can look at other ways that mean we don't need to use another library.

### LINKED ISSUE #238: Make <Environment /> work with a passed scene
When using a non-default scene, environment doesn't work. It's enough to add a `scene` prop:

```jsx
<Environment scene={myScene} />
```

This is a two-three lines change max

## TODO

- [ ] Make the change
- [ ] Add this case to the storybook

### LINKED ISSUE #241: Support EXT_meshopt_compression in useGLTF

### LINKED ISSUE #242: In softShadows the `frustum` attribute is being called `frustrum`

joshuaellis:
### Why

* resolves #242
* resolves #241 
* resolves #238 
* resolves #233 
* resolves #221 
* resolves #181 

### What

BREAKING CHANGES: we've had an internal refactor for #221, minimum supported `three` is 125

FEATURES: updated dependencies such as `troika-three-text`, added new props to components.

FIXES: Fixed spelling error & solved nextjs `blob is not defined error`

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### Things we want to do before release 

- [ ] Move documentation around
- [ ] Cleanup readme
- [ ] Update logo
- [x] Deprecate the `drei` package
- [x] Automate npm release
- [x] Min three version - 125 (add to peer deps)
- [x] Add Reflector component


joshuaellis:
@drcmda How were you imagining we'd move the documentation to the site? I don't mind shuffling all that around.
@gsimone i'd like to leave the logo sorting to you 👌🏼

joshuaellis:
> Automate npm release

Isn't this already done with tagging? Or did you want to use something like `semantic-release`?

gsimone:
> > Automate npm release
> 
> Isn't this already done with tagging? Or did you want to use something like `semantic-release`?

it's not automated though

joshuaellis:
semantic-release can be automated with a github [action](https://github.com/cycjimmy/semantic-release-action). But then you have to do a form of semantic commiting, I think you can still deploy manually if you want?

joshuaellis:
@gsimone docs plan is to move it to the docs work on the pmdrs site, do you think that's achievable? If you invite me as a maintainer I can help move over & help finish anything to get it ready?

gsimone:
@joshuaellis I can copypaste the README in the new docs, but some elements are still missing (for example the index table)

I'll give it a shot this evening

joshuaellis:
Sounds great, if you need any support @gsimone shout out.

PR #260, opened 2021-01-27
repurpose reflector, make it capable to blur

drcmda:
Screenshot 2021-01-27 at 10 33 52

sandbox: https://codesandbox.io/s/cell-fracture-forked-bfplr?file=/src/App.js:525-697

new props:

- args: allows us to set the plane, previously hardwired to 10/10
- mirror: when true reflects the scene instead of using a material color
- blur=[width, height]: when w/h are 0 it will skip blur and just reflect (default), being able to set width/height allows us to stretch blur downwards for a more realistic effect
- resolution: the size of the buffers, this better be as small as possible

new features:

renderchildren allow us to set the material, which wasn't possible before

```jsx
<Reflector>
  {(Material, props) => <Material {...props} roughness={0.5} color="black" />}
</Reflector>
```

TODO (in the future):

- make blur take roughness into account, that could increase blur on rough spots of the texture and decrease it on clearcoated spots, it would look amazing
- perhaps scale blur in distance, so that it increases vertically


src/Reflector.tsx
```diff
@@ -1,42 +1,128 @@
...
-      <primitive dispose={undefined} object={reflector} ref={ref as React.MutableRefObject<Mesh>} {...props}>
-        {React.Children.only(children)}
-      </primitive>
-    )
-  }
-)
+  return (
+    <mesh ref={meshRef} {...props}>
+      <planeBufferGeometry args={args} />
+      {children ? children('meshReflectorMaterial', reflectorProps) : <meshReflectorMaterial {...reflectorProps} />}
```
joshuaellis, line 125: im not quite sure what you're wanting to happen here with `children()`

joshuaellis:
I moved some of the materials out of their folder to avoid noise. with the types it was especially hard to follow what was going on. Just need some clarity on the `children` bit & then the story needs updating and we're good.

PR #262, opened 2021-01-27
Upgrade Three.js to r125

jure:
### Why

This purpose of this PR is to upgrade Three.js to r125 - mainly to deal with the removal of THREE.Geometry. It additionally fixes a few other issues (caused by the upgrade) along the way.

More details about the removal of SubdivisionModifier:
https://github.com/mrdoob/three.js/pull/21072

Some more details about the `THREE.Geometry` to `THREE.BufferGeometry` change can be found here: 
https://discourse.threejs.org/t/three-geometry-will-be-removed-from-core-with-r125/22401

### What

- Bump Three.js
- Updated `useEdgeSplit` (also supports the new `tryKeepNormals` option)
- Removed FaceNormalsHelper from useHelper example, as it no longer exists in Three.js
- Updated the `useTesselation` modifier, also so that it works when changing parameters in the storybook
- Updated `useSimplification` modifier
- Removed `useSubdivision` modifier, as it no longer exists in Three.js
- Updated typings for `CurveModifier`

### Checklist

- [x] Documentation updated
- [x] Updated stories where applicable
- [x] Ready to be merged


joshuaellis:
Only thing, is would you mind adding a peer dep for `three` it shoud be 0.125.x and above

joshuaellis:
Otherwise, really great work ⭐

PR #279, opened 2021-02-01
reflectorchanges

drcmda:
before (beta stage):

Screenshot 2021-02-01 at 13 17 37

after (stable release):

Screenshot 2021-02-01 at 13 20 09

after (stable release), setting depth to 0:

Screenshot 2021-02-01 at 13 17 44

it appears like depth adds some kind of smudge effect and misses out the clearcoat (roughness 0) spots which should reflect without blur. that was the purpose of the mixBlur factor, it's made to spike the difference in roughness for a more realistic effect. the lower it gets the more will some areas act like clear "mirrors". the higher the value goes, the more will blur take over.

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,4 +1,4 @@
 {
-  "sandboxes": ["4nc1u"],
+  "sandboxes": ["4nc1u", "bfplr"],
   "packages": ["dist"]
 }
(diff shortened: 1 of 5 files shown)


gsimone:
@emmelleppi I added the drei reflector sandbox to CI so you can use that to review :D

emmelleppi (approved):

emmelleppi:
Looks 👌 to me

PR #280, opened 2021-02-01
fix: Makes Loader go from 0 to 100 instead of 0 to 1

gsimone:
Come on.

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,4 +1,4 @@
 {
-  "sandboxes": ["4nc1u", "bfplr"],
+  "sandboxes": ["4nc1u", "bfplr", "1wh6f"],
   "packages": ["dist"]
 }
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ Drei currently exports OrbitControls [![](https://img.shields.io/badge/-codesand
 <Sphere>
   <meshBasicMaterial attach="material" color="hotpink" />
 </Sphere>
-``` 
+```
 
 #### RoundedBox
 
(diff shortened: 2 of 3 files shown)


joshuaellis:
It's funny, I saw this all last week, and told myself "I must fix it" and I never did, what would we do without you?

gsimone:
Thankfully I'm here, to fix the stuff that REALLY matters, like this U_U

joshuaellis:
Do you think we still need Travis? I'm not actually clear on what it does.

RenaudRohlinger:
@joshuaellis @gsimone This one makes me remember of one bug. When we want to load something new, the loader assets will stack because we use the same instance of LoadingManager. So it means for example it will do 0 -> 1 on the first load. Then 0.5 -> 1 etc.... because this one is never reset :
https://github.com/mrdoob/three.js/blob/master/src/loaders/LoadingManager.js#L6

So I guess we should use LoadingManager instead of DefaultLoadingManager like this :
https://github.com/RenaudRohlinger/drei/blob/master/src/core/useProgress.tsx
(I don't want to PR yet because I can't test yet but it should work)

joshuaellis:
Are you talking about if there are multiple suspense wrappers in an app?

gsimone:
Makes sense to use multiple Loaders, since this component is intended to "hide" the whole app - so it implies that it's only ever used one at a time

joshuaellis:
Can we write that in the docs then, please? I know it's a bit trivial, I do remember someone having an issue with that, and I did have a similar scenario myself and was confused about what was going on till I really thought about it 👀

drcmda (approved):

PR #282, opened 2021-02-02
refactor: replace storybook knobs with controls

### LINKED ISSUE #252: Storybook Knobs -> Controls

Transitioning from knobs to controls (and some of the theorized advantages) have been documented [here](https://medium.com/storybookjs/storybook-controls-ce82af93e430).

@gsimone discussed an attempt at this transition; any knowledge you have on some of the pitfalls you ran into would be appreciated.

Another relevant resource is [this open issue](https://github.com/storybookjs/storybook/issues/11984) on transitioning dynamic knobs:

jbixon13:
### Why

Replacing Storybook knobs with controls should allow for documentation with less Storybook-specific boilerplate. Resolves #252 

### What
- Installed `@storybook/addon-controls` to dev dependencies
- Refactored existing stories to use controls instead of knobs

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -2,7 +2,12 @@ const path = require('path')
 
 module.exports = {
   stories: ['./stories/**/*.stories.{js,ts,tsx}'],
-  addons: ['@storybook/addon-knobs/register', '@storybook/addon-actions', '@storybook/addon-storysource'],
+  addons: [
+    '@storybook/addon-knobs/register',
+    '@storybook/addon-controls',
+    '@storybook/addon-actions',
+    '@storybook/addon-storysource',
+  ],
   webpackFinal: (config) => {
     config.module.rules.push({
       test: /\.(glsl|vs|fs|vert|frag)$/,
--- a/.storybook/readme.md
+++ b/.storybook/readme.md
@@ -1,6 +1,6 @@
 #### Live Playground
 
-For examples of _drei_ in action, visit [https://drei.react-spring.io/](https://drei.react-spring.io/)
+For examples of _drei_ in action, visit [https://drei.pmndrs.vercel.app](https://drei.pmndrs.vercel.app)
 
 Or, run the demo storybook on your computer:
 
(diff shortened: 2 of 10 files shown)


jbixon13:
@joshuaellis let me know if you have any issues with how the `Billboard` story was converted before I continue this approach. I modeled it after the `useCameraShake` story in the PR you referenced.

joshuaellis:
LGTM, although before we continue, just wanna get @gsimone to have a quick look and see if he has any thoughts on this.

gsimone:
👍 👍

joshuaellis:
Full steam ahead @jbixon13 🚀  thanks again for doing this work, I know from experience it can be a proper chore editing **every** story. We really appreciate it.

jbixon13:
@joshuaellis I'm running into an issue in the `<Text/>` stories that will likely show up in other stories: 

- The current implementation has no knobs for all but the last story, but I was hoping to give the stories controls while I was refactoring them
- There are knobs/controls that are relevant to all `<Text/>` stories, and a few knobs/controls that are only applicable to specific ones
- An example of this is `outlineColor`, which is needed in the Outline story but not others. 
- This is usually not an issue, only relevant args are kept for each story
- This becomes an issue when an argType needs to be defined for that specific arg because it results in the control being present in stories where it isn't used
- An argType is needed for `outlineColor` so that its control is a color picker instead of a string
- There is an open issue that illustrates this: https://github.com/storybookjs/storybook/issues/12744 

Possible solutions: 
- Accept that controls that need argTypes will have to be present where they aren't used and the awkward UX that comes with that
- Provide controls only for args that are present in all versions of a component's stories, so a user cannot change these args

joshuaellis:
> Accept that controls that need argTypes will have to be present where they aren't used and the awkward UX that comes with that

This probably makes the most sense imo. I think we could perhaps have all the props that are currently set now & turn them to controls. Then set some predefined defaults that showcase the stories for Text we have now but then the user can play with them if they want, then it won't be *too* weird?

I hope that makes sense.

jbixon13:
That's pretty much what I'm currently doing, I'm looking at the props defined on the stories as they currently are and exposing controls for them (you can see this in the `clouds` story, there were no knobs exposed so I used the props available to create controls for the user). 

I'll push the stories as they are which includes the args where they aren't needed as you've said. If it ends up looking too weird we can try another approach. Thanks for your input!

joshuaellis:
Is this ready to be merged @jbixon13?

PR #286, opened 2021-02-03
fix: remove target prop type in controls where it isn't applicable

jure:
### Why

After looking at various controls during (#285) I noticed they all have `target` specified in their types. Only three actually need it/can use it on the Three.js side: `OrbitControls`, `TrackballControls` and `MapControls` (which is OrbitControls underneath).

### What

Removes prop types where they are not applicable/are misplaced. For PointerLockControls, this fix is handled in #285.

### Checklist

- [x] Ready to be merged


joshuaellis:
There's a fix on beta for ESM modules that I need to check this evening, so I'm gonna merge them into beta first. I have no doubt that you're code works though 🦖

PR #292, opened 2021-02-07
Make Reflector HMR-proof

CharlieHess:
### Why

Any time we wire up a useFrame, furiously null-check refs inside it as an HMR (in Next.js) will trigger an unmount but not before a final frame ticks off, crashing us.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged


joshuaellis:
LGTM, thanks!

PR #298, opened 2021-02-11
feat: Adds CubeCamera component

drcmda:
Help needed with the storybook

#### CubeCamera

[](https://codesandbox.io/s/frosty-reflector-forked-6k71y)

A [THREE.CubeCamera](https://threejs.org/docs/index.html#api/en/cameras/CubeCamera) that returns its texture as a render-prop. It wraps children and makes them invisible while rendering to the internal buffer. Using the `frames` prop you can control if this renders indefinitively or statically (a given number of times). If you have two static objects in the scene, make it `frames={2}` for instance, so that both objects get to "see" one another in the reflections, which takes multiple renders. If you have moving objects, unset the prop and render with a smaller resolution instead.

```jsx
<CubeCamera
  resolution={256} // Size of the off-buffer (256 by default)
  frames={Infinity} // How many frames it should render (Indefinitively by default)
  fog={customFog} // Allows you to pass a Fog or FogExp2 instance for a smaller frustrum
  near={1}
  far={1000}
>
  {(texture) => (
    <mesh>
      <sphereGeometry />
      <meshStandardMaterial envMap={texture} />
    </mesh>
  )}
</CubeCamera>
```

### TODO

- [ ] Add storybook example 
- [ ] Add storybook example for custom fog (maybe using FogExp2)


gsimone:
@joshuaellis I added the story but it's missing controls, thinking wether if I should add them or wait until we convert all stories to the new controls?

joshuaellis (approved):
@gsimone IMO let all the story changes be done in one go, let's fracture that movement, it's also easier to asses the whole thing in one go too.

PR #304, opened 2021-02-14
fix useprogress multi instanced and improved the related story

### LINKED ISSUE #302: useProgress is broken
- `three` version: 0.125.1
- `react-three-fiber` version: 5.3.18
- `@react-three/drei` version: 3.5.1

### Problem description:

useProgress is broken

### Relevant code:

just stays on 0 & also is never "active" therefore `Loader` will also be "broken"

https://drei-nsbsdwc7g.vercel.app/?path=/story/misc-useprogress--use-progress-scene-st

RenaudRohlinger:
### Why

fix #302 

### What

It seems that threejs is by default always using the DefaultLoadingManager so using a new instance of the LoadingManager won't work.
Because it uses the DefaultLoadingManager values are never reset.
I fixed it and added an offset that decrements the value for the new elements loaded when the previous loading was finished.

Before it would do :
0 -> 100% (let's say 1 -> 10 items where loaded)
Then for any new element loaded :
90% -> 100% ( 10 -> 11 items)

Now it will do the progress based on only the new element that is loading (10 - (last total loaded 10) -> 11 = 1).

I improved the story using multiple dynamic loading with Knobs to prevent errors.

The best solution would be to add a reset method to the LoadingManager in threejs.

### Checklist

- [x] Storybook entry added
- [x] Ready to merge

Demo : 
https://drei-git-fork-renaudrohlinger-fix-useprogress.pmndrs.vercel.app/?path=/story/misc-useprogress--use-progress-scene-st


joshuaellis (changes requested):
This just needs clearing up for me. Do we need to amend the documentation also?

RenaudRohlinger:
@joshuaellis I set back the use of DefaultLoadingManager and I just checked, nothing needs to be changed in the documentation. It should be safe and ready to merge.

Demo: 
https://drei-git-fork-renaudrohlinger-fix-useprogress.pmndrs.vercel.app/?path=/story/misc-useprogress--use-progress-scene-st

joshuaellis (approved):
LGTM, thank you for fixing this!

PR #307, opened 2021-02-16
fix: make Html component HMR-proof

CharlieHess:
### Why

image

### What

More safety checks that we need to make unmounting a peaceful process.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -83,7 +83,7 @@ export const Html = React.forwardRef(
     const [el] = React.useState(() => document.createElement('div'))
     const group = React.useRef<Group>(null)
     const old = React.useRef([0, 0])
-    const target = portal?.current ?? gl.domElement.parentNode
+    const target = portal?.current ?? gl.domElement?.parentNode
 
     React.useEffect(() => {
       if (group.current) {


src/web/Html.tsx
```diff
@@ -83,7 +83,7 @@ export const Html = React.forwardRef(
     const [el] = React.useState(() => document.createElement('div'))
     const group = React.useRef<Group>(null)
     const old = React.useRef([0, 0])
-    const target = portal?.current ?? gl.domElement.parentNode
+    const target = portal?.current ?? gl.domElement?.parentNode
```
joshuaellis, line 86: Do we do this else where? I'm wondering if we need to have a little audit. But tbh, `gl` comes from `react-three-fiber` so I'm surprised `typescript` isn't picking this up. Maybe the type needs to be changed in `r3f`.... What do you think?

joshuaellis (changes requested):
see comment

src/web/Html.tsx
CharlieHess: probably a good idea, although I understand from their perspective why they'd assume the DOM is always there. this really only bites you in SSR cases

CharlieHess: lemme take a peek into the type defs over there

CharlieHess: @joshuaellis OK upon closer inspection this type comes from [three](https://github.com/mrdoob/three.js/blob/master/src/renderers/WebGLRenderer.d.ts#L123), not react-three-fiber. so I think we'll have a tough time making a case for optionality there 😅

joshuaellis: Okay, I'll open an issue on [`three-ts-types`](https://github.com/three-types/three-ts-types) to discuss and see what others have to say. Maybe its something we should add, I'm not sure tbh. Thanks for looking into it.

joshuaellis (approved):

src/web/Html.tsx
joshuaellis: I've had a bit more of a think, this is too niche, although if you want to discuss it further, feel free to make the issue in `three-ts-types` lol

Issue #308, opened 2021-02-17
HTML Orthographic Camera does not zoom in and out accordingly


nathanbryantdev:
- `three` version: ^0.112.1
- `react-three-fiber` version:^4.2.21
- `@react-three/drei` version:^1.5.7
- `node` version: v12.18.4
- `npm` (or `yarn`) version: 6.14.6

### Problem description:

While using an Orthographic camera, the objectScale is not producing the correct scaling. It is blowing up the proportion of the HTML markers by 1000x

### Relevant code:
`function objectScale(el, camera) {
  if (camera instanceof THREE.PerspectiveCamera) {
    var objectPos = v1.setFromMatrixPosition(el.matrixWorld);
    var cameraPos = v2.setFromMatrixPosition(camera.matrixWorld);
    var vFOV = camera.fov * Math.PI / 180;
    var dist = objectPos.distanceTo(cameraPos);
    return 1 / (2 * Math.tan(vFOV / 2) * dist);
  }

  if (camera instanceof THREE.OrthographicCamera) return camera.zoom;
  return 1;
}`

### Suggested solution:
This will fix the issue: 
`
function objectScale(el, camera) {
  var objectPos = v1.setFromMatrixPosition(el.matrixWorld);
  var cameraPos = v2.setFromMatrixPosition(camera.matrixWorld);
  var vFOV = (camera.fov * Math.PI) / 180;
  var dist = objectPos.distanceTo(cameraPos);

  if (camera instanceof THREE.OrthographicCamera) {
    return 1 / ((2 * Math.tan(vFOV / 2) * dist) / camera.zoom);
  } else {
    return 1 / (2 * Math.tan(vFOV / 2) * dist);
  }
}
`

joshuaellis:
closed via #311

PR #309, opened 2021-02-17
fix: HTML scales with orthographic camera

nathanbryantdev:
HTML Orthographic Camera does not zoom in and out accordingly #308
https://github.com/pmndrs/drei/issues/308

The solution was to multiply the scale of the object by the camera.scale so that it lines up with what you would see in a Perspective Camera, but for Orthographic camera. Checks the instance of each and applies the calculation accordingly.

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -23,16 +23,19 @@ function isObjectBehindCamera(el: Object3D, camera: Camera) {
   return deltaCamObj.angleTo(camDir) > Math.PI / 2
 }
 
-function objectScale(el: Object3D, camera: Camera) {
-  if (camera instanceof PerspectiveCamera) {
-    const objectPos = v1.setFromMatrixPosition(el.matrixWorld)
-    const cameraPos = v2.setFromMatrixPosition(camera.matrixWorld)
-    const vFOV = (camera.fov * Math.PI) / 180
-    const dist = objectPos.distanceTo(cameraPos)
+function objectScale(el, camera) {
+  const objectPos = v1.setFromMatrixPosition(el.matrixWorld)
+  const cameraPos = v2.setFromMatrixPosition(camera.matrixWorld)
+  const vFOV = (camera.fov * Math.PI) / 180
+  const dist = objectPos.distanceTo(cameraPos)
+
+  if (camera instanceof OrthographicCamera) {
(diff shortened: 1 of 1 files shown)


joshuaellis:
Hi @nathanbryantdev,

I can see your deps are pretty outdated. I'd appreciate it if you could update those and double-check the bug remains (I'm sure it will). It'd be helpful if you could attach an example containing your changes (you can use the PR to create a codesandbox) just so we can be sure there are no unintended side effects as we don't have a story that captures this use case.

nathanbryantdev:
> Hi @nathanbryantdev,
> 
> I can see your deps are pretty outdated. I'd appreciate it if you could update those and double-check the bug remains (I'm sure it will). It'd be helpful if you could attach an example containing your changes (you can use the PR to create a codesandbox) just so we can be sure there are no unintended side effects as we don't have a story that captures this use case.

Hi @joshuaellis,

Ah awesome, let me give that a go and see if it helps.
I'll follow up with a codesandbox if the issue still persists.

I appreciate your quick response!

gsimone:
If you see this is still an issue and the PR fixes it, would you mind adding a story for HTML with Orthographic camera?

nathanbryantdev:
> If you see this is still an issue and the PR fixes it, would you mind adding a story for HTML with Orthographic camera?

Sure thing @gsimone 

Also @joshuaellis, I've gone ahead and updated all of the deps and the issue's still persisting, this PR remains the fix for this particular issue so let me go ahead and create a storybook story with the ortho cam with HTML for Gianmarco and throw up a codesandbox of the issue in question.

You wouldn't happen to be able to link me to a template with the packages installed already? (I assume you may have one handy or know how to spin up a template codesandbox React instance faster than myself), else I'll get it sorted ASAP.

Thanks both.

nathanbryantdev:
Potentially a silly question, how would I go about created a codesandbox from this PR?
I've given
`npm i https://pkg.csb.dev/pmndrs/drei/commit/bd370a2a/@react-three/drei`
a try but to no avail, any tips?

nathanbryantdev:
Morning all, closed this PR as I've opened up another with the storybook added, not entirely sure what happened with my git but I encountered some issues and was unable to edit this PR.

PR #311, opened 2021-02-17
Bugfix/html scaling for orthorgraphic camera

nathanbryantdev:
HTML Orthographic Camera does not zoom in and out accordingly https://github.com/pmndrs/drei/issues/308

The solution was to multiply the scale of the object by the camera.scale so that it lines up with what you would see in a Perspective Camera, but for Orthographic camera. Checks the instance of each and applies the calculation accordingly.

Added Storybook story with animation that mimics zoom in and out of HTML utilising the Orthographic Camera.


joshuaellis:
Can you resolve conflicts please?

nathanbryantdev:
> Can you resolve conflicts please?

@joshuaellis Sure, done 👍

src/web/Html.tsx
```diff
@@ -26,16 +26,19 @@ function isObjectBehindCamera(el: Object3D, camera: Camera) {
...
-    const vFOV = (camera.fov * Math.PI) / 180
-    const dist = objectPos.distanceTo(cameraPos)
+function objectScale(el, camera) {
+  const objectPos = v1.setFromMatrixPosition(el.matrixWorld)
+  const cameraPos = v2.setFromMatrixPosition(camera.matrixWorld)
+  const vFOV = (camera.fov * Math.PI) / 180
+  const dist = objectPos.distanceTo(cameraPos)
+
+  if (camera instanceof OrthographicCamera) {
+    return 1 / ((2 * Math.tan(vFOV / 2) * dist) / camera.zoom)
```
joshuaellis, line 36: do you think we could extract this into `const objScale = 1 / ((2 * Math.tan(vFOV / 2) * dist)` to avoid the dupe & make it clearer what the difference is between perspective camera  & orthographic?

nathanbryantdev: Ahhh of course, without a doubt! I'll fix it up now.

joshuaellis (approved):
Great work! Thanks for your contribution.

nathanbryantdev:
@joshuaellis My pleasure, all the best!

PR #312, opened 2021-02-17
fix: calculates the reflector depth texel with the right coordinates

emmelleppi:
### Why
in presence of the normalMap and the depthBuffer some artifacts of the depth texels were visualized

### What
if it uses the normalMap distortion then it calculates the depth texel with the distorted coordinates

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


joshuaellis:
Hey @emmelleppi, no rush on this, in this PR would you mind wrapping `Reflector` in `forwardRef`?

joshuaellis (approved):
Lot of voodoo in there that is beyond me right now, but the stuff I do understand looks 💯 you should definitely celebrate this change, everyone loves a performance boost 🎉

PR #315, opened 2021-02-27
[Ready !] Use three-stdlib instead of examples/jsm

RenaudRohlinger:
### Why

Tree shaking, transpilation not required anymore,  cleaner code...

### What

Replace three/examples/jsm by three-stdlib

### Checklist

- [ x ] Documentation updated
- [ x ] Ready to be merged
- [ x ] Updated stories

### DIFF
--- a/.storybook/stories/Line.stories.tsx
+++ b/.storybook/stories/Line.stories.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { Vector3 } from 'three'
-import { GeometryUtils } from 'three/examples/jsm/utils/GeometryUtils'
+import { hilbert3D } from 'three-stdlib/utils/GeometryUtils'
 import { withKnobs, number, color, boolean } from '@storybook/addon-knobs'
 
 import { Setup } from '../Setup'
@@ -12,7 +12,7 @@ export default {
   component: Line,
 }
 
-const points = GeometryUtils.hilbert3D(new Vector3(0), 5).map((p) => [p.x, p.y, p.z]) as [number, number, number][]
+const points = hilbert3D(new Vector3(0), 5).map((p) => [p.x, p.y, p.z]) as [number, number, number][]
 
 const colors = new Array(points.length).fill(0).map(() => [Math.random(), Math.random(), Math.random()]) as [
   number,
--- a/.storybook/stories/MapControls.stories.tsx
+++ b/.storybook/stories/MapControls.stories.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { SVGLoader } from 'three/examples/jsm/loaders/SVGLoader'
+import { SVGLoader } from 'three-stdlib/loaders/SVGLoader'
 import { Box3, Sphere, Vector3 } from 'three'
 import { useLoader, Canvas } from 'react-three-fiber'
 
--- a/.storybook/stories/Shapes.Parametric.stories.tsx
+++ b/.storybook/stories/Shapes.Parametric.stories.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { Vector3 } from 'three'
-import { ParametricGeometries } from 'three/examples/jsm/geometries/ParametricGeometries.js'
+import { ParametricGeometries } from 'three-stdlib/geometries/ParametricGeometries.js'
 
 import { Setup } from '../Setup'
 import { useTurntable } from '../useTurntable'
--- a/.storybook/stories/TransformControls.stories.tsx
+++ b/.storybook/stories/TransformControls.stories.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { withKnobs, optionsKnob, boolean } from '@storybook/addon-knobs'
-import { TransformControls as TransformControlsImpl } from 'three/examples/jsm/controls/TransformControls'
+import { TransformControls as TransformControlsImpl } from 'three-stdlib/controls/TransformControls'
 
 import { Setup } from '../Setup'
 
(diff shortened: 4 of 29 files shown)


joshuaellis:
I'm gonna be honest here, I don't think we should be moving `drei` over to `three-stdlib` straight away. I'd rather wait till we're confident in `three-stdlib` to be at least v1. 

This library is heavily used this being a breaking change as we're going to be forcing users to install `three-stdlib`. I'd really like to be confident in the whole move over, `three-stdlib` doesn't have an typings yet, and I don't think `@types/three` will correctly link them?

If we do want to get the ball rolling on v4, which I'm open to, we should be merging to beta and release a beta for thorough testing, but straight into `master` I think could lead to problems.

drcmda:
> I'm gonna be honest here, I don't think we should be moving `drei` over to `three-stdlib` straight away. I'd rather wait till we're confident in `three-stdlib` to be at least v1.
> 
> This library is heavily used this being a breaking change as we're going to be forcing users to install `three-stdlib`. I'd really like to be confident in the whole move over, `three-stdlib` doesn't have an typings yet, and I don't think `@types/three` will correctly link them?
> 
> If we do want to get the ball rolling on v4, which I'm open to, we should be merging to beta and release a beta for thorough testing, but straight into `master` I think could lead to problems.
> 
> I'd also like to explore some possibilities where this library could still be used with `three` for the examples, but I'm not sure if I'm reaching too far, even if it was for a depreciated version of the files for users so bugs can still be fixed.

is it a breaking change though? drei is mostly abstracting jsm/examples and it should not have any adverse effects. even if someone uses a ref.current, that would be so quick to typecast.

right now we just cause quite large issues: 

- people can no use drei in gatsby, next, etc, without weird transpilation hacks
- all our controls do not work on mobile devices (due to bugs in jsm), that is worst imo
- bundles get large bc jsm is unoptimzed

users wouldn't have to download three-stdlib, that would just be a dependency, and only for those things that we exchange, while other things can still pull from jsm.

joshuaellis:
> is it a breaking change though? drei is mostly abstracting jsm/examples and it should not have any adverse effects. even if someone uses a ref.current, that would be so quick to typecast to @types/three. that to me is not that big of a deal because if three can just take away types altogether, we could ask one user out of thousands to please typecast that one ref. alternatively we just copy the types quick.
> 
> right now we have these issues:
> 
>     * people can no use drei in gatsby, next, etc, without weird transpilation hacks
> 
>     * people can not test drei easily bc doesn't run under node
> 
>     * all our controls do not work on mobile devices (due to bugs in jsm), that is worst imo
> 
>     * bundles get large bc jsm is unoptimzed
> 
> 
> and they all could be fixed, without any major annoyance that i see.
> 
> users wouldn't have to download three-stdlib, that would just be a dependency, and only for those things that we exchange, while other things can still pull from jsm. there would be no bundle overhead or anything like that, mostly bundles would just get smaller.
> 
> i'd be in favour of at least starting to replace single primitives, im thinking controls - because it would be such a relief to finally put a dam into the stream of bug messages r3f receives because mobile apps can't be interacted with.

ATM we don't bundle deps like `three` & `three-stdlib`, they're peer deps. I would prefer to keep it that way otherwise we have to ensue that the version of `three-stdlib` we bundle has the same `three` peer dep as `drei` that's an overhead we can probably live without? Otherwise, people complain that things aren't compatible. Therefore I believe adding a new peer dep would be a breaking change.

The TS issues aren't a major blocker for me, it's annoying but the benefits on the library outweigh this.

Totally agree with the problems you listed and I'm not saying we shouldn't do this right away, I just think putting it on `master` could be shooting ourselves in the foot. I'd much rather be safer and put it in `beta` to do some actual testing in real projects. Similar to the v3 release, we had a lot of issues we were able to solve through `beta`-ing the package that reduced the stress of maintaining.

The sticking point for me is, do we do it as a dep or a peer dep and what are the implications and maintenance overheads associated.

drcmda:
stdlib is also just a peer dep. i would propose we add types to the classes we replace then. that will take a couple minutes at most.

joshuaellis:
If it's the same why isn't `three-stdlib` published as v1? 

Also, you want to bundle the library, there's a lot to be done in the library, I can see its growing quickly which is wonderful but also see small bugs coming through like `let & const`. Minor I know but every time you release the library we'd have to re-release `drei` so I'd prefer to keep it as a peer dep, it's an unnecessary overhead, time we as maintainers could have better spent on fixing bugs and improving libraries.

drcmda:
sure, but lets keep it simple, stdlib has, what, 300 primitives. drei uses 3 or 4, or let it be 10. we could start with a single one, say orbitcontrols, making sure it works. then trackball. then another. these are relatively contained and wouldn't need that much churn.

if we're supposed to make sure everything in std is working then that won't happen, the scope of this is far too big. for some things we need people to run into bugs to know about them. but in drei's case where we curate a small amount, it's easier to move forward.

joshuaellis:
No, I don't expect everything in `three-stdlib` to be working, that would be ridiculous lol. Just the tiny bugs, but let's do this:

* merge all this
* merge as peer dep
* release beta v4
* review for official release next week

That way we can revise as we go with a small impact for the time being & test it next week in our own projects.

joshuaellis:
@RenaudRohlinger can you change this from a draft when it’s ready for review please? ☺️

RenaudRohlinger:
@joshuaellis @drcmda @gsimone I tried to test and fix all the stories and some codesandbox and everything looks good. We can release a canary version 👍  https://codesandbox.io/s/drei-reflector-forked-dz16x?file=/src/App.js

joshuaellis (approved):

PR #318, opened 2021-03-01
Added Html scaling for orthographic zooming

JDihlmann:
### Why
Orthographic zooming will not change Html scaling. 

### What
Html useFrame wont update scaling without position or transform change, therefore only zooming an orthographic camera would have no impact. Added support for register zooming. Additionally fixed zooming scale since FOV might be 0 (someone should take a look at that - FOV should not be zero).


src/web/Html.tsx
```diff
@@ -34,7 +34,9 @@ function objectScale(el: Object3D, camera) {
   const scaleFOV = 2 * Math.tan(vFOV / 2) * dist
 
   if (camera instanceof OrthographicCamera) {
-    return 1 / (scaleFOV / camera.zoom)
+    // Maybe someone should take a look at this scaleFOV can be 0
```
neftaly, line 37: Might be worth putting the check on a separate line instead of in a ternary, for readability and to make it easier to delete if someone fixes the problem. Is it also worth creating a new github issue?
```
if (scaleFOV === 0) return camera.zoom
return 1 / (scaleFOV / camera.zoom)
```

joshuaellis: I'd rather have this done in 1 PR tbh – @JDihlmann ?

JDihlmann: So I looked at the definition of the [OrthographicCamera ](https://github.com/mrdoob/three.js/blob/master/src/cameras/OrthographicCamera.js) of Three that is implemented in drei and it seems like it does not even have an fov property or am I missing something? Therefore why does camera.fov evaluates to 0 in drei shouldn't it be undefined? Either way if camera.fov is 0 -> vFov is 0 -> tan(0)=0 -> scaleFOV is 0. What I don't get is why I got changed in this [commit ](https://github.com/pmndrs/drei/pull/311) if the orthographic camera does not hold any fov property @nathanbryantdev

JDihlmann: So I would cahnge it to something like:
`if (camera instanceof OrthographicCamera) {
    return camera.zoom
  } else if (camera instanceof PerspectiveCamera) {
    const objectPos = v1.setFromMatrixPosition(el.matrixWorld)
    const cameraPos = v2.setFromMatrixPosition(camera.matrixWorld)
    const vFOV = (camera.fov * Math.PI) / 180
    const dist = objectPos.distanceTo(cameraPos)
    const scaleFOV = 2 * Math.tan(vFOV / 2) * dist
    return 1 / scaleFOV
  } else {
    return 1
  }
`
But again I don't know why it got changed in #311

joshuaellis: @JDihlmann, thanks for giving an example, it's really helpful, I think we're right to ask @nathanbryantdev his thoughts, I don't often use OrthographicCamera myself so I'm really leaving it to you folk! But I'm definitely interested in learning the solution.

JDihlmann: I also noticed the scaling offset described in #311 but I simply fixed it by setting the scaleFactor of the Html elements to 0.01. So maybe the solution is as simple as returning 1 / camera.zoom for the Orthograpic Camera rather than only camera.zoom. But yes lets wait for a response from @nathanbryantdev @NathanBryant

nathanbryantdev: I do tend to agree with your logic @JDihlmann, what you've said makes sense. I am now under the impression that this issue is not a one shoe fits all solution as a result of ScaleFactor attribute applied to the component and the 

To fully leverage the ability to use either Orthographic or Perspective camera with the HTML component a few different things need to be taken into account:
ScaleFactor, camera FOV, and camera Zoom

ScaleFactor requires the scale factor of 1 to be used in conjunction with orthographic camera with the above snippet of code you've provided, whereas the perspective camera requires the FOV be taken into account as well as the camera zoom and scale factor for the above snippet.

I would propose an additional change to the factoring of the scale:
`function objectScale(el: Object3D, camera, scaleFactor: number) {
  const objectPos = v1.setFromMatrixPosition(el.matrixWorld);
  const cameraPos = v2.setFromMatrixPosition(camera.matrixWorld);
  const vFOV = (camera.fov * Math.PI) / 180;
  const dist = objectPos.distanceTo(cameraPos);
  const scaleFOV = 2 * Math.tan(vFOV / 2) * dist;

  if (camera instanceof OrthographicCamera) {
    const orthographicDivisor = 2.5;
    return camera.zoom / orthographicDivisor;
  } else if (camera instanceof PerspectiveCamera) {
    return (1 / scaleFOV) * scaleFactor;
  } else {
    return 1;
  }
}

const scale =
            scaleFactor === undefined
              ? 1
              : objectScale(group.current, camera, scaleFactor);
` 

This will span all use cases I believe, can we get some confirmation please @JDihlmann

JDihlmann: I tested it and first of all the calculation of FOV need to be moved within the if statement since the typed camera object does not contain a .fov property. Furthermore I do not really understand why your ignoring the scaleFactor property for orthographic camera and introduce a non variable constant? You just could realize the orthographicDivisor by using the scaleFactor rather than ignoring it. 

```
function objectScale(el: Object3D, camera: Camera) {
  if (camera instanceof OrthographicCamera) {
    return camera.zoom 
  } else if (camera instanceof PerspectiveCamera) {
    const objectPos = v1.setFromMatrixPosition(el.matrixWorld);
    const cameraPos = v2.setFromMatrixPosition(camera.matrixWorld);
    const vFOV = (camera.fov * Math.PI) / 180;
    const dist = objectPos.distanceTo(cameraPos);
    const scaleFOV = 2 * Math.tan(vFOV / 2) * dist;
    return (1 / scaleFOV)
  } else {
    return 1;
  }
}
```

If you would choose scaleFactor to be 1 / 2.5 you would have the same effect

JDihlmann: So ok I thought about a little longer and you are right in the sense that the scaling for an Html object with an orthographic camera and a set scaleFactor should depend on three things.
- The scaleFactor itself
- The distance to the corresponding group the Html lies in
- The zoom of the camera 
The question is how do we stitch those three together since in our last two proposals we completely ignore the distance to the object. We would need a calculation like distFactor * zoomFactor * scaleFactor, the scale and zoom are easy but how do we calculate the distance (we cannot use the code of the perspective camera since we do not have FOV)

JDihlmann: Some more thoughts distFactor * zoomFactor * scaleFactor shouldn't be a multiplication since the orthographic camera is able to be at the same position as the object (dist=0) and still render it. The same holds for the perspective camera 1 / scaleFOV could turn out to be a division by 0 if the distance to the object is 0. I think here are several issues at play and the whole function should be refactored. But this is out of my league.

JDihlmann: Okay I take everything back (long monologue). Per definition an orthographic camera does not care about the distance to an object. Therefore the render distance can only be varied with the zoom. I think what happened here is that the previous implementation before @nathanbryantdev committed #311 was the intended behavior of scaleFactor, since the documentation states: "scaleFactor={10} // If set (default: undefined), children will be scaled by this factor, and also by distance to a PerspectiveCamera if transform is set to false." Therefore similar to the perspective camera  if the scaleFactor 
 is set it should change by distance <=> zoom for an orthographic camera. What you want @nathanbryantdev is a new feature, such that the Html element changes with the scale of the corresponding mesh. I took me so long to figure that out because the storybook entry you provided  @nathanbryantdev seemed like as you are moving the object back and forth but you are adjusting the scale of the object. So @joshuaellis, @neftaly, @gsimone I would propose that we close this PR by changing the scaleFactor back to the original state by returning camera.zoom (keeping my implementation in the useFrame such that the scene really updates on zoom change) and additionally introduce a new feature that scales an Html element according to its parent scale in a new PR. After all changing the Htmls scale by guessing the distance to the camera is in my opinion the wrong approach it should be only dependent on group.current.getWorldScale.

joshuaellis: Thats interesting. This makes sense to me. What does everyone else think?

neftaly: Sounds good to me, as far as I understand it - good catch. If it gets the go-ahead it would be good to break this into 2 commits, e.g. "Fix scene updating on zoom change" and "Revert #311" (or #308).

JDihlmann: Maybe we could consider renaming scaleFactor to something like distanceFactor, because the name is indeed confusing.

JDihlmann: How do we proceed?

joshuaellis: If you haven't already, renaming makes sense, I wouldn't worry about the commit thing, we squash them all. Poke me when it's ready to be merged. Unless I've missed something?

JDihlmann:
@joshuaellis I renamed scaleFactor to distanceFactor, edited the documentation so people understand its related to the zoom property of the OrthopgraphicCamera and exchanged the orthographic story with a new one that uses the zoom and the distanceFactor. Anything else?

joshuaellis:
Are you happy to work on this afterwards? 

> additionally introduce a new feature that scales an Html element according to its parent scale in a new PR

joshuaellis (approved):

JDihlmann:
I am not really in the need of this particular feature, so maybe @nathanbryantdev could take a second shot at this? As a said before from my understanding the Html scale should be additionally multiplied with the group.current.scale (in world space) and the useFrame should update if group.current.scale changes. Overall I do not think that this feature should be a standard feature of drei and rather a fork, since it is heavily depended on the use case. There are a lot of open questions: Which dimension of group.current.scale should be used? The one aligned with the camera plane? The magnitude? What if scale is negative?

PR #321, opened 2021-03-06
feat: Controls can now receive a custom camera as prop

gsimone:
Demo 
https://codesandbox.io/s/r3f-multi-view-mdnhl?file=/src/Panel.js


gsimone:
Making stories for each control type would have been super time consuming and a lot of boilerplate code. If anybody wants to do that in the future, cool, but they are not really necessary.

joshuaellis (approved):
🎉  LGTM

PR #327, opened 2021-03-11
feat: Adds Bezier Lines Components

giulioz:
### Why

Currently you have to do bezier lines by hand, it would be convenient to have a component for that as well.

### What

Added two new components for bezier lines.

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


src/core/CubicBezierLine.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+type Props = Omit<LineProps, 'points'> & {
+  start: Vector3 | [number, number, number]
+  end: Vector3 | [number, number, number]
+  midA: Vector3 | [number, number, number]
+  midB: Vector3 | [number, number, number]
+  segments?: number
+}
+
+export const CubicBezierLine = React.forwardRef<Line2, Props>(function CubicBezierLine(
```
joshuaellis, line 14: Should Line2 not be Line? then you don't have to cast any

giulioz: Actually that's not the problem. The ref type should be the underlying Line2 from the Line component, not the Line component itself.

If I remove the `as any` cast I get this error:
```
Type 'Ref<ReactNode> | ((instance: Line2 | null) => void) | MutableRefObject<Line2 | null>' is not assignable to type 'Ref<Line2> | undefined'.
  Type 'RefObject<ReactNode>' is not assignable to type 'Ref<Line2> | undefined'.
    Type 'RefObject<ReactNode>' is not assignable to type 'RefObject<Line2>'.
      Type 'ReactNode' is not assignable to type 'Line2'.
        Type 'undefined' is not assignable to type 'Line2'.
```

joshuaellis (approved):
thanks for getting rid of `any`!

PR #330, opened 2021-03-16
[fix] Capture .tsx files & add terser

### LINKED ISSUE #329: How to correctly import / build the Html component as a dependencie
Hello everyone,

I was initially using the drei Html component in react-three-a11y but I replaced it with my own "fork" and it fixed my problem.
So if this "issue" don't seem related to drei feel free to close it.

react-three-a11y ( https://github.com/pmndrs/react-three-a11y/ ) is bundled through tsdx ( which use Rollup )

If you look at this component https://github.com/pmndrs/react-three-a11y/blob/flat_import/src/A11y.tsx 
it used to import Html this way 
`import { Html } from '@react-three/drei/web/Html'`
When doing it this way it seems that the cjs output will look like something like this 
`var Html = require('@react-three/drei/web/Html');`
The require require a file using import and therefore cause an error saying `SyntaxError: Cannot use import statement outside a module`

Because of that I tried importing it this way 
`import { Html } from '@react-three/drei';`
But if I do it, my cjs build goes from 5kb to 150kb as it seems to import all of drei
But at least this way I don't have any issue with remaining import in the file.

I tried my best with tsdx / rollup configs but I've find no way to fix this.
In case it's a problem caused by drei, we suggested that I should open this issue.

Again, builds, esm, cjs, etc.. not my cup of tea, so maybe It's just something that is not related to drei

Happy to provide more information that could help.

joshuaellis:
### Why

resolves #329 

### What

We weren't capturing .tsx files for`cjs` exports, only `.ts` files. This has now changed & we also use `tersrer` to minify the `cjs` code.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
     "rollup-plugin-multi-input": "^1.1.1",
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-size-snapshot": "^0.12.0",
+    "rollup-plugin-terser": "^7.0.2",
     "three": "0.125.0",
     "typescript": "^3.9.7",
     "use-asset": "^1.0.2"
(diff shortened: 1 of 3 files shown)


joshuaellis:
@AlaricBaraou this PR should fix your problems, there's a beta for v4 `drei` so I will release this on `@alpha` for you to test when you can. If it works, just let me know on #329 so I can push to master.

PR #336, opened 2021-03-23
feat: adaptive components

drcmda:
### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


gsimone:
1. how crazy would it be to let users use meshBounds on regression? So that you could simplify raycasting while regressing without totally removing it?
2. <AdaptiveMaterial materials={[<material1 />, <material2 />]} /> ?
3. <AdaptiveGeometry geometries={[<geometry1 />, <geometry2 />]} />

Starting to think wether we should have a generic <Adaptive abstraction that just gets regression and picks from an array of things to mount/unmount

nksaraf:
Is this something that can also be used for something like an LOD (level of detail) system for games based on the camera?

gsimone:
@nksaraf not really, these are components meant to work with the new performance regression system in react-three-fiber@v6. For what you need, you might take a look at the <Detailed /> component here, but a custom system might be necessary

joshuaellis:
@drcmda is this just missing the storybook entry?

joshuaellis (approved):

PR #338, opened 2021-03-24
fix reflector render loop

### LINKED ISSUE #333: Reflector gives "TypeError: gl.setRenderTarget is not a function" error on unmount
- `three` version: 0.124.0
- `react-three-fiber` version: 5.3.19
- `@react-three/drei` version: v3.11.2
- `node` version: v12.18.3
- `npm` (or `yarn`) version: npm v6.14.6
- `next` version: 10.0.9

### Problem description:

When navigating away from page containing a R3F canvas which contains a Reflector I receive an error "TypeError: gl.setRenderTarget is not a function" I believe this occurs when unmounting the ThreeJS component.

```Call Stack
Object.current
node_modules\@react-three\drei\core\Reflector.js (163:0)
renderGl
/_next/static/chunks/2.js (64560:38)
<unknown>
node_modules\react-three-fiber\web.js (103:0)
Map.forEach
<anonymous>
renderLoop
/_next/static/chunks/2.js (64589:9)
```

### Relevant code:

```js
    <Reflector
      resolution={512}
      args={[100, 500]}
      position={[0, 0, 150]}
      mirror={0.3}
      mixBlur={1}
      mixStrength={2}
      rotation={[-Math.PI / 2, 0, 0]}
      blur={[200, 50]}
      dispose={null}
    >
      {(Material, props) => (
        <Material metalness={0} map={normal} normalScale={[1, 1]} {...props} />
      )}
    </Reflector>
```

### Suggested solution:

It might be related to https://github.com/pmndrs/react-three-fiber/issues/722?

emmelleppi:
### Why
Fixes #333 

### What
It adds a control on the `gl` variable inside the reflector renderloop 

### Checklist

- [ ] Test -> https://codesandbox.io/s/drei-reflector-forked-87gu0
- [ ] Ready to be merged

### DIFF
--- a/src/core/Reflector.tsx
+++ b/src/core/Reflector.tsx
@@ -239,7 +239,7 @@ export const Reflector = React.forwardRef<Mesh, ReflectorProps>(
       if (mixBlur !== 0) {
         renderTargets.forEach((fbo) => {
           gl?.setRenderTarget?.(fbo)
-          gl?.state.buffers.depth.setMask(true)
+          gl?.state?.buffers?.depth?.setMask(true)
           gl?.render?.(scene, virtualCamera)
         })
       }
--- a/src/core/Reflector.tsx
+++ b/src/core/Reflector.tsx
@@ -230,21 +230,21 @@ export const Reflector = React.forwardRef<Mesh, ReflectorProps>(
     ])
 
     useFrame(({ gl }) => {
-      if (!(meshRef?.current && gl)) return
+      if (!meshRef?.current) return
       meshRef.current.visible = false
       beforeRender()
-      gl.setRenderTarget(fbo1)
-      gl.state.buffers.depth.setMask(true)
-      gl.render(scene, virtualCamera)
+      gl?.setRenderTarget?.(fbo1)
+      gl?.state.buffers.depth.setMask(true)
+      gl?.render?.(scene, virtualCamera)
       if (mixBlur !== 0) {
         renderTargets.forEach((fbo) => {
-          gl.setRenderTarget(fbo)
-          gl.state.buffers.depth.setMask(true)
-          gl.render(scene, virtualCamera)
+          gl?.setRenderTarget?.(fbo)
+          gl?.state.buffers.depth.setMask(true)
+          gl?.render?.(scene, virtualCamera)
         })
       }
       meshRef.current.visible = true
-      gl.setRenderTarget(null)
+      gl?.setRenderTarget?.(null)
     })
 
     return (
--- a/src/core/Reflector.tsx
+++ b/src/core/Reflector.tsx
@@ -234,7 +234,7 @@ export const Reflector = React.forwardRef<Mesh, ReflectorProps>(
       meshRef.current.visible = false
       beforeRender()
       gl?.setRenderTarget?.(fbo1)
-      gl?.state.buffers.depth.setMask(true)
+      gl?.state?.buffers?.depth?.setMask(true)
       gl?.render?.(scene, virtualCamera)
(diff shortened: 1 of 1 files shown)


gsimone:
Can you quickly check if this would also fix it? I assume it's losing the reference to the gl defined in the component, but using the one from useFrame would fix the same problem, maybe

```
useFrame(({ gl }) => {
  
})
```

emmelleppi:
> Can you quickly check if this would also fix it? I assume it's losing the reference to the gl defined in the component, but using the one from useFrame would fix the same problem, maybe
> 
> ```
> useFrame(({ gl }) => {
>   
> })
> ```

It seems that a simple control on the `gl` it doesn't work. 
I'm guessing, but most likely the problem is that when the Canvas is unmounted the Reflector renderloop is still processing.
In that case, the problem is not really with the Reflector, but with the handling of the useFrame callback, which should be stopped as soon as the canvas is unmounted. (Maybe? 🤷🏻‍♂️)

Anyway, I have now put a check on each function called by the renderer inside the renderloop and it should work.

joshuaellis:
> > Can you quickly check if this would also fix it? I assume it's losing the reference to the gl defined in the component, but using the one from useFrame would fix the same problem, maybe
> > ```
> > useFrame(({ gl }) => {
> >   
> > })
> > ```
> 
> It seems that a simple control on the `gl` it doesn't work.
> I'm guessing, but most likely the problem is that when the Canvas is unmounted the Reflector renderloop is still processing.
> In that case, the problem is not really with the Reflector, but with the handling of the useFrame callback, which should be stopped as soon as the canvas is unmounted. (Maybe? 🤷🏻‍♂️)
> 
> ~Anyway, I have now put a check on each function called by the renderer inside the renderloop and it should work.~

I wonder if this persists in v4 of `drei` where we're using v6 of `r3f`. If it is then we need to add some cleanups to cancel raf's, that would fix the issue in `useFrame`?

emmelleppi:
Ok now it works. the solution to the problem pretty much sucks: put controls everywhere.

I think we will have the same problem with v4. I'm noticing now that `cancelAnimationFrame` is never used in r3f, am I wrong?

joshuaellis:
I think we should wait to see if the error can be resolved in `r3f` first before merging the change?

emmelleppi:
> I think we should wait to see if the error can be resolved in `r3f` first before merging the change?

seems reasonable to me 👍

joshuaellis:
@emmelleppi Can you test this in the latest `beta`? I've updated `r3f` to the latest alpha I'm hoping it's fixed... If it is, this fix should just wait till v4

emmelleppi:
> @emmelleppi Can you test this in the latest `beta`? I've updated `r3f` to the latest alpha I'm hoping it's fixed... If it is, this fix should just wait till v4

Ok, the error is no longer thrown, which is good. However, if I now try to continuously mount and unmount the Canvas (by clicking over and over again, not immediately) the performance drops.
I have no idea what this might be due to 

https://codesandbox.io/s/drei-reflector-forked-z0895?file=/src/App.js

joshuaellis:
Ah sorry it’s published under @react-three/fiber now, the alpha number is 14 I think? That sandbox has alpha 11 which I think still had some issues in it regarding disposal on unmount.

emmelleppi:
@joshuaellis sorry for the late reply, it works!

https://codesandbox.io/s/drei-reflector-forked-z0895?file=/src/App.js

joshuaellis:
@emmelleppi great news, can we close this PR then? It feels like we don't need it anymore if I understand?

emmelleppi:
yep!

PR #339, opened 2021-03-24
Fix: OrbitControls `dispose` is null

susiwen8:
### Why
In [`react-three-fiber@renderer.ts#L315`](https://github.com/pmndrs/react-three-fiber/blob/v6/release/packages/fiber/src/core/renderer.ts#L315) will reset `primitive` `dispose` to `null` which throw error
Screen Shot 2021-03-24 at 19 38 32

### What

Add `dispose={undefined}` for `primitive`
Inspire by [DeviceOrientationControls](https://github.com/pmndrs/drei/blob/beta/src/core/DeviceOrientationControls.tsx#L31)

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/OrbitControls.tsx
+++ b/src/core/OrbitControls.tsx
@@ -41,6 +41,8 @@ export const OrbitControls = React.forwardRef<OrbitControlsImpl, OrbitControlsPr
       }
     }, [controls, invalidate])
 
-    return <primitive ref={ref} object={controls} enableDampening={enableDampening} {...restProps} />
+    return (
+      <primitive ref={ref} dispose={undefined} object={controls} enableDampening={enableDampening} {...restProps} />
+    )
   }
 )


joshuaellis:
@drcmda you were chatting about this in discord, do you want to fix this in `r3f` or shall we fix it in `drei`?

susiwen8:
Looks like `dispose={undefined}` is for concurrent mode 47fd6a4fc7b0e97fc7b4ed5d74fd0dbdeb0684fc

drcmda:
this is a r3f thing, it needs to be solved there.


 in v5 there was some hack that allowed primitives, which normally should never dispose, to do it anyway if dispose===undefined. but you couldn't dispose yourself bc r3f would overwrite the dispose method.

i decided to remove this. it doesn't touch the method, it will never dispose on its own. but im thinking about giving primitive an optional extra flag, like `local`, that that it can clear out local objects.

basically this:

### global

```jsx
const global = new THREE.Thing()

function Foo() {
  // This will not dispose on unmount
  return <primitive object={global} />
```

### local

```jsx
function Foo() {
  const [local] = useState(() => new THREE.Thing())
  // This will dispose on unmount
  return <primitive object={local} local />
```

what do you think?

joshuaellis:
IMO this is confusing behaviour having this component dispose in different ways depending on whether you put your initialized class in the component or not. Is there a reason you want to remove the disposing behaviour we have in v5? 

Is it because `dispose` overwrites the instance's method of disposing? If so we can change dispose as a `boolean` and pretending it the internal `r3f` object?

drcmda:
v5 did that, it overwrote the dispose method. v6 doesn't. but there's still the open question: how should primitives deal with local data. it should imo dispose of it on mount because the data will be 100% gone. that is different from global data which persists. 

> Is it because dispose overwrites the instance's method of disposing? If so we can change dispose as a boolean and pretending it the internal r3f object?

that's the v6 proposal more or less.

```jsx
    if (type === 'primitive') {
      if (props.object === undefined) throw `Primitives without 'object' are invalid!`
      const object = props.object as Instance
      const dispose = (props.dispose as unknown) as boolean
      // Remove object from props
      delete props.object
      delete props.dispose
      instance = prepare<Instance>(object, { root, instance: true, dispose })
```

PR #337, opened 2021-03-23
feat: hide <Html /> when occluded by an object 

drcmda:
### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

PR #341, opened 2021-03-24
feat: Add hidden raycasting check to <Html />

### LINKED ISSUE #340: Add hidden check to <Html/>

Hey, I want to suggest a feature to check if the <Html> element's position is hidden and hide it if it is.

### Suggested implementation:

Could be solved by providing an additional property which enables checking if the <Html> object is hidden. The check itself can be realized with raycasting and a distance check.

I wrote an implementation in my fork I'd like to PR.

j-era:
### Why

Check visibility by distance
Resolves #340 
### What

Add a 'checkDepth' property which can be set to `true` to activate testing if the HTML object is visible from the camera. Check is handled by raycasting.

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


gsimone:
Hey 👋 did you see https://github.com/pmndrs/drei/pull/337 ? I did the same thing but using an explicit occluder object to save on raycaster checks. What do you think?

j-era:
Hi @gsimone, no unfortunately I didn't. The check I wrote tests if the <Html /> position is visible in front of any geomtry hit in the scene. Attached to the same object it has to be positioned manually to ensure it is "outside" of the geometry. 
But the option to limit the check to one (or more?) occluder object(s) and fading would be a good addition (instead of raycasting all scene children). Maybe those are individual features?

gsimone:
The "all scene" check could be implemented via the occluder API, like this

```
const { scene } = useThree()

<Html occluder={scene.children} /> 
```

The fading didn't really work so I'd probably just drop it, but I'd prefer defaulting to explicitely passing the object to <Html />, to avoid performance footguns.

I'll close #337  in favor of this, anyway ✅

j-era:
Ok fine :) I'll update the implementation.

drcmda:
just like @gsimone's pr, having this in would be wonderful. most requested feature for html. occluder sounds pretty clever, but in whichever way - it could be finetuned little by little.

drcmda:
btw @gsimone @joshuaellis  it would be pretty crazy if drei had a quick but not so precise bounds only raycaster. or something configurable.

```jsx
import { Raycaster } from 'three-stdlib'

const ray = new RayCaster({ method: 'AABB' | 'Convex' | 'Vertex' ... })
ray.intersect(...)
```

if we had this, r3f could use that by default. it could switch to low-res piercing on regress (movement) and things like `<Html/>` could pierce using the lowest setting. you think that would be possible?

joshuaellis:
@gsimone i'm gonna leave this to you 👌🏼

gsimone:
I'll review as soon as possible 🙏

joshuaellis:
@j-era how can I test this?

j-era:
sry @joshuaellis, no time for nothing ;) i've added a storybook example you can test with -> https://drei-git-fork-j-era-feat-340-add-hidden-check-to-html-pmndrs.vercel.app/?path=/story/misc-html--html-raycast-occluder-st

joshuaellis (approved):
This is less code than I thought it would be, but looks like it works great 🙏🏼

drcmda:
nice! let's get it in if it works :-)

drcmda:
Could we amend the docs a little? I think it's a little confusing, the story book mentions a occluder, and docs don't but do mention checkDepth instead. I think it's not entirely clear how this feature works and how it can be enabled.

drcmda:
made a small test here https://codesandbox.io/s/mixing-html-and-webgl-forked-hpqd3?file=/src/Html.tsx

it changes the prop to "occlude" and allows it to be a boolean. a true would occlude the whole scene. it also sets opacity instead of display, which would allow the user to animate it. i guess since we haven't documented or published the feature yet we could still just count it as a patch.

nwaxiomatic:
Hey, This feature is really cool but it seems to drop the frame rate of DRASTICALLY when in use. Anyone else have this problem?

PR #342, opened 2021-03-25
chore: update all our deps

dbismut:
### Why

Removes react-spring dependency from the package, knowing that this react-spring/web is only used for the <Loader />` component.

### What

I replaced react-spring `useTransition` and AnimatedValue transitions with a manual interpolation (which shouldn't trigger more renders though).

I'm not sure why the `initialState` needs to be a function tbh but I kept it that way.

At the moment I'm having issues with triggering the unmount animation: it seems like `<Suspense>` unmounts the Loader component used as a fallback **before** it has the chances to reach `100%`. 

### Checklist

- [ ] Documentation updated (not needed)
- [ ] Storybook entry added (not needed: Storybook uses custom Loader it seems)
- [ ] Ready to be merged


gsimone:
That component doesn't have a story, could you drop a codesandbox in the PR?

dbismut:
You mean story right? Not codesandbox? Is it ok if I use the useProgress story to show it?

gsimone:
Oh, if you can show it in a story, even betterino

.storybook/stories/useProgress.stories.tsx
```diff
@@ -46,11 +46,27 @@ function LoadExtras() {
...
+      fallback={
+        <Html>
+          <Loader />
+        </Html>
+      }
+    >
+      {boolean('Load extras', false) ? <LoadExtras /> : <Helmet />}
+    </React.Suspense>
+  )
+}
```
gsimone, line 72: Tiny nitpick, this could be the default and the other one could be "With custom Loader"

dbismut:
@gsimone pushed here: https://drei-git-chore-remove-react-spring-deps-for-loader-pmndrs.vercel.app/?path=/story/misc-useprogress--with-out-of-the-box-loader

But doesn't seem to work, the loader doesn't show. In dev mode everything works. After investigating on the built version of storybook, this is the log I get from `useProgress` (the variable `shown` is the delayed derived state from `active`):

image

As you can see, `active` never toggles to `true`, so no Loader is shown 🤷‍♂️

joshuaellis:
I fixed the issue @dbismut was having just by ensuring we set `active` as true in `onProgress`. I think there might be an issue with `DefaultLoadingManager` but I don't wanna go down that rabbit hole when we have an easy fix.

I then got carried away and did the monotonous task of updating all our deps to bring them into line, e.g lots of our babel, could now go 🚀  So a general all around tidy up to be honest.

PR #343, opened 2021-03-26
fix: clear cache when Environment is unmounted

### LINKED ISSUE #332: Environment component only works on first mount, then stays black
- `three` version: 0.126.1
- `react-three-fiber` version: 5.3.19
- `@react-three/drei` version: 3.11.1
- `node` version: 10.20.1 (CodeSandbox default?)
- `npm` (or `yarn`) version: CodeSandbox

### Problem description:

When using `<Environment />` it seems to only work when mounting a `<Canvas />` for the first time, then it renders a black screen, no errors. My usecase is that I'm mounting it on specific routes, so the client-side routing mounts and unmounts the three component.

### Relevant code:

I made a CodeSandbox reproducing the problem. The official storybook has the same problem when navigating to Environment for the second time.
https://codesandbox.io/s/snowy-cherry-ve9fl?file=/src/index.js

### Suggested solution:

I did some digging and logged some stuff, but couldn't find anything suspicious.

susiwen8:
### Why

Thanks to @drcmda  https://github.com/pmndrs/drei/issues/332#issuecomment-808113558 it should clear cache when Environment is unmounted

Close #332 

### What

Clear cache at useLayoutEffect clean up function.

### Checklist

- [] Documentation updated
- [] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "rollup-plugin-terser": "^7.0.2",
     "three": "^0.126.1",
     "typescript": "^4.2.3",
-    "use-asset": "^1.0.2"
+    "use-asset": "^1.0.4"
   },
   "peerDependencies": {
     "@react-three/fiber": ">=6.0",
--- a/src/core/Environment.tsx
+++ b/src/core/Environment.tsx
@@ -76,6 +76,7 @@ export function Environment({
       }
     }
     return () => {
+      useAsset.clear()
       if (scene) {
         scene.environment = oldenv
         scene.background = oldbg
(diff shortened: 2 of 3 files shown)


drcmda:
imo this can't be something that drei or r3f calls. it will plain delete all cached assets systemwide. textures, modules, images, promises, ... all gone, just because some canvas unmounts.

the issue isn't env unmounting, nothing will happen to the assets, the black screen only came about when he deleted the canvas. but even that shouldn't clear cache.

drcmda:
i just see that useLoader only has preload: https://github.com/pmndrs/react-three-fiber/blob/master/src/hooks.ts#L138
it should have clear and peek as well. but i wouldnt bother with this in v5. v6 should have it. fetching useAsset from `use-asset` and calling clear in there is probably good enough for now.

PR #348, opened 2021-03-30
fix: ssr issue because no flat import meshopt

RenaudRohlinger:
### Why

Fix ssr by doing flat import in useGLTF

- [ x ] Ready to be merged


joshuaellis (approved):
Thanks for the fix! I'll have a did into `cjs` resolving at some point..... Because with `drei` you can do `import { Html } from '@react-three/drei/web/Html` and it will work under next etc....

Issue #169, opened 2020-10-22
React concurrent mode breaks custom <OrbitControls /> properties


johnmarinelli:
csb: https://9s00o.csb.app/

for some reason, using `concurrent: true` in the `<Canvas />` will make `enableZoom={false}` on `<OrbitControls />` not work.  concurrent mode probably breaks other things too, but this is just one that I found.

drcmda:
wa-what? :-D

It’s  is defined as:

```jsx
export const OrbitControls = forwardRef((props: OrbitControls = { enableDamping: true }, ref) => {
  const { camera, gl, invalidate } = useThree()
  const controls = useMemo(() => new OrbitControlsImpl(camera, gl.domElement), [camera, gl])
  ...
  return <primitive dispose={null} object={controls} ref={ref} enableDamping {...props} />
```

i dont quite see how concurrent mode could affect the props being spread over it. would you please copy the orbit-helpers code and use it in your file to track down what it is? also better go to the latest version 5.x.

hmans:
I'm also seeing strange behavior around MapControls (which, as I understand, is just [OrbitControls with different defaults](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/controls/OrbitControls.js#L1212-L1224).)

I'm using the default panning.

With `concurrent` disabled, panning the camera also moves the control's target, as it should be.

With `concurrent` **enabled**, panning the camera no longer moves the control's target, and the camera will instead keep looking at [0, 0, 0].

hmans:
To make things even more weird, the bug completely disappears when I run the same code [in production mode](https://hmans-sky.netlify.app/).

@johnmarinelli, can you check if this is also the case for you?

hmans:
I think I've figured it out. It's possible that this only occurs when you create a custom camera on top of the one that <Canvas> will create by default, because this will end up invoking the `useMemo` call twice, without calling `.dispose()` on the controls instance that was created first.

My first attempt at fixing it was removing the `dispose={false}` in the MapControls implementation, but that gave me errors. So instead I very haphazardly rebuilt the component to do everything through useEffect instead:

~~~ js
import React, { forwardRef, useEffect, useState } from "react"
import { useFrame, useThree } from "react-three-fiber"
import { MapControls as MapControlsImpl } from "three/examples/jsm/controls/OrbitControls"

export const MapControls = forwardRef(
  (props = { enableDamping: true }, ref) => {
    const [controls, setControls] = useState()
[...]

PR #354, opened 2021-04-02
fix: stop registering dom event handlers in useState in MapControls

### LINKED ISSUE #351: MapControls doesn't work as expected
- `three` version: 0.126.1
- `@react-three/fiber` version: 6.0.0
- `node` version: 12.18.3
- `npm` (or `yarn`) version: 1.22.5
- browser version: `chrome@latest`/`edge@latest`

### Problem description:
When visiting https://drei.pmnd.rs/?path=/story/controls-mapcontrols--map-controls-scene-st everything works as expected.

When I tried running drei's storybook locally, I get this:
MapControls

hasparus:
### Why

MapControls init and register handlers twice — a bit broken.

A similar problem is illustrated in this CodeSandbox:
https://codesandbox.io/s/r3f-contact-shadow-forked-drey6?file=/src/index.js

Closes #351.

### What

I fixed this on `MapControls` component level by moving side effects to useEffect.

### Possible further work?

We could do a more direct fix for the root cause — yeet side effects out of OrbitControls — it's a breaking change but three-stdlib doesn't have to be backwards compatible with Three, innit? Side effects in class constructors are 🤮

See 
- https://github.com/mrdoob/three.js/issues/20575
- https://github.com/pmndrs/drei/issues/169

### Checklist

- [x] Ready to be merged

#### Before

https://user-images.githubusercontent.com/15332326/113441515-bac67580-93ee-11eb-9d22-c55e4b7df387.mp4

#### After

https://user-images.githubusercontent.com/15332326/113441531-c023c000-93ee-11eb-8458-b5aa27f00747.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "lodash.pick": "^4.4.0",
     "react-merge-refs": "^1.0.0",
     "stats.js": "^0.17.0",
-    "three-stdlib": "^1.1.3",
+    "three-stdlib": "^1.1.4",
     "troika-three-text": "^0.38.1",
     "use-asset": "^1.0.4",
     "utility-types": "^3.10.0",
--- a/src/core/MapControls.tsx
+++ b/src/core/MapControls.tsx
@@ -17,12 +17,10 @@ declare global {
 
 export const MapControls = React.forwardRef<MapControlsImpl, MapControls>((props = { enableDamping: true }, ref) => {
   const { camera, ...rest } = props
-  const { invalidate, domElement, defaultCamera } = useThree(({ invalidate, camera, gl }) => ({
-    invalidate,
-    defaultCamera: camera,
-    domElement: gl.domElement,
-  }))
-
+  const invalidate = useThree(({ invalidate }) => invalidate);
+  const defaultCamera = useThree(({ camera }) => camera);
+  const domElement = useThree(({ gl }) => gl.domElement);
+  
   const explCamera = camera || defaultCamera
   const [controls] = React.useState(() => new MapControlsImpl(explCamera))
 
--- a/src/core/MapControls.tsx
+++ b/src/core/MapControls.tsx
@@ -17,17 +17,26 @@ declare global {
 
 export const MapControls = React.forwardRef<MapControlsImpl, MapControls>((props = { enableDamping: true }, ref) => {
   const { camera, ...rest } = props
-  const invalidate = useThree(({ invalidate }) => invalidate)
-  const defaultCamera = useThree(({ camera }) => camera)
-  const gl = useThree(({ gl }) => gl)
+  const { invalidate, domElement, defaultCamera } = useThree(({ invalidate, camera, gl }) => ({
+    invalidate,
+    defaultCamera: camera,
+    domElement: gl.domElement,
+  }))
+
   const explCamera = camera || defaultCamera
-  const [controls] = React.useState(() => new MapControlsImpl(explCamera, gl.domElement))
+  const [controls] = React.useState(() => new MapControlsImpl(explCamera))
+
+  React.useEffect(() => {
+    controls.connect(domElement)
+    controls.addEventListener('change', invalidate)
+
+    return () => {
+      controls.dispose()
+      controls.removeEventListener('change', invalidate)
+    }
+  }, [controls, invalidate, domElement])
 
   useFrame(() => controls?.update())
-  React.useEffect(() => {
-    controls?.addEventListener?.('change', invalidate)
-    return () => controls?.removeEventListener?.('change', invalidate)
-  }, [controls, invalidate])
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -72,7 +72,7 @@
     "lodash.pick": "^4.4.0",
     "react-merge-refs": "^1.0.0",
     "stats.js": "^0.17.0",
-    "three-stdlib": "^1.1.3",
+    "three-stdlib": "^1.1.4",
```
hasparus, line 75: I fixed types in 1.1.4 to allow calling MapControlsImpl without domElement.

src/core/MapControls.tsx
```diff
@@ -17,17 +17,26 @@ declare global {
...
+    controls.connect(domElement)
+    controls.addEventListener('change', invalidate)
+
+    return () => {
+      controls.dispose()
+      controls.removeEventListener('change', invalidate)
+    }
+  }, [controls, invalidate, domElement])
+
+  useFrame(() => controls?.update())
```
hasparus, line 39:
```suggestion
  useFrame(() => controls.update())
```

joshuaellis:
@hasparus if you're interested in kicking those side effects out I think @drcmda would be very happy 👀 there's an issue in `three-stdlib` that basically says what you wanna do if I'm correct – https://github.com/pmndrs/three-stdlib/issues/22

joshuaellis:
And no stdlib does not have to be backwards compatible with three. Just need to be careful when we change files that we turn them into Typescript at the same time so we have our own unique `.d.ts` files.

hasparus:
> @hasparus if you're interested in kicking those side effects out I think @drcmda would be very happy eyes there's an issue in three-stdlib that basically says what you wanna do if I'm correct – pmndrs/three-stdlib#22

I won't manage to do this this weekend, **and** even if we do it, I think the change from this PR should be merged anyway (as `.connect` inside of `useEffect` would be the only way to register events).

src/core/MapControls.tsx
```diff
@@ -17,17 +17,26 @@ declare global {
 
 export const MapControls = React.forwardRef<MapControlsImpl, MapControls>((props = { enableDamping: true }, ref) => {
   const { camera, ...rest } = props
-  const invalidate = useThree(({ invalidate }) => invalidate)
-  const defaultCamera = useThree(({ camera }) => camera)
-  const gl = useThree(({ gl }) => gl)
+  const { invalidate, domElement, defaultCamera } = useThree(({ invalidate, camera, gl }) => ({
```
joshuaellis, line 20: This is a no go, we don't have a comparison function: https://github.com/pmndrs/zustand#selecting-multiple-state-slices

joshuaellis (changes requested):

src/core/MapControls.tsx
hasparus: Oh, I forgot you have to add `shallow`. Thanks! I'll go back to multiple `useThree` calls.

hasparus: Should be fixed now. Should I do change anything else here?

src/core/MapControls.tsx
```diff
@@ -17,17 +17,24 @@ declare global {
...
-  const invalidate = useThree(({ invalidate }) => invalidate)
-  const defaultCamera = useThree(({ camera }) => camera)
-  const gl = useThree(({ gl }) => gl)
+  const invalidate = useThree(({ invalidate }) => invalidate);
+  const defaultCamera = useThree(({ camera }) => camera);
+  const domElement = useThree(({ gl }) => gl.domElement);
+  
   const explCamera = camera || defaultCamera
-  const [controls] = React.useState(() => new MapControlsImpl(explCamera, gl.domElement))
+  const [controls] = React.useState(() => new MapControlsImpl(explCamera))
```
GuskiS, line 25: While we are at it, there is another bug - controls camera is never updated in case it changes.
Meaning, if you do something like this (see `makeDefault`):
```ts
  const camera = useRef();

  return (
    <>
      <MapControls camera={camera.current} />
      <OrthographicCamera ref={camera} makeDefault {...CAMERA_PROPS} />
    </>
  );
```
change will be triggered by:
```ts
const defaultCamera = useThree(({ camera }) => camera);
```
however, `controls` will still use old `explCamera`.

Testing it locally, this seems to fix the problem:
```ts
  const [controls] = React.useState(() => new MapControlsImpl(explCamera));
  controls.object = explCamera;
```

not sure if that is the best way, tho.

joshuaellis: We can swap to use useMemo which IMO would be better :)

hasparus: Changed `useState` to `useMemo`.

joshuaellis (approved):
Thank you for the contribution! ⭐

hasparus:
Thanks!

PR #363, opened 2021-04-09
fix(types): add types to enviroment preset

SaraVieira:
### Why

When I tried to use the Stage and pass a different `environment` it just said string and I had to go to GH to fid out what environments existed

### What

Added types to it and imported it

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


joshuaellis (approved):
Thanks for the contribution ☀️

Issue #358, opened 2021-04-07
Gizmo helpers not showing up.


zeeshanali33096:
Hello, 

I am trying to use the Gizmo helper but they don't show up. Can you please tell me where am I going wrong?

```js
import { GizmoHelper, GizmoViewcube, OrbitControls } from "@react-three/drei";
import { OrbitControls as OrbitControlsImpl } from "three-stdlib";   
// Typescript would not let me have OrbitControls from @react-three/drei as type for the controlRef 

import { useRef } from "react";

import { Vector3 } from "three";

const CameraComponentOrbit = () => {
  const controlsRef = useRef<OrbitControlsImpl>(null); 
  return (
    <>
      <OrbitControls ref={controlsRef} />

      <GizmoHelper
        alignment="top-right"
        margin={[50, 50]}
        onTarget={() => controlsRef?.current?.target0 as Vector3}
        onUpdate={() => {
          if (controlsRef.current) {
            controlsRef.current.update();
          }
        }}
      >
        <GizmoViewcube />
      </GizmoHelper>
    </>
  );
};

export default CameraComponentOrbit;
```

I already have a default Perspective Camera included.

My package versions: 

`"@react-three/cannon": "^1.0.0",
    "@react-three/drei": "^4.0.3",
    "@react-three/fiber": "^6.0.8",
    "three": "^0.127.0",
    "zustand": "^3.4.0"
`

joshuaellis:
Have you had a look at the storybook example we have? 

https://drei.pmnd.rs/?path=/story/gizmos-gizmoviewcube--default-story

If you can create the example in a sandbox, that would be really helpful to debug.

zeeshanali33096:
I did go through the storybook [example](https://drei.pmnd.rs/?path=/story/gizmos-gizmoviewcube--default-story). 

At line 21, 

`const controlsRef = React.useRef<OrbitControls>()`

This is where typescript messes up. I tried adding a `//@ts-ignore` but no change, the Gizmo doesn't show up.

Here is the [sandbox](https://codesandbox.io/s/billowing-wood-ovs0m)

joshuaellis:
Well, I'm stumped – @w3dot0 got any thoughts? 😓

w3dot0:
@zeeshanali33096  For some reason attaching the background color breaks the Gizmo Display, @joshuaellis maybe a render order issues? I'm still digging, but if you comment that line out you can see the cube.
[...]

PR #370, opened 2021-04-19
#358 fix bg color and gizmo

zeeshanali33096:
### Why

#358 Background color attached to the scene does not render the `GizmoHelper`.

### What

Added a check in the `GizmoHelper` for any background attached in the scene. If found, it removes the background from the actual scene and adds it to the virtual scene. The background for the actual scene is restored when the `GizmoHelper` is unmounted.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

Pending - Background texture attached to the scene overrides the gizmo's material.

### DIFF
--- a/src/core/GizmoHelper.tsx
+++ b/src/core/GizmoHelper.tsx
@@ -1,6 +1,19 @@
 import * as React from 'react'
 import { createPortal, useFrame, useThree } from '@react-three/fiber'
-import { Camera, Group, Intersection, Matrix4, Object3D, Quaternion, Raycaster, Scene, Vector3 } from 'three'
+import {
+  Camera,
+  Color,
+  Group,
+  Intersection,
+  Matrix4,
+  Object3D,
+  Quaternion,
+  Raycaster,
+  Scene,
+  Texture,
+  Vector3,
+  WebGLCubeRenderTarget,
+} from 'three'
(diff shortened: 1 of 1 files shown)


joshuaellis:
Would you be able explain why you've done what you've done for someone as tired as I? 😅

joshuaellis:
@gsimone or @drcmda it'd be great to get your impression on this?

zeeshanali33096:
@joshuaellis @gsimone 
As @drcmda [pointed](https://github.com/pmndrs/react-three-fiber/discussions/1260#discussioncomment-631965) that was a render behaviour. The other implementation could be rendering the gizmo in a separate node (outside of the #root) similar to what we have for [drei/Stats](https://github.com/pmndrs/drei#stats). This way we can get rid of the portal and it will support color as well as background texture.

joshuaellis:
Wouldn't that require a new canvas?

zeeshanali33096:
> Wouldn't that require a new canvas?

yep that would. Currently even applying a post processing cause problems with the depth and the gizmos dont show up.

joshuaellis:
Would be able to make a code sandbox implementation idea so we can review it clearly please? This might be the only way... I'm very conscious of performance overhead...

PR #375, opened 2021-04-25
feat(ssr-script): added script to test ssr

RenaudRohlinger:
I made two new scripts to test if the Drei will crash in SSR:

test-ssr:build-server will generate a minimal react starter that includes drei in it 
test-ssr:start will try to start the previous generated build

For example once we have generated the code with test-ssr:build-server, if we do test-ssr:start it will try to start drei in SSR:

image


saurabhiam:
https://github.com/pmndrs/drei/issues/376 Facing issue with SSR

joshuaellis:
I've done a few house keeping changes to this, but nothing changing what work you've done. TBH, I want to find a way to automate this test, it'll never be ran before we push code (sorry, but it is true) and I do think what you've done is important. Just need to find a way to automate it. Maybe we avoid running the nodemon server and just do an RTL test for it. Or would `ReactDOMServer.renderToString` throw the error? It should I guess....?

joshuaellis (changes requested):
See above comment ⭐

saurabhiam:
> I've done a few house keeping changes to this, but nothing changing what work you've done. TBH, I want to find a way to automate this test, it'll never be ran before we push code (sorry, but it is true) and I do think what you've done is important. Just need to find a way to automate it. Maybe we avoid running the nodemon server and just do an RTL test for it. Or would `ReactDOMServer.renderToString` throw the error? It should I guess....?

Run it on Husky pre-commit hook?

RenaudRohlinger:
@joshuaellis Thanks! I totally agree with your point. I reduced it to the fewest amount of code possible. Do you think it would be possible to run a Github Action that runs the script "test:ssr-start" and check if nodemon compiles or throws an error?

From my tests, it runs the build + compile test in less than 2 seconds so it would not be too much annoying to add this one to the pipeline.

If we achieve to automate this test I'd be happy to add this to three-stdlib too as it's often the one troublesome.

joshuaellis:
Nodemon is a process runner, I dont think we should run it in an action? Otherwise *if* it works, it'll keep on going forever...
I think we need to write a small jest test to test. Did you look if calling `renderToString` fails if document is apparent?

Issue #377, opened 2021-04-25
Actions are not set when calling useAnimations


ekaradon:
- `three` version: 0.127.0
- `@react-three/fiber` version: 6.0.16
- `@react-three/drei` version: 4.2.0
- `node` version: v15.9.0
-  `yarn` version: 1.22.10
-  `typescript` version: 4.1.2
- `tsc` version: Version 4.2.4 

### Problem description:

The actions are never set when using `useAnimations`, they are just initialized to null.
![failing to set actions](https://i.imgur.com/lAXvXLq.png)

### Relevant code:
https://codesandbox.io/s/priceless-visvesvaraya-08tvs?file=/src/App.tsx

### Suggested solution:
I see two problems here:
 - A typescript error
 - A bug on the set

For the error:
![typescript error](https://i.imgur.com/NXfJQer.png)
Here is what I propose:

```typescript

// those come and replace the old definition
type Api<T extends string, U = { [K in T]?: AnimationAction }> = {
  ref: React.MutableRefObject<Object3D | undefined | null>
  clips: AnimationClip[]
  mixer: AnimationMixer
  names: T[]
  actions: U
}
type ExtendedAnimationClip<T extends String> = AnimationClip & { name?: T }

...

// the new signature of the hook
export function useAnimations<T extends string>(
  clips: ExtendedAnimationClip<T>[],
  root?: React.MutableRefObject<Object3D | undefined | null>,
): Api<T> {


...

// the parameter where the typescript error happened
      clips.forEach((clip: ExtendedAnimationClip<T>) => {

```

For the bug itself, I suggest just surrounding the assignment of brackets. 

```typescript

      clips.forEach((clip: ExtendedAnimationClip<T>) => {
        api.actions[clip.name] = mixer.clipAction(clip, currentRoot)
      })

```

I don't explain why or how this happen but it seems that not surrounding an assignment of brackets can lead in some case to the assignment not taking place at all. If you have any idea why I'm interested.

I can also do a pull request if you want.

drcmda:
it needs to render out the elements first for them to be available in the mixer, this can only happen in useEffect/useLayoutEffect, that's why the actions are first null and then filled, this is not a bug. the brackets don't change anything, but i guess hot-reload made it look like they're now filled? we have multiple examples in the docs pages that use animations - the system is definitively working.

ekaradon:
[...]

PR #378, opened 2021-04-26
fix: useAnimations was not assigning the actions

ekaradon:
### Why

#377 useAnimations did not set the actions.

### What

Retyping the hook signature because as soon as I was starting to use it I got an error. Also surrouding the assignment of actions by curly brackets.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

I don't know about this checklist so I'm just going to let them uncheck for now.


joshuaellis:
It'd be great if you could show us how we can test this fix works?

drcmda:
actions are set in useLayoutEffect. you can't do:

```jsx
const actions = useAnimations(...)
actions.foo.play()
```

because that's a side effect. play has to be in either useEffect or useLayoutEffect and by that time the actions will be filled.

ekaradon:
I intended to use the animation within a `useEffect` liked showed in the sandbox but before doing it I wanted to know which animations were available from my model, and none showed up like demonstrated in the screenshot of the console in the initial issue?

However, to my surprise, when I tried to demonstrate the issue a bit more by tieding the action inside a useEffect, the animation was played. I don't really get it because from my point of view, a useEffect call involves that the parent component was called? It seems that I had a bad understanding of useEffect without any dependencies. It may be called even when the component is not called? Or is it specific to the WebGL context? I don't know.

Still, this pull request solves the `null` values showed up in the console in the component itself in my project. I tried to setup a codesandbox to show the difference in behavior but I got weird errors from codesandbox complaining about unknown prototype... 
https://codesandbox.io/s/laughing-roentgen-342z9?file=/src/useAnimationsFix.ts
If someone knows why I can't use Object.Values here, maybe I could show the difference between those two. 

And it also solves an error showing up for the typescript part?

src/core/useAnimations.tsx
```diff
@@ -2,36 +2,48 @@ import * as React from 'react'
...
+    actions: clips.map((clip) => clip.name).reduce((prev, name: T) => ({ ...prev, [name]: null }), {}),
+    names: clips.map((clip) => clip.name),
+    mixer,
+  })
   useFrame((state, delta) => mixer.update(delta))
   React.useEffect(() => {
     const currentRoot = actualRef.current
     if (currentRoot) {
-      clips.forEach((clip) => (api.actions[clip.name] = mixer.clipAction(clip, currentRoot)))
+      setApi((oldApi) => ({
```
joshuaellis, line 33: How come you've changed this? I'm a bit out of the loop.

ekaradon: Well, I don't really know where the discussion should happen. In the initial issue or in the pull request? But to sum up the discussion over there:
 
 - the API is not returned in the render loop with the actions set (aka every actions are null)
 - I thought that wrapping it with curly brackets fixed the issue but in fact it was just the hot-reload which was fixing the issue
 - However, within the effect loop (aka `useEffect` or `useLayoutEffect`), because we mutate actions directly within `useAnimations`, it's filled and we can use it

So, because it got me confused I suggested here to change the approach for a state one, where the actions could be seen from the component. And I pushed a fixup showing what it could look like for an example.

From my point of view, it makes more sense. But I can see problems with this approach:
 - We add another rendering (the one giving us the actions set)
 - Every bit of code using this hook could now run with the actions not set at first

This means that when we write something like (took from the story book): 
```javascript
  React.useEffect(() => {
    actions[selectedAction].reset().fadeIn(blendDuration).play()
    return () => void actions[selectedAction].fadeOut(blendDuration)
  }, [actions, selectedAction, blendDuration])
```
It would fail because the initial actions is not updated anymore. 
We could also mix both of these strategies (mutate the old one and update the state) so we don't have any breaking change.

drcmda:
Imo null is correct, it cannot have anything else because it has not yet rendered the nodes out, which are critical for the animations. If you're or changes that then I wonder from where these entries come from? You can already see the animation keys if you log them out still. Additionally gltfjx has a typescript option which makes all contents type safe, go here: gltf.pmnd.rs and drop in your model, click TS

ekaradon:
What I suggest here is not to change the fact that it is initially null, just to update the api (and so re render) once those actions are known? I'm all about immutability, const and frozen data and I really just started this week to use `react-three-fiber` so I may be a bit biased. And I may lack of background in this matter. I just think that going through mutations is going against the wind when using React.
 
I knew about `gltf.pmnd.rs` but I cannot successfully use it since all my gltf files are referencing external files like `scene.bin` and it crash. However, I have been using `npx gltfjsx model.gltf` and it has indeed an option (-t) for typescript, thanks for the tip, it's neat.

joshuaellis:
@ekaradon this PR has changed quite a lot from the original one, can you give us an update on what you're trying to achieve or can I close this and not merge?

ekaradon:
> @ekaradon this PR has changed quite a lot from the original one, can you give us an update on what you're trying to achieve or can I close this and not merge?

I think I already did when you asked the first time [here](https://github.com/pmndrs/drei/pull/378#discussion_r620702343) and [here too](https://github.com/pmndrs/drei/pull/378#issuecomment-827451886). It's still unclear?
But to be honest, even if this is disturbing (for me anyway), I suppose this hook works like intended. I'm still thinking that a hook should not return a mutable object but that's my personal opinion on the matter and I wanted to have your opinions.
So if you don't think it's appropriate or relevant, you can close both the issue and the pull request. I would still like however to know why this approach has been used because it got me confused in the first place?

joshuaellis:
Every time I try to run your fix in codesandbox the window crashes. I think there's an issue if your fix 🤔 and then when I try run it in storybook i see this error: 
image
But I do think your fix is great, I never noticed the issue and it does make debugging a lot better 🤓  I am cautious or re-renders so lets see if we can get it working and go from there, it's probably not that noticeable! Thank you for your work so far!! ⭐

joshuaellis:
I don't mind Breaking Change 😄

ekaradon:
The storybook should work now.

joshuaellis:
Would you mind resolving the merge conflicts and we can get this merged? To confirm, it's a breaking change because we now return `undefined` from `useAnimations` to start with and then update them?

ekaradon:
> Would you mind resolving the merge conflicts and we can get this merged? To confirm, it's a breaking change because we now return `undefined` from `useAnimations` to start with and then update them?

It does not return `undefined`. It returns the initial data provided within `useState`, eg:

```
{
    ref: actualRef,
    clips,
    actions: clips.map((clip) => clip.name).reduce((prev, name: T) => ({ ...prev, [name]: null }), {}),
    names: clips.map((clip) => clip.name),
    mixer,
  }
```
The key difference is that previously this data, specifically the `actions` part was mutated in a useEffect so it would be filled with the `mixer.clipAction` method as soon as the reference was set (`currentRoot`). 
But the object reference was kept, so that the component would not known that values have been set for every action, but the `useEffect` inside the component would not known that those values were not initialized in the first place. 

Here, what we do is first return an object with actions not set (we set them to `null` instead). 
So the first time both the component and the `useEffect` see that the actions are not set. Then the `useEffect` from the `useAnimations` hook create a new object from the initial one and set this time the actions.

joshuaellis:
Thank you for the explanation! There's a TS error with the PR now. But it's all making sense 👍🏼

joshuaellis (approved):

drcmda:
> What I suggest here is not to change the fact that it is initially null, just to update the api (and so re render) once those actions are known? I'm all about immutability, const and frozen data and I really just started this week to use react-three-fiber so I may be a bit biased. And I may lack of background in this matter. I just think that going through mutations is going against the wind when using React.

i think mutation is fine in certain situations. i agree it's dirty, but with the current situation we're putting an internal problem on the users shoulder. i just bumped into it myself - had no idea what's going on, my codebase simply crashed. and i can imagine lots of people are going through the same right now. 

the api itself is a fixed singleton, no need to re-create it. imo a solution could be getters.

ekaradon:
I can retry something with getters if you wish. It would be a lazy initialization then?

drcmda:
this seems to work: https://codesandbox.io/s/exploding-crystal-9eeo2?file=/src/t.tsx:611-612

Issue #379, opened 2021-04-26
Text leaks geometry when unmounted


lojjic:
- `@react-three/drei` version: 4.3.1

### Problem description:

When a `Text` is unmounted, the underlying mesh's `dispose` method is not being called. This is causing that object's geometry to remain in memory indefinitely. I've created a sandbox that logs `gl.info.memory.geometries` and shows it increasing over time as a single Text instance is replaced: https://codesandbox.io/s/hopeful-diffie-sddhj

I initially thought this may be due to:

```js
<primitive
  dispose={undefined}
```

However removing that still doesn't result in `dispose` being called.

Adding something like this does fix it, but I don't know if it's the best way:

```js
    React.useEffect(() => {
      return () => troikaMesh.dispose()
    }, [])
```

drcmda:
thats by design, primitives never dispose because they channel external objects. but dreis Text should absolutely dispose. i think the dispose={undefined} is a v5 leftover.

lojjic:
Thanks, I've created PR #380 to add that useEffect cleanup handler and remove the dispose={undefined}

PR #380, opened 2021-04-26
fix: dispose Text primitive when unmounting to prevent leak (#379)

lojjic:
### Why

See issue #379 -- this fixes a memory leak when unmounting `Text` instances.

### What

- Added the `useEffect` cleanup handler to explicitly call dispose
- Removed the leftover `dispose={undefined}`

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged


joshuaellis (approved):
Great spot, thanks!

PR #385, opened 2021-05-04
feat: add Billboarded wrapper component

pbeshai:
### Why

I wanted to be able to billboard other components besides planes, such as text, but the existing Billboard component is limited to planes. In order to avoid breaking backwards compatibility, I created a new component **Billboarded**.

### What

Adds an alternative approach to billboarding that allows any child to be billboarded, not just planes.

```js
<Billboarded
  follow={true} // Follow the camera (default=true)
  lockX={false} // Lock the rotation on the x axis (default=false)
  lockY={false} // Lock the rotation on the y axis (default=false)
  lockZ={false} // Lock the rotation on the z axis (default=false)
>
  <Text fontSize={1}>I'm a billboard</Text>
</Billboarded>
```

May-03-2021 19-09-58

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [ ] Ready to be merged


drcmda:
I think billboard can and should be used like that, I have been putting stuff in it I'm sure. Let's look at what might cause a breaking changes again, would be best imo to add this functionality to billboard if it doesn't exist yet. Also instead of cloning you can create a group and put the ref on it, cloning would break react composition.

joshuaellis (changes requested):
see comments from @drcmda above.

gsimone:
To avoid a breaking change, we could add a fork to <Billboard

```jsx
<Billboard plane={false} /> // bad name
```

```jsx
// billboard.tsx
if (plane)   return <Plane ref={mergeRefs([localRef, ref])} {...props} />
return <group ref={mergeRefs([localRef, ref])} {...props} />
```

drcmda:
> To avoid a breaking change, we could add a fork to <Billboard
> 
> ```js
> <Billboard plane={false} /> // bad name
> ```
> 
> ```js
> // billboard.tsx
> if (plane)   return <Plane ref={mergeRefs([localRef, ref])} {...props} />
> return <group ref={mergeRefs([localRef, ref])} {...props} />
> ```

i agree, that would be a good solution! and inspecting billboard deeper i start to think it makes no sense for it to be a plane, like ... why. from an api standpoint plane should be false by default but that would break code, so let's use the prop.

also, has anyone noticed that billboards flip? i don't know what the lock props are for but no combination has ever produced a plane that faces the camera and stays up. is that even possible to implement?

m9dfukc:
I wonder if the Billboard does a proper "billboard" implementation anyway?! Shouldn't a billboarded object always face the camera and keep it's orientation like [here](https://threejsfundamentals.org/threejs/lessons/threejs-billboards.html)?!

gsimone:
It does seem a bit weird how sometimes the billboards rotate, fixes are welcome - and easier to release than API changes

drcmda:
> I wonder if the Billboard does a proper "billboard" implementation anyway?! Shouldn't a billboarded object always face the camera and keep it's orientation like [here](https://threejsfundamentals.org/threejs/lessons/threejs-billboards.html)?!

most definitively it should.

so i'd be in favour of scrapping the current impl completely then. we can say it's a bug with confidence. and then just make it channel any arbitrary content instead of a plane with correct transformation.

gsimone:
It's still a breaking change, so 5.0.0, just to make sure we don't go around breaking stuff

pbeshai:
I'm happy to do some updates to match the feedback here, but I don't know the math to make billboard work like sprites/points. The code currently uses Object3D.lookAt which does quaternion magic to me :) 

Is this the correct list of things to do:

- [ ] Delete Billboarded and move functionality into Billboard
- [ ] Replace Plane with group and simplify prop types to match Billboarded. 
- [ ] Use a better billboarding method than lookAt() ?

@drcmda for my own education can you explain or point me to an article about how cloneElement breaks react composition? I don't use it much but it seemed reasonable in this case.

pbeshai:
One downside of using `<group>` instead of cloneElement is you have to split up where you put your positioning:

```js
// with group
<Billboard position={[0, 1.05, 0]}>
  <Text>hello</Text>
</Billboard>

// with cloneElement
<Billboard>
  <Text position={[0, 1.05, 0]}>hello</Text>
</Billboard>
```

So swapping between using it and not is a bit more work. I did consider doing it as a hook and using a ref but that seemed less convenient. 

I admittedly don't know much about 3d graphics. Am I conflating two different things here: billboarding and "always looking at the camera"? I don't know if there's a subtle difference between them (e.g. is billboarding just 2d stuff that isn't affected by perspective or can't rotate and "always look at" is just simply the object rotates to look at the camera?). If so, maybe this should've been `TargetCamera` or something instead of billboard?

m9dfukc:
@pbeshai the way I do it (in pseudo code) is rather simple (no guarantee that this is the correct way to do it):

1. store the position of the object in a var
2. move it to (0,0,0)
3. apply the lookAt(camera) 
4. move the object back to the stored position

joshuaellis:
@pbeshai please request a review when this is ready.

joshuaellis:
@gsimone youre a camera whiz, any thoughts on this? 😄

drcmda:
i bumped into this usecase again and i just can't make it using dreis billboard. i don't really now what the exact usecase is, i don't understand what locking axes does. stackoverflows answer seems to be super simple: 

```jsx
useFrame(({ camera }) => ref.current.quaternion.copy(camera.quaternion))
```

and indeed, this does exactly what i through a billboard/sprite is for: https://codesandbox.io/s/wordcloud-yup2o?file=/src/App.js

pbeshai:
Can confirm this fixes the nested child case as well. I can update the PR with it since it seems to make it work perfectly. 

Should we also remove the lock props?

stephencorwin:
> Should we also remove the lock props?

The lock props are in there for 2.5D experiences and other things where you might want to lock a particular access and allow a thing to rotate towards you but still be constrained.
For example maybe you don't want it to flip on the z-axis but you do want it to rotate on the x and y.

I'm all for it being a group where it just forces the children to lookAt the camera. I'm curious as to why the new implementation is copying the `quaternions` instead of just using `lookAt()` though.

drcmda:
you can try it in the demo @stephencorwin doesnt seem to be the same, it has that rotation problem

saitonakamura:
Hi folks, can we ressurect this bit? I think the difference isn't in quaternions, but in
* Rotate the object to face the camera
vs
* Rotate the object the same way the camera is rotated

drcmda:
please ressurect it! i'd be fine with with ways, face cam, rotate towards the cam, though i think the latter is probably what most people need. the first could also be useful, but the upside-down rotations are weird and i think that should be treated as a bug.

pbeshai:
I believe it's just waiting a review? I've tried to fix the merge conflicts, hopefully didn't mess any github PR things up 😅

drcmda (approved):
looks perfect to me

PR #388, opened 2021-05-05
feat: add DragControls component

### LINKED ISSUE #197: DragControls feature proposition
Is it possible to create a DragControls component that has event listeners as props the same way a mesh does in React-Three-Fiver: 
```
<mesh
  onClick={(e) => console.log('click')}
  onContextMenu={(e) => console.log('context menu')}
  onDoubleClick={(e) => console.log('double click')}
  onWheel={(e) => console.log('wheel spins')}
  onPointerUp={(e) => console.log('up')}
  onPointerDown={(e) => console.log('down')}
  onPointerOver={(e) => console.log('over')}
  onPointerOut={(e) => console.log('out')}
  onPointerEnter={(e) => console.log('enter')} // see note 1
  onPointerLeave={(e) => console.log('leave')} // see note 1
  onPointerMove={(e) => console.log('move')}
  onPointerMissed={() => console.log('missed')}
  onUpdate={(self) => console.log('props have been updated')}
/>
```
Except, instead of being able to pass a callback to events like **onClick**, we pass them to the **DragControls** triggered events:
``` 
dragstart
drag
dragend
hoveron
hoveroff
```
So, it becomes something like: 
```
<DragControls 
        objects=[ ... someObjects ] 
        camera={someCamera} 
        domElement={someContainer} 
        onHoverOn={ () => console.log("hovering") }
/>
```

Pringels:
### Why

Resolved #197 

### What

Added a DragControls component

### Checklist

- [ ] Documentation updated
- [x] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/.storybook/stories/DragControls.stories.tsx
+++ b/.storybook/stories/DragControls.stories.tsx
@@ -1,0 +1,56 @@
+import * as React from 'react'
+import { withKnobs, number, boolean } from '@storybook/addon-knobs'
+import { DragControls as DragControlsImpl } from 'three-stdlib'
+import { Setup } from '../Setup'
+import { Event, Object3D } from 'three'
+
+import { Box, DragControls } from '../../src'
+
+function DragControlsStory() {
+  const meshRef1 = React.useRef<Object3D>(null!)
+  const meshRef2 = React.useRef<Object3D>(null!)
+  const dragControls = React.useRef<DragControlsImpl>(null!)
+
+  const [objects, setObjects] = React.useState<Object3D[]>()
+
+  React.useEffect(() => {
+    if (!objects) setObjects([meshRef1.current, meshRef2.current])
+    if (dragControls.current) {
+      const { current: controls } = dragControls
+      const handleDragStart = (event: Event) => {
+        event.object.material.wireframe = false
+      }
+      const handleDragEnd = (event: Event) => {
+        event.object.material.wireframe = true
+      }
+      controls.addEventListener('dragstart', handleDragStart)
+      controls.addEventListener('dragend', handleDragEnd)
+      return () => (
+        controls.removeEventListener('dragstart', handleDragStart),
+        controls.removeEventListener('dragend', handleDragEnd)
+      )
+    }
+  }, [objects])
+
+  return (
+    <>
+      {objects && <DragControls ref={dragControls} objects={objects} />}
+      <Box ref={meshRef1}>
+        <meshBasicMaterial attach="material" wireframe />
+      </Box>
+      <Box ref={meshRef2}>
+        <meshBasicMaterial attach="material" wireframe />
+      </Box>
+    </>
+  )
+}
+
+export const DragControlsSt = () => <DragControlsStory />
+
+DragControlsSt.storyName = 'Default'
+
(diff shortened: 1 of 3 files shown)


Pringels:
Having slept on it I think a hook might be more appropriate for drag controls. It would be inconsistent with the rest of the control components, but in this case it would simplify things quite a lot.

src/core/DragControls.tsx
```diff
@@ -0,0 +1,30 @@
...
+>
+
+export const DragControls = React.forwardRef<DragControlsImpl, DragControls>(
+  ({ objects, camera, children, ...restProps }, ref) => {
+    const invalidate = useThree(({ invalidate }) => invalidate)
+    const defaultCamera = useThree(({ camera }) => camera)
+    const gl = useThree(({ gl }) => gl)
+
+    const explCamera = camera || defaultCamera
+    const [controls] = React.useState<DragControlsImpl>(() => new DragControlsImpl(objects, explCamera, gl.domElement))
```
joshuaellis, line 21: So you have to attach dragControls to the meshs your want to use?
It might make more sense as a HOC to be consistent with the other Control APIs we have.

.storybook/stories/DragControls.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+function DragControlsStory() {
+  const meshRef1 = React.useRef<Object3D>(null!)
+  const meshRef2 = React.useRef<Object3D>(null!)
+  const dragControls = React.useRef<DragControlsImpl>(null!)
+
+  const [objects, setObjects] = React.useState<Object3D[]>()
+
+  React.useEffect(() => {
+    if (!objects) setObjects([meshRef1.current, meshRef2.current])
```
joshuaellis, line 17: If it were a HOC, you could potentially make this redundant.

joshuaellis (changes requested):
See comments

src/core/DragControls.tsx
Pringels: I considered using a HOC. The problem is that one can't attach/detach meshes after instantation like one can with `TransformControls`, as all mesh instances have to be passed to the constructor. So forwarding a controls `ref` to the parent component can only happen once all child meshes have rendered.

But yeah you're right HOC is probably better. I will try make it work.

.storybook/stories/DragControls.stories.tsx
Pringels: Yes this is quite an akward API compared to the other `drei` components. I will try to find a cleaner solution.

src/core/DragControls.tsx
joshuaellis: Is there a downside to re-making the the `DragControl` impl when new children are added?

Pringels: I tried that but it causes all event listeners added by the parent component to no longer work. Perhaps I could store event callbacks in the HOC and simply re-apply them to the new `controls` instance. (This would have to be done anyway given that some children might be rendered conditionally)

gsimone:
Closing this since it's been stale for 9 months, feel free to reopen

PR #400, opened 2021-05-20
Lazyanim

drcmda:
uses lazy init for useAnimation actions. this maintains the api as a singleton, but it would also save memory as only the actions that are called will be created.

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


drcmda:
@joshuaellis the build error seems unrelated to the pr? looks like some rollup flag has been deprecated?

joshuaellis (approved):

PR #411, opened 2021-05-25
Remove "might be removed" note

marcofugaro:
Please don't remove that feature, it is very useful 🙏


joshuaellis:
I think we were worried about the overhead it bought when refactoring (if we were too) but tbh, it does seem a lot of people use this feature.... so sure!

PR #415, opened 2021-05-27
fix: dispose line geom cache when points updating

slahser1992:
### Why

memory leak were caused by updating line points

### What

clean up the last frame geometry before drawing new line mesh. 
### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged


joshuaellis (approved):
Thanks for the contribution!

PR #433, opened 2021-06-05
fix: Prevent CameraShake from resetting the camera angle

CharlieHess:
### Why
Cameras that already have a rotation have it reset to `0,0,0` by `CameraShake`, and `additive` is not a good solution (because you're permanently changing the viewing angle of the camera).

### What
A better fix is to remember the initial rotation of the camera, and _always_ apply the shake vector additively. That way when we decay the shake, the camera resets to its original viewing angle.

### Checklist
- [x] Documentation updated
- [x] Storybook verified
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,7 @@ const config = {
   intensity: 1, // initial intensity of the shake
   decay: false, // should the intensity decay over time
   decayRate: 0.65, // if decay = true this is the rate at which intensity will reduce at
+  controls: undefined, // optionally pass a ref to OrbitControls
 }
 
 <CameraShake {...config} />
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ const config = {
   intensity: 1, // initial intensity of the shake
   decay: false, // should the intensity decay over time
   decayRate: 0.65, // if decay = true this is the rate at which intensity will reduce at
-  controls: undefined, // optionally pass a ref to OrbitControls
+  controls: undefined, // if using orbit controls, pass a ref here so we can update the rotation
 }
 
 <CameraShake {...config} />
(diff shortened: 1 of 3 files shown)


joshuaellis (changes requested):
This breaks the story with `OrbitControls` when you try to move the camera, however, I like the idea of removing additive if possible.

Code47X:
@CharlieHess Agreed, I don't think I communicated this well in the original pr. Initially we used a separate internal camera that copied the original position/rotations each frame to serve as a base, and then used this camera to render only if there was any shake being applied at that moment. This allowed the original camera to more/rotate freely without the shake effecting it directly - but there were problems with that solution as well. (Mainly using `setDefaultCamera` in useFrame to toggle between shaking camera and original camera)

CharlieHess:
@joshuaellis what do you think about the latest commit? bit more explicit than setting additive but still requires an extra prop to make `OrbitControls` work

README.md
```diff
@@ -189,9 +189,9 @@ const config = {
   pitchFrequency: 1, // Frequency of the pitch rotation
   rollFrequency: 1, // Frequency of the roll rotation
   intensity: 1, // initial intensity of the shake
-  decay: false // should the intensity decay over time
-  decayRate: 0.65 // if decay = true this is the rate at which intensity will reduce at
-  additive: false // this should be used when your scene has orbit controls
+  decay: false, // should the intensity decay over time
+  decayRate: 0.65, // if decay = true this is the rate at which intensity will reduce at
+  controls: undefined, // optionally pass a ref to OrbitControls
```
joshuaellis, line 194: Maybe can we just add why this should be used in the context of `OrbitControls`?

joshuaellis (changes requested):
Really great solution. I think it'd be helpful if we can just clarify the readme a bit better ⭐

README.md
CharlieHess: good call; what do you think about my edit?

joshuaellis (approved):
Edit is clear, thanks again ⭐

PR #437, opened 2021-06-09
fix: avoid updating OrbitControls when they're disabled

rvanlaarhoven:
### Why

I noticed that the `OrbitControls` are being updated even when they're disabled. I'm not sure if this is the desired behaviour, but if I understood correctly the `enabled` prop is meant to be used if you want to manually adjust camera (e.g. change the camera rotation), which is currently impossible.

### What

I've updated the OrbitControls component to only trigger a `controls.update` whenever `controls.enabled` is true.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged


joshuaellis (approved):
thanks for the contribution! ⭐

PR #443, opened 2021-06-12
Callback to extend the gltf loader for GLTF Extensions

RenaudRohlinger:
### Why
This PR grants access to the instance of the GLTFLoader. It can be necessary, for example with the plugin system of GLTFLoader.

Example :
  ```jsx
  const gltf = useGLTF('/glb/instance.glb', false, false, (loader) => {
    loader.register((parser) => new GLTFInstancingExtension(parser, THREE))
  })
```

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/useGLTF.tsx
+++ b/src/core/useGLTF.tsx
@@ -3,10 +3,10 @@ import { Loader } from 'three'
 import { GLTFLoader, DRACOLoader, MeshoptDecoder } from 'three-stdlib'
 import { useLoader } from '@react-three/fiber'
 
-function extensions(useDraco: boolean | string, useMeshopt: boolean, extendLoader?: (loader: Loader) => void) {
+function extensions(useDraco: boolean | string, useMeshopt: boolean, extendLoader?: (loader: GLTFLoader) => void) {
   return (loader: Loader) => {
     if (extendLoader) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/core/useGLTF.tsx [hunk 1]
```diff
@@ -16,10 +19,15 @@ function extensions(useDraco: boolean | string, useMeshopt: boolean) {
   }
 }
 
-export function useGLTF(path: string, useDraco: boolean | string = true, useMeshOpt: boolean = true) {
-  const gltf = useLoader(GLTFLoader, path, extensions(useDraco, useMeshOpt))
+export function useGLTF(
+  path: string,
+  useDraco: boolean | string = true,
+  useMeshOpt: boolean = true,
+  extendLoader: any
```
joshuaellis, line 26: This shouldn't be any. This should be a function type. I'm assuming it returns void?

joshuaellis (changes requested):

src/core/useGLTF.tsx
RenaudRohlinger: @joshuaellis Yes, it returns void. Fixed the ts.

joshuaellis (approved):
Awesome! 🚀

PR #446, opened 2021-06-13
feat: implemented bvh hook

RenaudRohlinger:
### What

A hook to speed up the default raycasting by using the [BVH Implementation by @gkjohnnson](https://github.com/gkjohnson/three-mesh-bvh).

```jsx
const mesh = useRef()
useBVH(mesh)

<mesh ref={mesh} ... />
```

Storybook:
https://drei-git-fork-renaudrohlinger-bvh-hook-pmndrs.vercel.app/?path=/story/performance-usebvh--default-story

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


src/core/useBVH.tsx
```diff
@@ -0,0 +1,33 @@
...
+export interface BVHOptions {
+  splitStrategy?: 'CENTER' | 'AVERAGE' | 'SAH'
+  verbose?: boolean
+  setBoundingBox?: boolean
+  maxDepth?: number
+  maxLeafTris?: number
+}
+
+export function useBVH(object3D: React.MutableRefObject<Mesh | undefined>, options?: BVHOptions) {
+  const ref = React.useRef<Object3D>()
```
joshuaellis, line 14: This ref seems a bit redundant? We're not using it for anything I don't think?

src/core/useBVH.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+      return () => {
+        if (geometry.boundsTree) {
+          geometry.disposeBoundsTree()
+        }
+      }
+    }
+  }, [object3D, options])
+
+  return ref
```
joshuaellis, line 32: Equally this return would therefore be redundant. Is this because you _have_ to return something from the hook? if so, it should just return undefined...

src/core/useBVH.tsx
```diff
@@ -0,0 +1,33 @@
...
+      geometry.disposeBoundsTree = disposeBoundsTree
+      geometry.computeBoundsTree(options)
+
+      return () => {
+        if (geometry.boundsTree) {
+          geometry.disposeBoundsTree()
+        }
+      }
+    }
+  }, [object3D, options])
```
joshuaellis, line 30: afaik the ref we're passing won't cause a re-render of any kind? i.e if you change what the ref is attached too the effect won't re-run. If we want to do that, we should return a `useCallback` styled hook like [this](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node).

joshuaellis (approved):
Some discussion points around implementation left! ⭐

src/core/useBVH.tsx
RenaudRohlinger: Yeah, I took useHelper() as a template that was using the same logic. I'm ok with removing it

joshuaellis:
damn, @RenaudRohlinger i was meant to click `request changes` when it's ready again for a review, would you mind re-requesting please? If you wanna chat more about it, let me know 😄

src/core/useBVH.tsx
RenaudRohlinger: @joshuaellis So for the ref I totally agree, but if we use useCallback here I think it will make the feature more complex to use? Maybe I did not understand, I'm talking about ref={useBVH} here.
Also, I think we need useEffect to enable BVH onMount. If you think that useCallback is easier I'm totally ok to change to that, I just want to call "computeBoundsTree" once the component is mounted. About the unMount we don't really care as anyway the geometry will get disposed.

joshuaellis: Because using `useRef` won't cause a re-render, so the `useEffect` hook won't refire when the dependancy changes cause it's a ref, instead we should look at using the `useCallback` pattern:

```js
export function useBVH(options) {
  const ref =  React.useCallback(
    (node) => {
      node.raycast = acceleratedRaycast
      let geometry: any = node.geometry
      geometry.computeBoundsTree = computeBoundsTree
      geometry.disposeBoundsTree = disposeBoundsTree
      geometry.computeBoundsTree(options)
    },
    [options]
  )

  return ref
}
```

So you'd pass the ref returned from the function to the `jsx` component.

RenaudRohlinger: @joshuaellis well I tested a bit and I don't think the useCallback approach would work in this usecase. The useEffect approach is working and re-render when needed, I can enable/disable BVH if required, and even dynamic options works as you can see in this example: https://drei-git-fork-renaudrohlinger-bvh-hook-pmndrs.vercel.app/?path=/story/performance-usebvh--default-story.

I update the PR so you can merge it.
But if you really think that useCallback is better I would be happy to learn from it but I'm pretty sure it will get called multiple times unnecessarily and we won't be able to use any hook such as useHelper.

joshuaellis: No no, it was just an idea, if it's re-rendering correctly lets keep your method! ⭐  Thanks for investigating it though 😄

joshuaellis (approved):

src/core/useBVH.tsx
RenaudRohlinger: Ah I see, thanks for the heads up 😊

drcmda:
@RenaudRohlinger using this for a while makes me think, should this be a hook? in practice it has gotten a little cumbersome because:

```jsx
const ref1 = useRef()
const ref2 = useRef()
const ref3 = useRef()
const ref4 = useRef()
const ref5 = useRef()
const ref6 = useRef()
...
useBVH(ref1)
useBVH(ref2)
useBVH(ref3)
useBVH(ref4)
useBVH(ref5)
useBVH(ref6)
...
<mesh ref={ref1} ...
<mesh ref={ref2} ...
<mesh ref={ref3} ...
<mesh ref={ref4} ...
<mesh ref={ref5} ...
<mesh ref={ref6} ...
```

i start to think, maybe it should be a plain three class?

```jsx
class BVHMesh extends THREE.Mesh {
  constructor(geometry, material, options = { ... }) {
    super(geometry, material)
    this.raycast = acceleratedRaycast
    this.geometry.computeBoundsTree = computeBoundsTree
    this.geometry.disposeBoundsTree = disposeBoundsTree
    this.geometry.computeBoundsTree(options)
  }

  dispose() {
    super.dispose()
    geometry.disposeBoundsTree()
  }
}
```

because now this would be possible

```jsx
import { extend } from '@react-three/fiber'
import { BVHMesh } from '@react-three/drei'

extend({ Mesh: BVHMesh })

<mesh>
  <sphereGeometry />
</mesh>
```

joshuaellis:
> because now this would be possible

If we choose to refactor, I think it should be a component that we extend in `drei`. But still export the raw class we've created from something (like we do with Reflector Material)

drcmda:
> > because now this would be possible
> 
> If we choose to refactor, I think it should be a component that we extend in `drei`. But still export the raw class we've created from something (like we do with Reflector Material)

a component would not be able to just turn the system into bvh. a class + extend could. imo a component is the wrong tool as well as it would force you to use it everywhere. we have the extend function, it's usable for cases like these.

joshuaellis:
No I think you're missing what i'm saying. From `drei` we'd export the component & the class. So the user doesn't have to extend if they dont want to.

drcmda:
so the component would then be like:

```jsx
function AccMesh({ geometry, material, options, ...props }) {
  cost mesh = useMemo(() => new BVHMesh(geometry, material, options), [geometry, material, options])
  useEffect(() => () => mesh.dispose(), [mesh])
  <primitive object={mesh} {...props} />
}
```

that and the class as exports? let's do it!

RenaudRohlinger:
@drcmda @joshuaellis I agree about the component but how would we handle the instancedmesh, lines, points meshe cases? 

I think a wrapper component would be the best solution:
```jsx


const Bvh = ({ children }) => {
  let bvhArr = []

  Children.forEach(children, (child) => {
     ...do the bvh stuff here
     push(newBvhMesh)
  })

  return  <group>{bvhArr}</group>
}

<Bvh>
  <mesh/>
  <point/>
  <instancedmesh/>
</Bvh>

```

drcmda:
react.children is sometimes fragile as it can break composition:

```jsx
const Foo = () => <mesh />

const Bvh = ({ children }) => {
  Children.forEach(children, (child) => {
    // child isn't a mesh nor is the mesh accessible from here
```

though the idea of creating bvh "bounds" to me sounds like the best solution: everything inside my bound has bvh, everything else doesn't. but i have no idea how this could be solved technically without cloning or messing with children.

PR #451, opened 2021-06-19
fix: avoid creating new Buffer Geometry

Gusted:
### Why

So, geometry variable will be overridden anyway.

### What

Removing the `let geometry = new THREE.BufferGeometry()` and replace it by the next variable declaration seems to be working and avoiding to create more unnecessary BufferGeometry's.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/useEdgeSplit.tsx
+++ b/src/core/useEdgeSplit.tsx
@@ -16,9 +16,7 @@ export function useEdgeSplit(cutOffAngle: number, tryKeepNormals: boolean = true
 
   React.useEffect(() => {
     if (original.current && ref.current && modifier.current) {
-      let geometry = original.current.clone()
-
-      const modifiedGeometry = modifier.current.modify(geometry, cutOffAngle, tryKeepNormals)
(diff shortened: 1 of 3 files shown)


Gusted:
@joshuaellis this the correct commit message, right?

drcmda:
Didn't clone also create new buffergeom?

Gusted:
> Doesnt clone also create new buffergeom?

It does. This PR is rather to avoid having an even extra creation of buffer geometry. 
But the clone seems necessary in my eyes, as not cloning somehow caused errors. But I can work around that issues. Let me fix that.

joshuaellis:
They all look good now @Gusted unless i've missed something? 

Also yes, this message is the correct format, but typically the maintainer can change it before hand anyway, so no stress 👍🏼

Gusted:
> They all look good now @Gusted unless i've missed something?

Yep, they look good. No extra buffer geometries are created now on the 3 functions.

joshuaellis (approved):
🚀

PR #452, opened 2021-06-19
refactor(eslint): enable prefer-const rule

Gusted:
### Why

Will enforce `const` when variable is not being re-declared and allow javascript engines for optimizations.

### What

Enabling the `prefer-const` eslint will automatically fix this.

### Checklist

- [x] Ready to be merged

Also did some manual refactoring when I was browsing trough the code. Just removing unused code.


joshuaellis (approved):

PR #454, opened 2021-06-23
fix: useDetectGPU is fetch-based and therefore should use suspense

drcmda:
twitter thread: https://twitter.com/lewy_blue/status/1407521953682915328

the current one isn't reliable because it shoots a blank first. since detect-gpu sends a fetch request it should be suspense.

#### useDetectGPU

[](https://drei.pmnd.rs/?path=/story/misc-usedetectgpu)

This hook uses [DetectGPU by @TimvanScherpenzeel](https://github.com/TimvanScherpenzeel/detect-gpu), wrapped into suspense, to determine what tier should be assigned to the user's GPU.

👉 This hook CAN be used outside the react-three-fiber `Canvas`.

```jsx
function App() {
  const GPUTier = useDetectGPU()
  // show a fallback for mobile or lowest tier GPUs
  return (
    {(GPUTier.tier === "0" || GPUTier.isMobile) ? <Fallback /> : <Canvas>...</Canvas>

<Suspense fallback={null}>
  <App />
```

### DIFF
--- a/src/core/useDetectGPU.tsx
+++ b/src/core/useDetectGPU.tsx
@@ -1,12 +1,8 @@
 import * as React from 'react'
 import { getGPUTier, GetGPUTier, TierResult } from 'detect-gpu'
+import { useAsset } from 'use-asset'
 
 export function useDetectGPU(props?: GetGPUTier): TierResult | null {
   const [GPUTier, setGPUTier] = React.useState<TierResult | null>(null)
-
-  React.useEffect(() => {
-    getGPUTier(props).then((result) => setGPUTier(result))
-  }, [props])
-
-  return GPUTier
+  return useAsset<TierResult, [string]>(() => getGPUTier(props), 'useDetectGPU')
 }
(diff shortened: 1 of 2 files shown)


src/core/useDetectGPU.tsx
```diff
@@ -1,12 +1,8 @@
 import * as React from 'react'
 import { getGPUTier, GetGPUTier, TierResult } from 'detect-gpu'
+import { useAsset } from 'use-asset'
 
 export function useDetectGPU(props?: GetGPUTier): TierResult | null {
```
joshuaellis, line 5: would this make the return value of this hook incorrect? i.e. it won't return null because it's suspended?

joshuaellis (changes requested):

src/core/useDetectGPU.tsx
drcmda: you're right!

joshuaellis (approved):
Looks good to me, this is a breaking change though. So we'll push this as that.

PR #455, opened 2021-06-23
feat: add a material property to receive shared material

slahser1992:
### Why

when mounted <Line/>, I had to create a new LineMaterial. it will waste a lot of  memory, so I hope to add a material property to receive a shared material to save memory.

### What

add a material property to receive shared material.

### Checklist

yes.

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

by the way, whether to give a common material hash map to store exist material to reuse.


src/core/Line.tsx
```diff
@@ -14,11 +14,11 @@ export type LineProps = {
   >
 
 export const Line = React.forwardRef<Line2, LineProps>(function Line(
-  { points, color = 'black', vertexColors, lineWidth, dashed, ...rest },
+  { points, color = 'black', vertexColors, lineWidth, dashed, material, ...rest },
   ref
 ) {
   const [line2] = React.useState(() => new Line2())
-  const [lineMaterial] = React.useState(() => new LineMaterial())
+  const [lineMaterial] = React.useState(() => (material ? material : new LineMaterial()))
```
joshuaellis, line 21: I wonder if we would be better `extending` the `LineMaterial` and using it as a child instead of a primitive, then we can pass children to the component 🤔 What do you think?

joshuaellis (changes requested):

src/core/Line.tsx
slahser1992: that's better, got it.

slahser1992: > I wonder if we would be better `extending` the `LineMaterial` and using it as a child instead of a primitive, then we can pass children to the component  What do you think?

that's better! I still have a question. previous users feeling as considered, I will support two choices to draw line.

1. old way.
2. if the material child exists, we will use the material component to draw line.

that's enough?

joshuaellis: yeah I was thinking it could be like this:

```
<primitive object={line2}>
{children ? children : <lineMaterial />
</primitive>
```

Simple check similar to what you've done already put in the return so no state setting.

slahser1992: > yeah I was thinking it could be like this:
> 
> ```
> <primitive object={line2}>
> {children ? children : <lineMaterial />
> </primitive>
> ```
> 
> Simple check similar to what you've done already put in the return so no state setting.

got it, thanks.

slahser1992: > yeah I was thinking it could be like this:
> 
> ```
> <primitive object={line2}>
> {children ? children : <lineMaterial />
> </primitive>
> ```
> 
> Simple check similar to what you've done already put in the return so no state setting.

oh, I'm having some problems about fix this, when I made a shared material as a component <lineMaterial/> mount, it always creating a new shaderMaterial. 

```
import { Vector2, Vector3, Color } from 'three'
import { ReactThreeFiber, extend } from '@react-three/fiber'
import { LineGeometry, LineMaterial, LineMaterialParameters, Line2 } from 'three-stdlib'
extend({ LineMaterial })

export type LineProps = {
  points: Array<Vector3 | [number, number, number]>
  children?: React.ReactNode
  vertexColors?: Array<Color | [number, number, number]>
} & LineMaterialParameters &
  Omit<ReactThreeFiber.Object3DNode<Line2, typeof Line2>, 'args'> &
  Omit<
    ReactThreeFiber.Object3DNode<LineMaterial, [LineMaterialParameters]>,
    'color' | 'vertexColors' | 'resolution' | 'args'
  >

export type LineMaterialProps = Pick<LineMaterialParameters, 'color' | 'dashed' | 'dashScale' | 'dashSize' | 'dashOffset' | 'resolution'>
  
declare global {
  namespace JSX {
    interface IntrinsicElements {
      lineMaterial: LineMaterialProps
    }
  }
}

export const Line = React.forwardRef<Line2, LineProps>(function Line(
  { points, color = 'black', vertexColors, lineWidth, dashed, children, ...rest },
  ref
) {
  const [line2] = React.useState(() => new Line2())
  const [lineMaterial] = React.useState(() => (children ? children : new LineMaterial()))
  const [resolution] = React.useState(() => new Vector2(512, 512));

  const lineGeom = React.useMemo(() => {
    const geom = new LineGeometry()
    const pValues = points.map((p) => (p instanceof Vector3 ? p.toArray() : p))
    geom.setPositions(pValues.flat())

    if (vertexColors) {
      const cValues = vertexColors.map((c) => (c instanceof Color ? c.toArray() : c))
      geom.setColors(cValues.flat())
    }

    return geom
  }, [points, vertexColors])

  React.useLayoutEffect(() => {
    line2.computeLineDistances()
  }, [points, line2])

  React.useLayoutEffect(() => {
    if (lineMaterial instanceof LineMaterial) {
      if (dashed) {
        lineMaterial.defines.USE_DASH = ''
      } else {
        // Setting lineMaterial.defines.USE_DASH to undefined is apparently not sufficient.
        delete lineMaterial.defines.USE_DASH
      }
      lineMaterial.needsUpdate = true
    }
  }, [dashed, lineMaterial])

  React.useEffect(() => {
    return () => lineGeom.dispose()
  }, [lineGeom])

  return (
    <primitive object={line2} ref={ref} {...rest}>
      <primitive object={lineGeom} attach="geometry"/>
      {
        children
          ? children
          : <primitive
              object={lineMaterial}
              attach="material"
              color={color}
              vertexColors={Boolean(vertexColors)}
              resolution={resolution}
              linewidth={lineWidth}
              dashed={dashed}
              {...rest}
            />
      }
    </primitive>
  )
})

```

slahser1992: so I think the initial pr is more reasonable.

joshuaellis: thats not how you pass children in react and the implementation you've tried to do is not what i was suggesting. I can take a look at this over the next few days.

slahser1992: > thats not how you pass children in react and the implementation you've tried to do is not what i was suggesting. I can take a look at this over the next few days.

OK.

FarazzShaikh:
Lets close this as stale?

PR #459, opened 2021-06-25
fix: pass domElement to OrbitControls

haywirez:
This should allow a prop specifying an optional pointer event target DOM node passed to OrbitControls, e.g. `<OrbitControls domElement={canvasRef.current} />`. With the latest consolidation in three.js to use pointer events, this is important for the controls to work properly on mobile (see https://github.com/pmndrs/three-stdlib/blob/main/src/controls/OrbitControls.ts#L102). 

It would be even better if a ref to @react-three/fiber `<Canvas/>` could be passed in implicitly by default, but I don't think it's exposed anywhere — suggestions welcome.

### DIFF
--- a/src/core/OrbitControls.tsx
+++ b/src/core/OrbitControls.tsx
@@ -7,19 +7,24 @@ export type OrbitControlsProps = ReactThreeFiber.Overwrite<
   {
     target?: ReactThreeFiber.Vector3
     camera?: THREE.Camera
+    domElement?: HTMLElement
     regress?: boolean
     enableDamping?: boolean
   }
 >
 
 export const OrbitControls = React.forwardRef<OrbitControlsImpl, OrbitControlsProps>(
-  ({ camera, regress, enableDamping = true, ...restProps }, ref) => {
+  ({ camera, regress, domElement, enableDamping = true, ...restProps }, ref) => {
     const invalidate = useThree(({ invalidate }) => invalidate)
     const defaultCamera = useThree(({ camera }) => camera)
     const gl = useThree(({ gl }) => gl)
     const performance = useThree(({ performance }) => performance)
     const explCamera = camera || defaultCamera
-    const controls = React.useMemo(() => new OrbitControlsImpl(explCamera), [explCamera])
+    const explDomElement = domElement || gl.domElement
+    const controls = React.useMemo(
+      () => new OrbitControlsImpl(explCamera, explDomElement),
+      [explCamera, explDomElement]
+    )
 
     useFrame(() => {
       if (controls.enabled) controls.update()
@@ -31,7 +36,6 @@ export const OrbitControls = React.forwardRef<OrbitControlsImpl, OrbitControlsPr
         if (regress) performance.regress()
       }
 
-      controls.connect(gl.domElement)
       controls.addEventListener('change', callback)
       return () => {
         controls.removeEventListener('change', callback)


src/core/OrbitControls.tsx
```diff
@@ -7,19 +7,20 @@ export type OrbitControlsProps = ReactThreeFiber.Overwrite<
...
     enableDamping?: boolean
   }
 >
 
 export const OrbitControls = React.forwardRef<OrbitControlsImpl, OrbitControlsProps>(
-  ({ camera, regress, enableDamping = true, ...restProps }, ref) => {
+  ({ camera, regress, domElement, enableDamping = true, ...restProps }, ref) => {
     const invalidate = useThree(({ invalidate }) => invalidate)
     const defaultCamera = useThree(({ camera }) => camera)
     const gl = useThree(({ gl }) => gl)
```
joshuaellis, line 20: couldn't we get the domElement from the gl?

joshuaellis (changes requested):

src/core/OrbitControls.tsx
njm222: Yes, but passing it as a prop provides more flexibility... so you don't have to use the same domElement from the gl

joshuaellis: I think if no `domElement` is present it should use the canvas as default, or does it already do that in the implementation from `three-stdlib`?

njm222: not 100% sure if it already does that but it should

haywirez: 👍 Force pushed updated code

haywirez: 🤔  Now that I think about it, I'm not sure that getting the `parentElement` is a smart approach. Would need to know whether `@react-three/fiber` was used via a <Canvas/> wrapper node or directly (without `react-dom` from v6).

joshuaellis: I think `gl.domElement` should probably be the default element @haywirez, this would resolve the requirement of knowing if someone used `<Canvas` or `render`

haywirez: OK, my latest change adds just that

joshuaellis (approved):
This addition looks good to me, is there a way we can test it work as intended?

haywirez:
> This addition looks good to me, is there a way we can test it work as intended?

Yes, check these urls with the mobile device toolbar enabled in chrome devtools:
https://drei.pmnd.rs/?path=/story/controls-orbitcontrols--orbit-controls-story [`master`] 
vs.
https://drei-git-fork-haywirez-orbitcontrols-pass-domelement-pmndrs.vercel.app/?path=/story/controls-orbitcontrols--orbit-controls-story [`this PR`]

src/core/OrbitControls.tsx
```diff
@@ -7,19 +7,24 @@ export type OrbitControlsProps = ReactThreeFiber.Overwrite<
...
     const performance = useThree(({ performance }) => performance)
     const explCamera = camera || defaultCamera
-    const controls = React.useMemo(() => new OrbitControlsImpl(explCamera), [explCamera])
+    const explDomElement = domElement || gl.domElement
+    const controls = React.useMemo(
+      () => new OrbitControlsImpl(explCamera, explDomElement),
```
stephencorwin, line 25: on L39, we may no longer need to `controls.connect(gl.domElement)` if we are connecting to the dom element in the constructor. L39 can likely be deleted with this change

stephencorwin (changes requested):

src/core/OrbitControls.tsx
joshuaellis: @haywirez if we can implement this, we can merge this today. Sorry for the delay!

haywirez: Thanks for catching this, just force pushed the update with the changes 👍

joshuaellis (approved):

Issue #460, opened 2021-06-26
Pointer events: auto in the HTML component makes showing the component on hover tricky


looeee:
- `three` version: 129
- `@react-three/fiber` version: 7.01
- `@react-three/drei` version: 6.2.0

### Problem description:

I've created a <HTML> component:

```js
export default function Label({visible}) {
  return (
    <Html
      rotation-x={-Math.PI / 2}
      transform
      sprite
      className="pointer-events-none"
    >
      <div
        className={
          visible
            ? "bg-red-400 pointer-events-none"
            : "bg-red-400 pointer-events-none opacity-0"
        }
      >
        <h1 className="pointer-events-none">TESTING</h1>
      </div>
    </Html>
  );
}


```

And I'm showing it on hover like this:

```js
  const [hovered, setHovered] = useState(false);
  const [labelPosition, setLabelPosition] = useState([0, 0, 0]);

  return (
    <group
      {...props}
      dispose={null}
      onPointerOver={(e) => {
        setHovered(true);
        e.stopPropagation();
      }}
      onPointerLeave={(e) => {
        setHovered(false);
        e.stopPropagation();
      }}
    >
      <Label
        visible={hovered ? true : false}
      ></Label>
   </group>)
```

That all works OK, except that the label overlays the mesh and when the mouse goes over the label it's no longer hitting the mesh so hover gets set to false, the label disappears, the raycast hits the mesh, so the label appears again, repeat... in short, flickering occurs. 

...hence all the `pointer-events-none` classes. The problem is, when `transform` is set to true on `<HTML>` the following lines run and set a `pointer-events: auto` on the `transformInnerRef` which I have no control over. 

https://github.com/pmndrs/drei/blob/12e45b7dfffb977a74b90abed92c524ffa2e36f6/src/web/Html.tsx#L190-L204

### Suggested solution:

I'm not sure what the purpose of `pointer-events: auto` on `transformInnerRef` is. Could it be removed?

njm222:
It's needed for pointer events on the HTML, but I think we can remove it since pointer events can still target the children.

So we can:
1. remove pointerEvents and leave it all to the users to decide how to deal with pointer events
2. move pointerEvents to a prop
3. other suggestions..?

I prefer removing it so that it inherits the `pointerEvents: none` from the parent and users can just add pointerEvents to divs that need it.

I can implement it pretty quick if other people are on board?

drcmda:
[...]

PR #492, opened 2021-07-23
fix: pointerEvents for Html

njm222:
### Why

There was an unecessary pointerEvents styles on one of the child divs of the Html component

### What

Removed the pointerEvent style

### Checklist

NA

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

resolves issue: #460

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -86,6 +86,19 @@ const getObjectCSSMatrix = ((scaleMultipliers: (n: number) => number[]) => {
   return (matrix: Matrix4, factor: number) => getCSSMatrix(matrix, scaleMultipliers(factor), 'translate(-50%,-50%)')
 })((f: number) => [1 / f, 1 / f, 1 / f, 1, -1 / f, -1 / f, -1 / f, -1, 1 / f, 1 / f, 1 / f, 1, 1, 1, 1, 1])
 
+type PointerEventsProperties =
+  | 'auto'
+  | 'none'
+  | 'visiblePainted'
+  | 'visibleFill'
+  | 'visibleStroke'
+  | 'visible'
+  | 'painted'
+  | 'fill'
+  | 'stroke'
+  | 'all'
+  | 'inherit'
+
 export interface HtmlProps
   extends Omit<Assign<React.HTMLAttributes<HTMLDivElement>, ReactThreeFiber.Object3DNode<Group, typeof Group>>, 'ref'> {
   prepend?: boolean
@@ -101,6 +114,7 @@ export interface HtmlProps
   onOcclude?: (visible: boolean) => null
   calculatePosition?: CalculatePosition
   as?: string
+  pointerEvents?: PointerEventsProperties
 }
 
 export const Html = React.forwardRef(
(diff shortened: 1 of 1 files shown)


drcmda:
quick ref to the pr that added it: https://github.com/pmndrs/drei/pull/296

> While on desktop the events are forwarded easily between html and webgl, on mobile you have to perform a double tap in order to switch between them (what I mean with "forwarding events" is simply setting pointerEvents to "none" or "auto" on the webgl container)

if we remove it, are users still allowed to fix this use case? and if it's mobile and it really doesnt work without that prop, doesn't it mean it must be in there?

njm222:
I haven't tested it, I will test it and get back to you.

But theoretically it should work by just adding a div with `pointerEvents: auto` as a child of the Html component.

Here is the doc for reference: https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events
you can take a look at the `none` value

drcmda:
what i mean is that the description makes it sound like without it is will fail on mobile so you'd always want it. if it's needed in 100% of all cases then it wouldn't make sense to make people add it, because they have to unless they want it to fail on mobile. again, i don't understand the problem - just going by that text the author of that pr wrote. just afraid we're taking it out for one use case, but then other use cases break on mobile.

njm222:
Ahhh I understand your concern. It is hard to understand what he means by that quote, I thought it was just a known issue with pointerEvents on touch displays and the double tap is how you get 'behind' the `Html` component even with `auto`

If we set the Html to have `pointerEvents: none` there is nothing stopping the user adding a child element to the Html with `pointerEvents: auto` so I don't think we will only hit 50% of the audience,  but I could be mistaken..

If we don't get a response from @GiuseppeRaso I'll make it an optional prop just to be on the safe side

drcmda:
sounds good, though if we test mobile and no one sees any problem i think we can remove it without extra flags. what say you @joshuaellis @gsimone

GiuseppeRaso:
> While on desktop the events are forwarded easily between html and webgl, on mobile you have to perform a double tap in order to switch between them (what I mean with "forwarding events" is simply setting pointerEvents to "none" or "auto" on the webgl container

That comment of mine was referred to a previous version of the PR, after some discussion we decided to make the PR simpler so that comment is no longer valid and you can ignore it

My actual PR was similar to THREE js implementation, infact they both set pointer events  to none on the renderer and then to auto on the child
https://github.com/mrdoob/three.js/blob/38954eff2bcac74895fb0e7914ae33e9998fd0c5/examples/js/renderers/CSS3DRenderer.js#L14
https://github.com/mrdoob/three.js/blob/38954eff2bcac74895fb0e7914ae33e9998fd0c5/examples/js/renderers/CSS3DRenderer.js#L92

Besides I think it could be confusing for the element inside Html to not be clickable by default

looeee:
> Besides I think it could be confusing for the element inside Html to not be clickable by default

Yeah, and it looks like it would only be unclickable when `transform` is set which is definitely confusing 🤔

njm222:
So consensus on making `pointerEvents` an optional prop and by default it is set to `auto`?

joshuaellis:
> So consensus on making `pointerEvents` an optional prop for the child and by default it is set to `auto`?

Make a PR and let's check it out 💪🏼

joshuaellis:
> So consensus on making `pointerEvents` an optional prop for the child and by default it is set to `auto`?

Edit the PR and we'll go from there 🤓 💪🏼

njm222:
@joshuaellis changes have been made

joshuaellis (approved):

PR #503, opened 2021-08-01
fix: broken Misc/useDetectGPU page in the docs

shuta13:
### Why

The live sample of Misc/useDetectGPU page in the docs ( https://docs.pmnd.rs/drei/misc/use-detect-gpu ) was broken, as shown in the image below.

スクリーンショット 2021-08-02 7 23 36

### What

Added <React.Suspense fallback=...> based on the error messages.

### Checklist

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/.storybook/stories/useDetectGPU.stories.tsx
+++ b/.storybook/stories/useDetectGPU.stories.tsx
@@ -14,14 +14,18 @@ export default {
 function Simple() {
   const GPUTier = useDetectGPU()
 
-  return GPUTier ? (
-    <Text>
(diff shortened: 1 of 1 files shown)


joshuaellis (approved):
You probably need to make this change in the `docs` repo too.. I don't think they've managed to sort out pulling from the repos yet? But thanks for the fix on our side!

shuta13:
@joshuaellis 

> You probably need to make this change in the docs repo too..

Oh, It may indeed need to be changed. I'll check it out. Thank you for your response!

PR #521, opened 2021-08-16
feat(Html): allow import from root without react-dom

saitonakamura:
### Why

Importing drei component from the root (e.g. `import { OrbitControls } from '@react-three/drei'`) will fail if `react-dom` isn't installed in the app. This error is confusing (cause it's not directly connected with the import), plus `react-dom` is an optional dependency and root imports should not require optional deps IMO.

### What

I've used a `require` call that will silently fail if no `react-dom` is installed. However when Html component will be used, it'll throw warnings to the console telling the user to install it

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged

It's a draft cause I'm not entirely sure it'll work in all environments (although I've used this technique previously, e.g. in presentation library https://github.com/saitonakamura/presa/blob/master/packages/blocks/src/code/index.tsx#L22) and I'm not sure whether it's too hacky and we need to find a different approach

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react'
-import * as ReactDOM from 'react-dom'
 import { Vector3, Group, Object3D, Matrix4, Camera, PerspectiveCamera, OrthographicCamera, Raycaster } from 'three'
 import { Assign } from 'utility-types'
 import { ReactThreeFiber, useFrame, useThree } from '@react-three/fiber'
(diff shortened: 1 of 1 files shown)


saitonakamura:
Here's a codesandbox with the current latest, when it fails
https://codesandbox.io/s/hungry-wind-k8obp?file=/src/index.tsx

And here's a codesandbox with a build from the PR
https://codesandbox.io/s/ground-reflections-and-video-textures-forked-vlw3t?file=/src/index.js
It doesn't fail and even if you import and render Html, it'll still not fail but will show a warning in the console (we also have he options to fail when rendering btw)

joshuaellis:
I think it probably needs to fail if you try use HTML without having `react-dom` installed. Just having an error is a little too subtle.

saitonakamura:
Ok, done, example https://codesandbox.io/s/strange-bhaskara-x7ixo?file=/src/index.js, comment out Html to see the cube working

joshuaellis:
Does this need extending to the other component in `web`

saitonakamura:
The other components in web doesn't use `react-dom` import so I believe it's not

joshuaellis (approved):

DoctypeRosenthal:
I have an issue with this fix. See https://github.com/pmndrs/drei/issues/557.

PR #526, opened 2021-08-26
Feat/add support for multiple selector elements in PointerLockControls

atmasark:
### Why
PointerLockControls already supports an optional `selector` prop, which allows PointerLockControls to lock by clicking other elements than document. The problem with current solution is that it's not possible to use multiple elements as selectors.

### What
This PR expands the current functionality by adding click event listeners to all elements matching the given selector, not just one. 

### Checklist

- [x] Documentation updated
- [x] Storybook entry updated
- [x] Ready to be merged

### DIFF
--- a/src/core/PointerLockControls.tsx
+++ b/src/core/PointerLockControls.tsx
@@ -44,9 +44,11 @@ export const PointerLockControls = React.forwardRef<PointerLockControlsImpl, Poi
 
     React.useEffect(() => {
       const handler = () => controls?.lock()
-      const element = selector ? document.querySelector(selector) : document
-      element && element.addEventListener('click', handler)
-      return () => (element ? element.removeEventListener('click', handler) : undefined)
+      const elements = selector ? Array.from(document.querySelectorAll(selector)) : [document]
+      elements.map((element) => element && element.addEventListener('click', handler))
+      return () => {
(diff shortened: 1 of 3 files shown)


src/core/PointerLockControls.tsx
```diff
@@ -44,9 +44,11 @@ export const PointerLockControls = React.forwardRef<PointerLockControlsImpl, Poi
 
     React.useEffect(() => {
       const handler = () => controls?.lock()
-      const element = selector ? document.querySelector(selector) : document
-      element && element.addEventListener('click', handler)
-      return () => (element ? element.removeEventListener('click', handler) : undefined)
+      const elements = selector ? Array.from(document.querySelectorAll(selector)) : [document]
+      elements.map((element) => element && element.addEventListener('click', handler))
```
joshuaellis, line 48: why are we mapping? We could just `forEach` it? same with the return function.

joshuaellis (changes requested):

src/core/PointerLockControls.tsx
atmasark: Hmm, I decided to use mapping as I've always thought it to be faster (which was apparently a lie) and just more elegant solution.

Learned something new today and changed it to forEach :)

PR #542, opened 2021-09-16
feat(Stage) - Add optional contactShadowPosition prop to Stage

SaraVieira:
### Why

I ran into some issues with the stage shadows when creating an animated character as it went through the shadow:

CleanShot 2021-09-16 at 23 10 05@2x

### What

Add a prop that allows you to change the position of the contact shadow

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/Stage.tsx
+++ b/src/core/Stage.tsx
@@ -40,6 +40,7 @@ type Props = JSX.IntrinsicElements['group'] & {
   shadowBias?: number
   contactShadowBlur?: number
   contactShadowOpacity?: number
+  contactShadowPosition?: [x: number, y: number, z: number]
 }
 
 export function Stage({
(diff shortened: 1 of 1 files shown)


SaraVieira:
Will add docs if/when it gets merged

src/core/Stage.tsx
```diff
@@ -54,6 +55,7 @@ export function Stage({
   shadowBias = 0,
   contactShadowBlur = 2,
```
joshuaellis, line 56: do you think we should move the `contactShadowX` props to an object instead of individual props? I'm concerned the prop list will get out of hand... then you could provide any of the valid `ContactShadow` props...

Issue #551, opened 2021-09-22
useHelper doesn't work with RectAreaLightHelper


Return180bpm:
- `three` version: 0.132.2
- `@react-three/fiber` version: 7.0.7
- `@react-three/drei` version: 7.8.5
- `node` version: v12.16.2
- `npm` (or `yarn`) version: npm@7.6.1

### Problem description:

Using RectAreaLightHelper with useHelper causes a crash in Chrome: 
> useHelper.js:24 Uncaught TypeError: helper.current.update is not a function

That sucks.

### Relevant code:

As suggested [in the drei useHelper doc](https://docs.pmnd.rs/drei/misc/use-helper) and in the relevant Storybook:

```js
import { RectAreaLightHelper } from "three/examples/jsm/helpers/RectAreaLightHelper";

  const RectArealightWithHelper = ({ pos, width, height, color, intensity }) => {
    const rectAreaLight = useRef();
    useHelper(rectAreaLight, RectAreaLightHelper);

    return (
        <rectAreaLight
            ref={rectAreaLight}
            position={pos}
            width={width}
            height={height}
            color={color}
            intensity={intensity}
            castShadow
        />
    );
};
```
[Codesandbox](https://codesandbox.io/s/react-three-fiber-rectarealighthelper-demo-cd3rw?file=/src/App.js:1185-1482)

### Suggested solution:

I posted a fix using regular THREE syntax [here.](https://stackoverflow.com/questions/65031201/rectarealighthelper-not-moving-in-react-three-fiber/69285232#69285232)

Here's what I found out about this bug: 
it makes sense Chrome would throw an error like this

 > useHelper.js:24 Uncaught TypeError: helper.current.update is not a function

 since the RectAreaLightHelper class as defined in RectAreaLightHelper.js doesn't have a method called update. Instead, it has a method `updateMatrixWorld`.
If I change this method's name to `update`, it no longer crashes, and the helper appears on screen.
Unfortunately, it doesn't map on to the actual the rectAreaLight object on the screen and instead seems to be stuck at some position in space.
(I wish I could fix this, but I don't even know where to start.)

CodyJasonBennett:
I've made a fix in #552. See if this resolves your issue:
 - https://codesandbox.io/s/amazing-ives-vmq9z?file=/src/App.js

PR #552, opened 2021-09-22
fix(useHelper): skip updates on static helpers

CodyJasonBennett:
### Why

`useHelper` currently breaks when using static helpers which don't expose update methods (as seen in #551).

### What

This PR updates `useHelper` to ignore updates for helpers which don't supply update methods.

### Sandbox

Here's a recreation of the issue with changes from this PR: https://codesandbox.io/s/amazing-ives-vmq9z?file=/src/App.js.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/useHelper.tsx
+++ b/src/core/useHelper.tsx
@@ -28,6 +28,9 @@ export function useHelper<T>(object3D: React.MutableRefObject<Object3D | undefin
   useFrame(() => {
     if (helper.current?.update) {
       helper.current.update()
+    } else if (helper.current && object3D.current) {
+      helper.current.position.copy(object3D.current.position)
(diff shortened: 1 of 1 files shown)


CodyJasonBennett:
When updating the sandbox in #551, I'm noticing that the helper still doesn't inherit its parent's transforms:
 - https://codesandbox.io/s/flamboyant-ride-kdsg7

CodyJasonBennett:
Maybe `useHelper` should return a reference to the appended helper?

joshuaellis (approved):
Great fix! Thank you 💪🏼

Issue #499, opened 2021-07-29
Environment black for other canvases than the first


robclouth:
### Versions
"@react-three/drei": "7.2.2",
"@react-three/fiber": "7.0.6",
"three": "0.130.1",

### Problem description:

When having multiple canvases, only the first instance of Environment works. Subsequent ones are black. 

### Code
https://codesandbox.io/s/heuristic-tu-9q8o4?file=/src/index.js

CodyJasonBennett:
I've fixed this in #554, but I'm working on including resolution params in that PR.

Sandbox here: https://codesandbox.io/s/pensive-kirch-uc7s5

drcmda:
this is not a problem in newer react: https://codesandbox.io/s/the-difference-between-color-modes-dus84

env maps cannot be re-used among canvases and since env is cached in drei it creates problem. in react 18 each canvas has a cache bound, problem solved.

@CodyJasonBennett is prem needed? it is in useasset because prem takes so long that it creates a glitch. as in the model appears without env, then it calculates, then it gets applied. with suspense it happens in lock step which imo is important. maybe threejs has changed and prem isnt a necessity any longer but if it still takes time i think suspense is more important than the multi canvas usecase - which will 100% be solved by the react team in just a few weeks.

CodyJasonBennett:
@drcmda, I'm not sure. I tried removing it to solve both issues but will have to take a closer look at PMREM.

srcnalt:
Using the Environment component from #554 fixed the issue for me to display it on multiple canvases, however now on mobile devices I get no light at all. Will investigate more why this happens.

CodyJasonBennett:
I believe PMREM was contributing lighting information. Maybe I should be focusing on the background's implementation with three.

tigrik2017:
Still valid. In my case, it occurs not on navigation, but moving between two same 3D viewer components on the same page which contain environment components, making them visible one by one. In that case, the first one loses the environment and after hiding the second, it won't restore. I'm using the latest react v17.0.2

tigrik2017:
Still valid. In my case, it occurs not on navigation, but moving between two same 3D viewer components on the same page which contain environment components, making them visible one by one. In that case, the first one loses the environment and after hiding the second, it won't restore.

versions installed:
react: 17.0.2
drei: 8.6.1
fiber: 7.0.24
three: 0.136.0

w3ichen:
As of Aug 19 2022:

I tried #554 but no luck.
[...]

Issue #461, opened 2021-06-27
Bad Quality Drei Environment HDRI files


him666:
- `three` version: ^0.129.0
- `@react-three/fiber` version: ^7.0.1
- `@react-three/drei` version: ^6.2.0
- `node` version: v14.10.0

- `npm` (or `yarn`) version: yarn 1.22.5

### Problem description:

environment hdri files resolution is being downscaled.

### Relevant code:

```js
<Environment
            files={'/environments/preset.hdr'} 
            background
          />
```

### Suggested solution:

adding prop to receive resolution 
res="1k" or 256, 512
whatever suits better

```js
<Environment
            files={'/environments/preset.hdr'} 
            background
            resolution={512}
          />
```

joshuaellis:
Thanks @him666 feel free to make a PR for this!

magicspon:
Any updates on this...
the difference is quite shocking:
r3f: https://eq-site-c0nedilrs-spon.vercel.app/demo 
vanilla: https://mango-dentists.surge.sh/

gsimone:
Note that the Environment component is meant to generate a cube texture for environmental lighting, not to use as a cube background. The texture generated from the PMREMGenerator is always 256x256  https://threejs.org/docs/#api/en/extras/PMREMGenerator

drcmda:
i think for envmap lighting it makes sense to keep resolution as low as possible, as @gsimone said, we just use prem. there should be zero difference between vanilla and r3f if you use the same means to get your environment. but maybe environment can be extended with a resolution prop and then not use prem? but i know too little about it, three has changed drastically in that area and i didnt keep up.

PR #554, opened 2021-09-22
feat(Environment): custom resolution

CodyJasonBennett:
### Why

Environment maps are very pixelated and unrealistic (see #461). This PR investigates solutions around that.

### What

I've refactored the component to drop PMREMGenerator, fixing #499 ([see sandbox](https://codesandbox.io/s/pensive-kirch-uc7s5?file=/src/index.js)). Currently, I am looking into solutions for upscaling the environment without a major refactor in three.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged


CodyJasonBennett:
This will require some work in three if we don't implement our own solution.

PR #572, opened 2021-10-08
feat(Points): add attributes prop

CodyJasonBennett:
### Why

Drei's `<Points />` doesn't accept geometry attributes, so you can only set them imperatively:

```jsx
const pointsRef = useRef()

useEffect(() => {
  pointsRef.current.geometry.setAttribute( ... )
}, [])

return <Points ref={pointsRef} />
```

### What

This PR adds an `attributes` prop to specify `<bufferAttribute />`.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/core/Points.tsx
+++ b/src/core/Points.tsx
@@ -11,6 +11,7 @@ type Api = {
 type PointsProps = JSX.IntrinsicElements['points'] & {
   range?: number
   limit?: number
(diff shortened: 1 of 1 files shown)


CodyJasonBennett:
I'm not sure what you mean. Can we set custom attributes declaratively with `attachObject={["geometry-attributes", "foo"]}`?

PR #613, opened 2021-11-02
feat: add fast declarative segments with single geometry.

flavioschneider:
### Why

* Using many separated Line components is inefficient, as each one of them uses a new geometry.
* Line components cannot be separated into segments (to share the geometry).

### What

* Proposed are 2 components: Segments (a wrapper), and Segment, a component to add the segment to the geometry. 
* Segments/Segment components share the geometry in a declarative way (similar to Instances/Instance but for segments).
```jsx
<Segments limit={1000}>
  <Segment start={[0,0,0]} end={[10,10,0]} color={[1,1,1]} />
</Segments>
```

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/.storybook/stories/Segments.stories.tsx
+++ b/.storybook/stories/Segments.stories.tsx
@@ -1,0 +1,88 @@
+import * as React from 'react'
+import { Vector3 } from 'three'
+import { hilbert3D } from 'three-stdlib'
+import { withKnobs, number, color, boolean } from '@storybook/addon-knobs'
+
+import { Setup } from '../Setup'
+
+import { Segment, Segments, OrbitControls } from '../../src'
+import { useFrame } from '@react-three/fiber'
+
+export default {
+  title: 'Performance/Segments',
+  component: Segments,
+}
+
+const points = hilbert3D(new Vector3(0), 5).map((p) => [p.x, p.y, p.z]) as [number, number, number][]
+
+const colors = new Array(points.length).fill(0).map(() => [Math.random(), Math.random(), Math.random()]) as [
+  number,
+  number,
+  number
+][]
+
+export function BasicSegments() {
+  return (
+    <>
+      <Segments limit={6} lineWidth={2.0}>
+        <Segment start={[0, 0, 0]} end={[10, 0, 0]} color={[0, 0, 1]} />
+        <Segment start={[0, 0, 0]} end={[0, 10, 0]} color={[0, 1, 1]} />
+        <Segment start={[0, 0, 0]} end={[0, 0, 10]} color={[1, 0, 1]} />
+        <Segment start={[0, 0, 0]} end={[-10, 0, 0]} color={[1, 0, 0]} />
+        <Segment start={[0, 0, 0]} end={[0, -10, 0]} color={[0, 1, 0]} />
+        <Segment start={[0, 0, 0]} end={[0, 0, -10]} color={[1, 1, 0]} />
+      </Segments>
+      <OrbitControls />
+    </>
+  )
+}
+BasicSegments.storyName = 'Basic'
+
+BasicSegments.decorators = [
+  withKnobs,
+  (storyFn) => (
+    <Setup controls={false} cameraPosition={new Vector3(10, 10, 10)}>
+      {storyFn()}
+    </Setup>
+  ),
+]
+
+function Spinner({ delay }) {
+  const ref = React.useRef({ start: [0, 0, 0], end: [1, 1, 1], color: [1, 1, 1] })
+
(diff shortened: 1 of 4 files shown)


drcmda:
This is nice!

.storybook/stories/Segments.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useFrame } from '@react-three/fiber'
+
+export default {
+  title: 'Performance/Segments',
+  component: Segments,
+}
+
+const points = hilbert3D(new Vector3(0), 5).map((p) => [p.x, p.y, p.z]) as [number, number, number][]
+
+const colors = new Array(points.length).fill(0).map(() => [Math.random(), Math.random(), Math.random()]) as [
```
gsimone, line 18: You can write this as 

```
const colors = Array.from({ length: points.length }, () => [Math.random(), Math.random(), Math.random()])```

File: .storybook/stories/Segments.stories.tsx [hunk 2]
```diff
@@ -0,0 +1,88 @@
...
+  return <Segment ref={ref} />
+}
+
+export function ManySegments() {
+  return (
+    <>
+      <Segments limit={10000} lineWidth={0.1}>
+        {Array(10000)
+          .fill(0)
+          .map((_, i) => (
```
gsimone, line 71: See prev comment about creating the array

src/core/Segments.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  useFrame(() => {
+    positions.length = segments.length * 6
+    colors.length = segments.length * 6
+    segments.forEach((segmentRef, i) => {
+      const segment = segmentRef.current
+      if (segment) {
+        const segmentStart = arrPos(segment.start)
+        const segmentEnd = arrPos(segment.end)
+        const segmentColor = arrColor(segment.color)
```
gsimone, line 59: This doesn't work when passing a string color

src/core/Segments.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const ref = React.useRef<Segment>({ start: [0, 0, 0], end: [0, 0, 0], color: [1, 1, 1] })
+  React.useLayoutEffect(() => subscribe(forwardedRef || ref), [])
+  React.useLayoutEffect(() => {
+    var currRef = forwardedRef || ref
+    start && Object.assign((currRef as React.RefObject<Segment>).current, { start })
+    end && Object.assign((currRef as React.RefObject<Segment>).current, { end })
+    color && Object.assign((currRef as React.RefObject<Segment>).current, { color })
+  }, [start, end, color])
+  return null
+})
```
gsimone, line 101: I'd add a warning to only use this component inside the parent context

gsimone (changes requested):
This is 100% awesome! Left you a couple of minor comments

gsimone (commented):
I also noticed that you can't do

```
<Segment ref={ref} />
```

it would be cool if it failed gracefully and gave a warning

.storybook/stories/Segments.stories.tsx
flavioschneider: That's great, however those two lines aren't used at all. I should remove them.

flavioschneider:
> I also noticed that you can't do
> 
> ```
> <Segment ref={ref} />
> ```
> 
> it would be cool if it failed gracefully and gave a warning

It should be possible to use a reference only (see the story example), however it gives a TS warning. Maybe someone can help me with this one? I'm quite new to TS.

src/core/Segments.tsx
flavioschneider: It currently works the same way as vertexColors in the Line component. But yes it would be nice to pass strings as well.

gsimone:
> > I also noticed that you can't do
> > ```
> > <Segment ref={ref} />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it would be cool if it failed gracefully and gave a warning
> 
> It should be possible to use a reference only (see the story example), however it gives a TS warning. Maybe someone can help me with this one? I'm quite new to TS.

I think we can just leave it with the TS warning ( @ts-expect-error or @ts-ignore ) and add an issue so that someone can pick it up and fix it

gsimone (approved):

gsimone:
Merged, thank you so much for this and grats on your first contribution!

flavioschneider:
> Merged, thank you so much for this and grats on your first contribution!

Thank you! :)

PR #616, opened 2021-11-04
fix: cache DRACOLoader instance in useGLTF

### LINKED ISSUE #615: useGLTF with multiple Draco models fails to load in iOS
### Problem description:

Multiple DRACOLoader instances are known to cause problems on iOS: https://github.com/mrdoob/three.js/issues/22445

The useGLTF hook seems to be creating a new instance every time the hook is used.

### Suggested solution:

Could we cache the DRACOLoader instance as a local variable in the hook and check if it's already defined?

ffdead:
### Why

Fix #615 - Multiple DRACOLoader instances are known to cause problems on iOS: mrdoob/three.js#22445

### What

* dracoLoader instance is now cached using local variable in `useGLTF`
* updated draco lib CDN version to 1.4.3
* fixed storybook types

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/useGLTF.tsx
+++ b/src/core/useGLTF.tsx
@@ -3,13 +3,17 @@ import { Loader } from 'three'
 import { GLTFLoader, DRACOLoader, MeshoptDecoder } from 'three-stdlib'
 import { useLoader } from '@react-three/fiber'
 
+let dracoLoader: DRACOLoader | null = null
+
 function extensions(useDraco: boolean | string, useMeshopt: boolean, extendLoader?: (loader: GLTFLoader) => void) {
   return (loader: Loader) => {
     if (extendLoader) {
       extendLoader(loader as GLTFLoader)
     }
     if (useDraco) {
-      const dracoLoader = new DRACOLoader()
+      if (!dracoLoader) {
+        dracoLoader = new DRACOLoader()
+      }
       dracoLoader.setDecoderPath(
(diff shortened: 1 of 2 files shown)


src/core/useGLTF.tsx
```diff
@@ -3,15 +3,19 @@ import { Loader } from 'three'
...
       extendLoader(loader as GLTFLoader)
     }
     if (useDraco) {
-      const dracoLoader = new DRACOLoader()
+      if (!dracoLoader) {
+        dracoLoader = new DRACOLoader()
+      }
       dracoLoader.setDecoderPath(
-        typeof useDraco === 'string' ? useDraco : 'https://www.gstatic.com/draco/versioned/decoders/1.4.0/'
+        typeof useDraco === 'string' ? useDraco : 'https://www.gstatic.com/draco/versioned/decoders/1.4.3/'
```
drcmda, line 18: just making sure, certain that 1.4.3 is ok? they broke practically everything in a recent update and i don't trust their patch updates. can we be sure this is safe?

ffdead: Great question! We were discussing this here https://github.com/google/draco/issues/751

The versioned URLs are recommended and 1.4.3 is the latest.

I would expect useGLTF to pull the latest if not overriding it with my own local version 🤷‍♂️

drcmda (approved):

hengjiUSTC:
I am also getting multiple KTX loaders warning when using `useKTX2`. Should the patten allied to other loaders as well?

PR #627, opened 2021-11-14
feat: adds buffer version of Points component

gsimone:
👉  **PREVIEW https://drei-git-gsimone-points-buffer-pmndrs.vercel.app/?path=/story/performance-pointsbuffer--basic-points-buffer** 👈 

### TODO

- [x] Add README


drcmda:
i think points is nearly the same thing, we should maybe re-purpose it.

```jsx
function Points({ positions, stride = 3, ........, ...props }) {
    const internalPoints = ...
    const internalColors = ...
    return (
      <points matrixAutoUpdate={positions ? true: false} ref={mergeRefs([ref, parentRef])} {...(positions ? null : events)} {...props}>
        <bufferGeometry>
          <bufferAttribute
            attachObject={['attributes', 'position']}
            count={positions.length / 3}
            array={positions ? positions : internalPositions}
            itemSize={stride}
            usage={THREE.DynamicDrawUsage}
          />
          <bufferAttribute
            attachObject={['attributes', 'color']}
            count={colors.length / 3}
            array={colors ? colors: internalColors}
            itemSize={stride}
            usage={THREE.DynamicDrawUsage}
          />
        </bufferGeometry>
        {positions ? null : <context.Provider value={api}>{children}</context.Provider>}
      </points>
```

src/core/Points.tsx
```diff
@@ -121,4 +137,77 @@ const Point = React.forwardRef(({ children, ...props }, ref) => {
...
+
+export const Points = React.forwardRef<THREE.Points, PointsBuffersProps | PointsInstancesProps>(
+  (props, forwardedRef) => {
+    if ((props as PointsBuffersProps).positions instanceof Float32Array) {
+      return <PointsBuffer {...(props as PointsBuffersProps)} ref={forwardedRef} />
+    }
+
+    return <PointsInstances {...(props as PointsInstancesProps)} ref={forwardedRef} />
+  }
+)
```
gsimone, line 213: @drcmda the switch is based on positions being passed as a f32 array, feel free to change this to something else 🤷

gsimone:
Also added `size` to the instance-based points, since it's fairly often used

Issue #621, opened 2021-11-10
useNormalTexture wrap mode not working


ffdead:
- `@react-three/drei` version: 7.12.5

### Problem description:

After changing texture `repeat`, the texture is wrapped using the default `THREE.ClampToEdgeWrapping` instead of `THREE.RepeatWrapping`

This seems to be caused by 742a067392f80f57d836447c46841b001b2e2471 since textures are now premtively uploaded to the GPU.

This probably breaks all current usages of `useTexture` where the texture is changed before rendering. 

### Relevant code:

```js
const [normalMapPrimary] = useNormalTexture(
    42, // index of the normal textur
    // second argument is texture attributes
    {
      repeat: [1.5, 1.5]
    }
);
```

### Suggested solution:

1. Change useNormalTexture to set `normalTexture.needsUpdate = true;` to re-upload it to the GPU
2. Revert the change in `useTexture` as the user might want to change the texture, or want to control when that initial upload jank happens (or at least make it optional somehow to avoid double uploads) **(RECOMMENDED)**

joshuaellis:
I think the opt-out solution is probably more than ideal as I suspect if `three` are doing opt-in automatically (may be wrong there) it's because it's more useful to first comers as opposed to experienced devs. 

@drcmda you made the change, What do you think?

ffdead:
In my experience `initTexture` needs to be combined with `compile` and traversing the tree to make all objects visible in order to avoid scroll jank. But I'm happy to keep it as long as we can opt-out.

Please see the suggested fix in #629

drcmda:
@ffdead could you quickly detail how the texture is changed? because the init runs inside a useEffect. that should leave the user time to either mutate it in useMemo or useLayoutEffect. what makes it break eventually?

@joshuaellis the issue was that in threejs the texture upload take a long time. and it doesn't event happen in compile. so you have some objects that aren't in view, but when they are they cause the mainthread to freeze. i asked them to please fix compile but they didn't want to. i then asked if initializing textures is safe and they seemed to affirm.

ffdead:
> @ffdead could you quickly detail how the texture is changed? because the init runs inside a useEffect. that should leave the user time to either mutate it in useMemo or useLayoutEffect. what makes it break eventually?

This is the code use `useNormalTexture`

```
const normalTexture = useTexture(url, { init: false }) as Texture

  React.useEffect(() => {
    if (!normalTexture) return
[...]

PR #630, opened 2021-11-15
Fix/normal texture repeat

ffdead:
### Why

Fix for #621

### What

Changed `useEffect` to `useLayoutEffect` in `useNormalTexture` in order to set texture repeat before it is uploaded to GPU.

### Checklist

- [x] Ready to be merged


ffdead:
@drcmda murge?

Issue #679, opened 2021-12-10
Set className on outer element in drei HTML


joshon:
### Describe the feature you'd like:

I would like to be able to set a selector on the outermost div created by Drei HTML. In [my case I want it so that I can control the z-index manually](https://github.com/pmndrs/drei/discussions/660) - but I think it would be useful to expose in general. Currently, {...divProps} applies to the child div, I would like to be able to access the parent. You can access the parent through the "as" property which sets the element type, but that seems like a hack simply for CSS selector purposes.

### Suggested implementation:

Perhaps adding a property called wrapperClassName that added a className to the outermost div would be a good enhancement.

joshuaellis:
resolved by #690

PR #690, opened 2021-12-12
feat: add wrapperClass to HTML

joshon:
### Why

This adds the ability to set a class name on the outer element so that it can be addressed by CSS selectors.  This is in response to [this feature request](https://github.com/pmndrs/drei/issues/679) that I made. 

The use case that prompted me to make this change was that when I select the 3D element the HTML corresponds to, I want the z-index of that element to come to the top layer no matter what.

### What

I added a wrapperClass attribute to HTML, which takes a string. 
A useEffect hook updates the className on the main el.

### Checklist

- [ x] Documentation updated
- [ x] Ready to be merged

Thanks, this is my first ever pull request on Github, let alone drei - so please help guide me through this if I have done something wrong. One question I had is whether I should clean up in the useEffect with a return, it didn't seem necessary to me - but happy to be proved wrong on that.

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -176,7 +176,7 @@ export const Html = React.forwardRef(
       }
     }, [target, transform])
 
-    React.useEffect(() => {
+    React.useLayoutEffect(() => {
       if (wrapperClass) el.className = wrapperClass
     }, [wrapperClass])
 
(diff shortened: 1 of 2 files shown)


src/web/Html.tsx
```diff
@@ -174,6 +176,10 @@ export const Html = React.forwardRef(
       }
     }, [target, transform])
 
+    React.useEffect(() => {
```
drcmda, line 179: this should imo be useLayoutEffect or else it comes a frame too late.

drcmda (changes requested):

src/web/Html.tsx
joshon: Thanks, I updated it - I hope that I used the correct git commit syntax

PR #699, opened 2021-12-17
fix: handle instance events at the raycast

drcmda:
https://codesandbox.io/s/floating-instanced-shoes-forked-u6z5e?file=/src/App.js:786-949

### Why

Currently raycast on drei/instances are a forwarding hack, the main instance collects event props from the instances and forwards click handlers back into them. This has several disadvantages, there is bubbling, it cannot be grouped, overall it is not part of r3f's event system and merely works as an appendage.

### What

This solution moves events into the individual instances raycast method, which takes the parents (THREE.InstancedMesh) geometry and tests the ray against it. Now these events behave like any other.

### What about bounds?

Instances normally do not have bounds because they are virtual, only the main THREE.InstancedMesh has bounds. The instance is an empty object3d with nothing in it, it's just being used to track position. By adding the `geometry` getter we allow THREE.Box3.setFromObject to pick up on individual instance bounds.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/Instances.tsx
+++ b/src/core/Instances.tsx
@@ -47,7 +47,7 @@ const Instance = React.forwardRef(({ context, children, ...props }: InstanceProp
   )
 })
 
-const Instances = React.forwardRef<THREE.InstancedMesh, InstancesProps>(
+const Instances = React.forwardRef<InstancedMesh, InstancesProps>(
   ({ children, range, limit = 1000, frames = Infinity, ...props }, ref) => {
     const [{ context, instance }] = React.useState(() => {
       const context = React.createContext<Api>(null!)
--- a/src/core/Instances.tsx
+++ b/src/core/Instances.tsx
@@ -47,11 +47,8 @@ const Instance = React.forwardRef(({ context, children, ...props }: InstanceProp
   )
 })
 
-const Instances = React.forwardRef(
-  (
-    { children, range, limit = 1000, frames = Infinity, ...props }: InstancesProps,
-    ref: React.ForwardedRef<THREE.InstancedMesh>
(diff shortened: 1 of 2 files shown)


src/core/Instances.tsx
```diff
@@ -132,10 +119,11 @@ const Instances = React.forwardRef(
 
     return (
       <instancedMesh
+        userData={{ instances }}
         matrixAutoUpdate={false}
-        ref={mergeRefs([ref, parentRef])}
+        ref={mergeRefs([ref as any, parentRef])}
```
joshuaellis, line 124: you shouldn't need to do this. What happens if you type `forwardRef` in a different way:

```ts
const Instances = React.forwardRef<THREE.InstancedMesh, InstancesProps>(
  (
    { children, range, limit = 1000, frames = Infinity, ...props },
    ref
  )
```

`any` is bad practice imo

drcmda: i don't know why, it works locally, it crashes on the vercel ci. tsc ran without problems here.

drcmda: btw here is the fallout https://vercel.com/pmndrs/drei/HYTZjPhdCzNsbgsZpon9RvSUJQUZ

drcmda: @joshuaellis tried your suggestion but nothing: https://vercel.com/pmndrs/drei/P7fEUjFYSSRgLUgkps5K8dMakAcS

locally it works.

drcmda: removing the "THREE." did the trick

natarius (commented):
Looks great!

PR #701, opened 2021-12-21
fix: always use current default camera in MeshReflectorMaterial

unframework:
Hey folks, just a tiny fix for a useCallback dependencies issue.

### Why

Was trying to use this with my own PerspectiveCamera + makeDefault prop, and the reflection was not updating.

### What

Added latest  `camera` value to useCallback deps for beforeRender.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/MeshReflectorMaterial.tsx
+++ b/src/core/MeshReflectorMaterial.tsx
@@ -140,7 +140,7 @@ export const MeshReflectorMaterial = React.forwardRef<MeshReflectorMaterialImpl,
       projectionMatrix.elements[6] = clipPlane.y
       projectionMatrix.elements[10] = clipPlane.z + 1.0
       projectionMatrix.elements[14] = clipPlane.w
-    }, [])
+    }, [camera])
 
     const [fbo1, fbo2, blurpass, reflectorProps] = React.useMemo(() => {
       const parameters = {


gsimone:
Could you make a quick codesandbox example with this behaviour so we can see it failing and working correctly? You can add the codesandbox link to this PR and it will build with the new version

drcmda:
> Could you make a quick codesandbox example with this behaviour so we can see it failing and working correctly? You can add the codesandbox link to this PR and it will build with the new version

i think it makes sense, the camera is being used to calculate the streaks, if it's missing in the dep array it's stale. someone posted a repro on discord. this can safely be merged. :)

joshuaellis:
Don't we have a lint rule that's supposed to pick this up? Feels odd it's been missed at time of creation...

unframework:
> > Could you make a quick codesandbox example with this behaviour so we can see it failing and working correctly? You can add the codesandbox link to this PR and it will build with the new version
> 
> i think it makes sense, the camera is being used to calculate the streaks, if it's missing in the dep array it's stale. someone posted a repro on discord. this can safely be merged. :)

I tried to repro in this sandbox (with Drei v8.2.1): https://codesandbox.io/s/ragdoll-physics-forked-ns8dh?file=/src/App.js. What is odd is that in this case reflections do update correctly at first, but when I tweak the sandbox code (i.e. cause a new camera instance to spawn in live editor) the reflections (well, the angle) stop updating:

Screenshot 2021-12-21 120001

On my local environment it just did not update to begin with.

Anywho, thanks for the quick merge!

drcmda:
> > > Could you make a quick codesandbox example with this behaviour so we can see it failing and working correctly? You can add the codesandbox link to this PR and it will build with the new version
> > 
> > 
> > i think it makes sense, the camera is being used to calculate the streaks, if it's missing in the dep array it's stale. someone posted a repro on discord. this can safely be merged. :)
> 
> I tried to repro in this sandbox (with Drei v8.2.1): https://codesandbox.io/s/ragdoll-physics-forked-ns8dh?file=/src/App.js. What is odd is that in this case reflections do update correctly at first, but when I tweak the sandbox code (i.e. cause a new camera instance to spawn in live editor) the reflections (well, the angle) stop updating:
> 
> Screenshot 2021-12-21 120001
> 
> On my local environment it just did not update to begin with.
> 
> Anywho, thanks for the quick merge!

there's still a flaw, the geometry has to face 0/0/1 for the streaks to come out right. you could not, for instance, just use blender and gltfjsx and then slap meshreflectormaterial over some mesh because the orientation might be off. if someone knows a solution for that it would be a relief!

PR #703, opened 2021-12-23
TransformControls: Override the domElement

RenaudRohlinger:
Just like in OrbitControls through the connect method, this PR allows specifying another domElement with TransformControls.

### DIFF
--- a/src/core/TransformControls.tsx
+++ b/src/core/TransformControls.tsx
@@ -57,7 +57,7 @@ export const TransformControls = React.forwardRef<TransformControlsImpl, Transfo
     const defaultCamera = useThree(({ camera }) => camera)
     const invalidate = useThree(({ invalidate }) => invalidate)
     const explCamera = camera || defaultCamera
-    const [controls] = React.useState(() => new TransformControlsImpl(explCamera, domElement || gl.domElement))
+    const controls = React.useMemo(() => new TransformControlsImpl(explCamera, domElement || gl.domElement), [explCamera, domElement, gl.domElement])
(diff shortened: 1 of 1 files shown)


src/core/TransformControls.tsx
```diff
@@ -56,7 +57,7 @@ export const TransformControls = React.forwardRef<TransformControlsImpl, Transfo
     const defaultCamera = useThree(({ camera }) => camera)
     const invalidate = useThree(({ invalidate }) => invalidate)
     const explCamera = camera || defaultCamera
-    const [controls] = React.useState(() => new TransformControlsImpl(explCamera, gl.domElement))
+    const [controls] = React.useState(() => new TransformControlsImpl(explCamera, domElement || gl.domElement))
```
joshuaellis, line 60: Should this be a `useMemo`? I know it wasn't the current PR decision to put it in `useState` but looking at it, what happens if someone changes the `camera` or `domElement`? I suppose the former is more likely... What do you think?

RenaudRohlinger: @joshuaellis No idea, the `[state] = useState()` is pretty commonly used in Drei's hooks. I'm really not familiar with this syntax. I guess nothing will happen if we change the `camera` or `domElement`.
`const controls = useMemo(() => new TransformControlsImpl(explCamera, domElement || gl.domElement), [explCamera, domElement])` would seems more appropriate.

RenaudRohlinger: It also feels wrong to not dispose anything in this component (`dispose={undefined}`). Or am I missing something?

PR #714, opened 2022-01-05
chore: apply best practices to Adaptive storybook

bjornstar:
### Why

Super minor, but I'd like to apply some best practices to our storybooks:

- Only import what you are using
- Use `import type` when only the type is being used
- Avoid casting (ie. `as`) where possible

The first two help with tree shaking and keeping bundle sizes down.
The latter is typically used when types are incorrect or incomplete and tends to cause runtime problems.

### What

- Added `storybook-static` to `.prettierignore` it's the output directory for `npm run storybook-build`
- Applied best practices to the Adaptive storybook

### Checklist

- [x] Ready to be merged

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,3 @@
 dist/
+storybook-static/
 *.typeface.json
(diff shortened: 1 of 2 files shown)


joshuaellis:
I agree with all your motives for this. Typecasting should be avoided at all costs, and so should using `any`.

joshuaellis (approved):

bjornstar:
Do you want to merge this? I don't have merge rights on this repo.

joshuaellis:
yes, thanks for the reminder!

PR #716, opened 2022-01-10
fix: Add cleanup fn in transformcontrols

miyamonz:
### Why

I found a bug below.

reproduction code:
https://codesandbox.io/s/transformcontrols-bug-dhx2s?file=/src/index.js

When TransformControls gets not children but `object` prop, the control doesn't detach object when the TransformControls is unmounted.

It causes a problem like this.

- There is a boolean state that handles whether or not to show the TransformControls.
- I want to set it true when the object is clicked, and set if false when clicked outside.
- the target object cannot be in the children of TransformControl. (for example, I want to control the rotation of bones in a humanoid model)

In such a case, you will be able to control it even though the state is false once it's enabled.

like this
[Image from Gyazo](https://gyazo.com/ec858be902011a3bc84af962d948a319)

### What

I thought detach method may not be called, found this useEffect block, and fixed it.
https://github.com/pmndrs/drei/blob/67ed4a448f594355fa99100871a4977276de96fc/src/core/TransformControls.tsx#L62-L65

### Checklist

- [x] Storybook entry added
  - I added the story that an object is passed via prop, but if you think it's meaningless, cut it down.
- [x] Ready to be merged

### DIFF
--- a/src/core/TransformControls.tsx
+++ b/src/core/TransformControls.tsx
@@ -60,10 +60,11 @@ export const TransformControls = React.forwardRef<TransformControlsImpl, Transfo
     const group = React.useRef<THREE.Group>()
 
     React.useLayoutEffect(() => {
-      if (!controls) return () => {}
-      if (object) controls.attach(object instanceof THREE.Object3D ? object : object.current)
-      else controls.attach(group.current as THREE.Object3D)
-      return () => controls.detach()
+      if (object) controls?.attach(object instanceof THREE.Object3D ? object : object.current)
+      else controls?.attach(group.current as THREE.Object3D)
+      return () => {
+        controls?.detach()
+      }
     }, [object, children, controls])
 
     React.useEffect(() => {
--- a/src/core/TransformControls.tsx
+++ b/src/core/TransformControls.tsx
@@ -60,10 +60,10 @@ export const TransformControls = React.forwardRef<TransformControlsImpl, Transfo
     const group = React.useRef<THREE.Group>()
 
     React.useLayoutEffect(() => {
-      if (object) controls?.attach(object instanceof THREE.Object3D ? object : object.current)
-      else controls?.attach(group.current as THREE.Object3D)
+      if (object) controls.attach(object instanceof THREE.Object3D ? object : object.current)
+      else controls.attach(group.current as THREE.Object3D)
       return () => {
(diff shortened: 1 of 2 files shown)


src/core/TransformControls.tsx
```diff
@@ -60,8 +60,10 @@ export const TransformControls = React.forwardRef<TransformControlsImpl, Transfo
     const group = React.useRef<THREE.Group>()
 
     React.useLayoutEffect(() => {
-      if (object) controls?.attach(object instanceof THREE.Object3D ? object : object.current)
-      else controls?.attach(group.current as THREE.Object3D)
+      if (!controls) return () => {}
```
drcmda, line 63: i think it would be better writing 

```jsx
if (controls) {
  if (object) controls.attach(object instanceof THREE.Object3D ? object : object.current)
  else controls.attach(group.current as THREE.Object3D)
  return () => controls.detach()
}
```

btw i don't understand the code, how can controls ever be undefined when it's defined as:

```jsx
const [controls] = React.useState(() => new TransformControlsImpl(explCamera, gl.domElement))
```

drcmda (changes requested):

src/core/TransformControls.tsx
RenaudRohlinger: @drcmda Hey! This part is actually introducing issues. I solved it by using useMemo and also added a way to override the dom element. https://github.com/pmndrs/drei/pull/703

miyamonz: why `if (!controls) return () => {}` is to avoid TS type error.

see:
image

this is because the callback type is `function(): (() => TransformControlsImpl<THREE.Camera | Camera>) | undefined
`.

But now that I think about it, this was simpler.

```
if (object) controls?.attach(object instanceof THREE.Object3D ? object : object.current)
      else controls?.attach(group.current as THREE.Object3D)
      return () => {
        controls?.detach()
      }
```

miyamonz: @RenaudRohlinger 
I don't noticed it is not nullish already because of your patch, thanks! so every `?` can be removed.

miyamonz:
Is there any other problem to fix?

miyamonz:
Oh, it's already fixed by #749

PR #722, opened 2022-01-14
feat: add reflectorOffset to MeshReflectorMaterial

### LINKED ISSUE #721: MeshReflectorMaterial should be able to reflect from a plane offset from the object's origin

The current `MeshReflectorMaterial` works great for mirrors that are just thin planes attached to a surface. However I have a pre-existing 3D model of a cabinet, where the front surface of the cabinet needs the reflective material. When applying this material to that front surface, it does become reflective, but the reflected image is generated from the object's origin, not from the reflective surface. I.e.:

```
|
|-----
|    |
A    B
|    |
|-----
|
```

- Point A is the original of the cabinet model, and is where it attaches to a wall
- Point B is the middle of the reflective surface of the cabinet.

I need the reflected image to be generated from point B, but it's currently always generated from point A. When looking front-on, the difference is not noticeable:

1- fine

When looking from side-on though, there are two problems. The below screenshots are looking at the mirror from its right-hand edge.

1. Objects can go missing from the reflection, or be in slightly the wrong position. The black tap should be visible at the far edge of the mirror here, but it isn't:

2-wrong-reflection

2. The reflection can get clipped altogether from some angles:

3-clipped-reflection

I have a fix for this already and will open a PR shortly.

camjackson:
### Why

Allow reflective surfaces to be offset from the object's origin. Resolves #721.

### What

Added a simple `reflectorOffse: number` prop to the MeshReflectorMaterial component. The `reflectorWorldPosition` is then translated that distance, in the normal (z) direction.

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### Screenshots

Refer to #721 for screenshots of the problems being solved. Here are corresponding screenshots showing the resolution:

1. When viewed from side-on, objects are now reflected perfectly in the offset mirror:

4-wrong-reflection-fixed

2. When view from side-on, the reflection is no longer clipped:

5-clipped-reflection-fixed

### DIFF
--- a/.storybook/stories/Reflector.stories.tsx
+++ b/.storybook/stories/Reflector.stories.tsx
@@ -3,7 +3,7 @@ import { useFrame } from '@react-three/fiber'
 import { Vector3, Mesh, RepeatWrapping, Vector2 } from 'three'
 
 import { Setup } from '../Setup'
-import { MeshReflectorMaterial, useTexture, TorusKnot, Box } from '../../src'
+import { MeshReflectorMaterial, useTexture, TorusKnot, Box, Environment } from '../../src'
 
 export default {
   title: 'Shaders/MeshReflectorMaterial',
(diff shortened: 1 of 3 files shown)


camjackson:
cc: @af-tw.

drcmda:
nice, thanks!

PR #736, opened 2022-01-24
fix: README broken storybook links

hazem3500:
### Why

Storybook links in the README were broken.

### What

I've changed broken storybook links with the right ones.


joshuaellis (approved):
Thanks for the contribution ⭐

PR #737, opened 2022-01-25
fix: use npm instead of three/examples

RenaudRohlinger:
### Why
The library stats.js is imported but we do not use it, instead, three/examples is being used


CodyJasonBennett (approved):
Looks good to me. I [created a sandbox](https://codesandbox.io/s/wizardly-parm-33e0l) to play with.

I'm a little wary of stats' main field, considering it's a UMD, but since it doesn't wrap require (and has no dependencies) it should play nice with bundlers.

PR #748, opened 2022-02-01
feat: add variation of meshBounds for instancedMeshes

jatinsandilya:
### Why

Currently meshBounds does not support meshes which are instanced. This PR adds a variation of meshBounds to work in this use case. 

### What

I've add a variation of meshBounds as a solution. Potentially this could be done in the meshBounds implementation itself so its debatable. Let me know if this or that approach makes sense. 

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/core/instancedMeshBounds.tsx
+++ b/src/core/instancedMeshBounds.tsx
@@ -1,0 +1,33 @@
+import { Position } from '../helpers/Position'
+import { Ray, Raycaster, Intersection, Vector3, Matrix4 } from 'three'
+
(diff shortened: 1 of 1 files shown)


gsimone:
Could you add a story for this?

gsimone (changes requested):
Could you please add a working story for this?

jatinsandilya:
seems like a stale PR and probably something that we dont need anymore. Closing it.

PR #756, opened 2022-02-09
removed sRGBToLinear from MeshReflectorMaterial

b4sh-3r:
### Why

To solve:
ERROR: 0:1323: 'sRGBToLinear' : no matching overloaded function found

### What

I removed the "sRGBToLinear" part highlighted by @paulhenschel from the MeshReflectorMaterial

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [X ] Ready to be merged


gsimone (changes requested):
You shouldn't remove all the lines, just the sRGBToLinear call

eg. 

```
diffuseColor = (vec4(vec3(depthFactor), 1.0));
```

drcmda:
imo the debug modes just prevent shader branching, no? i would be ok with them going out. i think i already removed them in some other shader.

FarazzShaikh:
close as stale?

PR #757, opened 2022-02-09
fix(core): cleanup RGBFormat/sRGB decode usage, bump three peer dep to r137

CodyJasonBennett:
Continues #742.

Removes leftover RGBFormat usage and bumps three's peer deps accordingly.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged


CodyJasonBennett:
Not sure if the reflected diffuse color is correct.

Sandbox thumbnail (pre r137):
image

Current (post r137):
image

PR #759, opened 2022-02-11
feat: Add to method to Bounds

krispya:
### Why

I needed to be able to send the camera to specific orientations when it was not focused on an object.

### What

The `to` method can be used to manually orient the camera with a `position` and `target`. The `target` is optional and uses whatever is currently in `box`, so it is affected by `refresh` if you want to focus an object or the Bounds group.

Here is an example of how I use it: https://codesandbox.io/s/bounds-to-example-mipjh?file=/src/App.js


FarazzShaikh:
This looks good to me, lets fix conflicts and merge?

joshuaellis (approved):

PR #761, opened 2022-02-12
feat: makes useHelper togglable by passing first prop as false

### LINKED ISSUE #334: Add a toggle to useHelper
`useHelper(ref, constructor, toggle)` so that it ca be easily turned off

gsimone:
```
useHelper(false,...) 
```

removes the helper from the scene, so helpers can now be toggled

- closes #334 

This is the simple solution to avoid a breaking change


joshuaellis (approved):

TresAbhi:
This makes my life so much easier. Thanks!

PR #762, opened 2022-02-12
feat: Automatic Camera Helper with 'helper' attribute

gsimone:
Instead of having to go through refs, you can now directly tag the camera to get an helper

```jsx
const ref = useRef()
useHelper(ref, CameraHelper)
...

<PerspectiveCamera ref={ref} />
```
is now
```jsx
<PerspectiveCamera helper />
```

Depends on #761

### DIFF
--- a/.storybook/stories/OrthographicCamera.stories.tsx
+++ b/.storybook/stories/OrthographicCamera.stories.tsx
@@ -1,11 +1,19 @@
 import * as React from 'react'
 import { Canvas } from '@react-three/fiber'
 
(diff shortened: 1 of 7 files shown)


DennisSmolek:
PR Name is the same as #761 
Suggest feat: Automatic Camera Helper  with 'helper'  attribute

gsimone:
@DennisSmolek thanks for catching the title, I must have pasted the other one 😓 

The helper should update automatically (it's an unconditioned useFrame). If we want to make the camera helper accessible from a ref, we could use `useImperativeHandle` to expose the helper's ref, probably. Is that what you want?

PR #769, opened 2022-02-16
feat: Add isActive prop to control Select activation

contactsimonwilson:
### Why
Selection doesn't provide enough activation control to be used for a wide range of applications. Some applications may want control over the modifier keys used, or even which mouse buttons activate the selection

### What

Expose a `isActive` prop that allows the component consumer to control when/how activation of the select occurs.

### Checklist
- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged


gsimone:
Really good change imo! Do I understand correctly that this keeps the current client-facing behaviour where shit is the key used to activate?

contactsimonwilson:
> Really good change imo! Do I understand correctly that this keeps the current client-facing behaviour where shit is the key used to activate?

Yes, exactly. This maintains full backwards compatibility but allows consumers to override the shift behavior. Exposing it with a callback function that provides the event also allows any other event data to be used (such as controlling which mouse button is used, other modifiers, etc)

contactsimonwilson:
Converted back to draft because after more consideration, the API could be even more flexible (onStart/onEnd)

cyango:
What's missing for this to be approved?

PR #770, opened 2022-02-18
fix: text fouc, use suspend-react instead of use-asset

drcmda:
### Why

`<Text />` previously loads async assets but wasn't integrated into suspense, therefore it suffered from FOUC.

### What

according to @lojjic assets can be preloaded via preloadFont, this was wrapped into a suspense call.

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### Additional comments 

i removed all notions of (deprecated) use-asset and used suspend-react. it's smaller (500bytes), and typescript can completely infer results. this will be used by r3f v8 by default but it doesn't hurt to already use it here.


JoeLuciano:
This now causes the component to demount if you ever change the text being passed in. I've had to copy this component and remove the 'suspense' line to prevent my Canvas from flashing when I update the children prop

PR #782, opened 2022-03-01
fix: keep overrideMaterial while using ContactShadows

iuriiiurevich:
scene.overrideMaterial is being cleared in ContactShadows's render cycle.
Is it intentional?

### DIFF
--- a/src/core/ContactShadows.tsx
+++ b/src/core/ContactShadows.tsx
@@ -101,10 +101,11 @@ export const ContactShadows = React.forwardRef(
       if (shadowCamera.current && (frames === Infinity || count < frames)) {
         const initialBackground = scene.background
         scene.background = null
+        const initialOverrideMaterial = scene.overrideMaterial
         scene.overrideMaterial = depthMaterial
         gl.setRenderTarget(renderTarget)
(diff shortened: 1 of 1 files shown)


gsimone (approved):
With this, any override that's present at render time for ContactShadows will be restored after rendering on the shadows target, LGTM but if you could provide a reproduction on codesandbox it would be extremely beneficial for future reference :D

iuriiiurevich:
> With this, any override that's present at render time for ContactShadows will be restored after rendering on the shadows target, LGTM but if you could provide a reproduction on codesandbox it would be extremely beneficial for future reference :D

iuriiiurevich:
> With this, any override that's present at render time for ContactShadows will be restored after rendering on the shadows target, LGTM but if you could provide a reproduction on codesandbox it would be extremely beneficial for future reference :D

If I got you right, here is a sandbox demonstrating this issue: https://codesandbox.io/s/basic-demo-forked-tpb6cx

drcmda:
it's a bug, or an oversight. it really clears out the override which wasn't my intent. with your change it has no side-effects.

Issue #789, opened 2022-03-07
Support more than one class on the Stats component


discosultan:
### Describe the feature you'd like:

The `Stats` component has a `className` property for styling the component. Unlike all the React DOM built-in components, it only allows a single class to be passed. The following raises an exception:

```jsx
<Stats className="class-a class-b" />
```

It would be intuitive if the above scenario was allowed, correctly attaching the two classes to the underlying DOM element.

### Suggested implementation:

I would change the [current implementation on master](https://github.com/pmndrs/drei/blob/1e9fac88071a68f9b7d88bda3e30e1e8cb44be1e/src/core/Stats.tsx#L19) from:

```ts
if (className) stats.dom.classList.add(className)
```

to:

```ts
if (className) stats.dom.classList.add(...className.split(" ").filter((cls) => cls))
```

The `.filter((cls) => cls))` gets rid of empty strings in the array.

drcmda:
makes sense, would you make the pr?

discosultan:
@drcmda Will open a PR soon. How can I test or validate that the change doesn't break anything?

PR #790, opened 2022-03-08
feat: support multiple css classes on Stats component

discosultan:
### Why

The `Stats` component has a `className` property for styling the component. Unlike all the React DOM built-in components, it only allowed a single class to be passed. The following raised an exception:

```jsx
<Stats className="class-a class-b" />
```

### What

The `Stats` component has been reworked to split the passed `className` by whitespace and individual classes added to the underlying DOM element.

### Checklist

- [x] Ready to be merged

Tested manually through the Stats storybook.

Closes https://github.com/pmndrs/drei/issues/789.

### DIFF
--- a/src/core/Stats.tsx
+++ b/src/core/Stats.tsx
@@ -16,7 +16,7 @@ export function Stats({ showPanel = 0, className, parent }: Props): null {
       const node = (parent && parent.current) || document.body
       stats.showPanel(showPanel)
       node?.appendChild(stats.dom)
-      if (className) stats.dom.classList.add(className)
+      if (className) stats.dom.classList.add(...className.split(' ').filter((cls) => cls))
       const begin = addEffect(() => stats.begin())
       const end = addAfterEffect(() => stats.end())
       return () => {


src/core/Stats.tsx
```diff
@@ -16,7 +16,7 @@ export function Stats({ showPanel = 0, className, parent }: Props): null {
       const node = (parent && parent.current) || document.body
       stats.showPanel(showPanel)
       node?.appendChild(stats.dom)
-      if (className) stats.dom.classList.add(className)
+      if (className) stats.dom.classList.add(...className.split(' ').filter((cls) => cls))
```
discosultan, line 19: The `.filter((cls) => cls)` part is useful to remove empty strings from the array.

For example:
```js
const className = ' foo bar';
const classListWithEmptyStrings = className.split(' ');            // [' ', 'foo', 'bar']
const classList = classListWithEmptyStrings.filter((cls) => cls); // ['foo', 'bar']
```

gsimone (approved):

joshuaellis (approved):

PR #796, opened 2022-03-10
feat: useMeasure for ScrollHtml

nazarenooviedo:
### Why

I needed to be able to resize the browser and keep the HTML height dynamically

### What

I used a hook `useMeasure` to get the height, and I replaced the static page number with the bounds got by the hook.

### Checklist
- [X] Ready to be merged

I'm open to receiving other approaches, comments,  to polish it!


drcmda:
what i do not understand is what usemeasure is supposed to do here, useThree size.width/height will tell us the actual pixel size of the canvas and that already has been measured through use-measure ... (?)

PR #798, opened 2022-03-12
fix: WebGLMultisampleRenderTarget removed in Three r138

### LINKED ISSUE #797: WebGLMultisampleRenderTarget has been removed in Three r128
- `three` version: 0.138.3
- `@react-three/fiber` version: 7.0.26
- `@react-three/drei` version: 8.16.5
- `node` version: 17.3
- `npm` (or `yarn`) version: 8.3.0

### Problem description:

The latest version of Three has removed WebGLMultisampleRenderTarget.[ From the migration guide:](https://github.com/mrdoob/three.js/wiki/Migration-Guide#137--138)
```
WebGLMultisampleRenderTarget has been removed. To use multisampling as before, use WebGLRenderTarget and set the new samples property to a value greater 0.
```
As a result, console errors pop up.

### Suggested solution:

Replace WebGLMultisampleRenderTarget wherever possible and update the types.

disambiguator:
### Why

Resolves #797.

### What

Update Three and types to r138, replacing usages of WebGLMultisampleRenderTarget

### Checklist

- [x] Documentation updated
- [x] Ready to be merged

This is my first PR to the library to I could be missing things. I also do not understand `drei`'s strategy for backwards compatibility, because the new `samples` property will not exist in versions older than r138.


gsimone:
Any way we could avoid making this a breaking change? eg. checking MultisampleRenderTarget exists and using RenderTarget instead?

disambiguator:
@gsimone That specifically won't work because MultisampleRenderTarget still exists in r138, just with a deprecation warning. https://github.com/mrdoob/three.js/blob/c85b05789cd6d2ea011bf17f58bdc192031d9b35/src/Three.Legacy.js#L2096

I thought to use THREE.REVISION instead to check, but since that's a string, I worry that doing some sort of `if(Number(THREE.REVISION) < 138` will be kind of unstable. How has this been dealt with in the past?

drcmda:
i was unaware of this and already fixed usefbo, i noticed that effectcomposer is also completely broken atm. since fbo is already in i think this should also be merged, there's no downside, the worst thing that could happen are a few jagged edges on older three, but ultimatively i think drei should try to keep up with three.

PR #808, opened 2022-03-17
feat: Add `Trail` and `useTrail`

FarazzShaikh:
### Why
Trails are cool!

### What
Added `<Trail />` and `useTrail`

### Checklist

- [ x ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/core/Trail.tsx
+++ b/src/core/Trail.tsx
@@ -1,0 +1,103 @@
+import { useFrame, useThree } from '@react-three/fiber'
+import * as React from 'react'
+import { ColorRepresentation, Group, Mesh, Object3D, SkinnedMesh, Vector2, Vector3 } from 'three'
+
+// @ts-ignore
+import { MeshLine, MeshLineMaterial } from 'three.meshline'
+
+interface TrailProps {
+  width?: number
+  color?: ColorRepresentation
+  length?: number
+  decay?: number
+  attenuation?: (width: number) => number
+  target?: React.MutableRefObject<Object3D>
+}
+
+export function useTrail(target: React.MutableRefObject<Object3D>, length: number = 1, decay: number = 1) {
+  const points = React.useRef<Vector3[]>([])
+
+  React.useLayoutEffect(() => {
+    points.current = []
+    for (let i = 0; i < length * 1000; i++) {
+      points.current.push(new Vector3(i, i, i))
+    }
+  }, [length])
+
+  useFrame((_, dt) => {
+    if (!target.current) return
+
+    const steps = 100 * decay
+    for (let i = 0; i < steps; i++) {
+      const n = target.current.position.clone()
+
+      points.current.shift()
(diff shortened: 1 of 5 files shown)


gsimone:
Soundtrack: https://www.youtube.com/watch?v=YeZZk2czG1c

src/core/Trail.tsx
```diff
@@ -0,0 +1,125 @@
...
+  target?: React.MutableRefObject<Object3D>
+}
+
+const shiftLeft = (collection, steps = 1) => {
+  collection.set(collection.subarray(steps))
+  collection.fill(-Infinity, -steps)
+  return collection
+}
+
+export function useTrail(target: Object3D, length: number = 1, decay: number = 1, local = false) {
```
gsimone, line 27: I don't like how the hook uses the object while the component version uses a ref, the Object + state is useful in the hook since I want to re-calc the initial positions on target change ( especially for the initial state).

Suggestions & changes welcome

src/core/Trail.tsx
```diff
@@ -0,0 +1,125 @@
...
+            // @ts-ignore
+            typeof child.type === 'string' && child.type === 'meshLineMaterial'
+        )
+
+        // @ts-ignore
+        if (typeof matOverride?.props === 'object') {
+          // @ts-ignore
+          m.setValues(matOverride.props)
+        }
+      }
```
gsimone, line 104: Typing this would be cool, it feels brittle as is

.storybook/stories/Trail.stories.tsx
```diff
@@ -0,0 +1,94 @@
+import * as React from 'react'
+
+import { Setup } from '../Setup'
+
+import { Sphere, Trail, useTrail } from '../../src'
+import { useFrame } from '@react-three/fiber'
+import { InstancedMesh, Mesh, Object3D } from 'three'
+
+export default {
+  title: 'Misc/Trail',
```
gsimone, line 10: Is Misc a good place?

src/core/Trail.tsx
FarazzShaikh: There's actually little reason if any for the component to use a ref as the `target`. We can just as easily use non-ref here:

https://github.com/FarazzShaikh/drei/blob/b1d779d0df49d42e4cedc72067c740d9541f549a/src/core/Trail.tsx#L73

gsimone: We could make it both too

```
target = Ref | Object3d

and in checks

target?.current || target 
```

gsimone: But do feel free to change it back to ref, as long as it works it's no biggie

FarazzShaikh: Yep that works too, lets wait on what the other reviewers have to say

joshuaellis (approved):
damn that is some nice code.

PR #826, opened 2022-03-23
feat: adds Sampler & ComputedAttribute

gsimone:
<details>
<summary>Details</summary>
🐵   https://codesandbox.io/s/wn3cmj
🟢  https://codesandbox.io/s/r3f-fairly-realistic-grass-ehflx3
</details>

### TODO

- [x] readme

### DIFF
--- a/src/core/ComputedAttribute.tsx
+++ b/src/core/ComputedAttribute.tsx
@@ -1,3 +1,4 @@
+import { BufferAttributeProps } from '@react-three/fiber'
 import * as React from 'react'
 import { BufferAttribute, BufferGeometry } from 'three'
 
@@ -11,7 +12,11 @@ type Props = {
  * Computes the BufferAttribute by calling the `compute` function
  * and attaches the attribute to the geometry.
  */
-export const ComputedAttribute: React.FC<Props> = ({ compute, name, ...props }) => {
+export const ComputedAttribute = ({
+  compute,
+  name,
+  ...props
+}: React.PropsWithChildren<Props & BufferAttributeProps>) => {
   const [bufferAttribute] = React.useState(() => new BufferAttribute(new Float32Array(0), 1))
   const primitive = React.useRef<BufferAttribute>(null)
 
--- a/src/core/ComputedAttribute.tsx
+++ b/src/core/ComputedAttribute.tsx
@@ -17,8 +17,10 @@ export const ComputedAttribute: React.FC<Props> = ({ compute, name, ...props }) 
 
   React.useLayoutEffect(() => {
     if (primitive.current) {
-      // @ts-expect-error parent doesnt'e exist on BufferAttribute
-      const attr = compute(primitive.current.parent as BufferGeometry)
+      // @ts-expect-error brittle
+      const parent = (primitive.current.parent as BufferGeometry) ?? primitive.current.__r3f.parent
+
+      const attr = compute(parent)
       primitive.current.copy(attr)
     }
   }, [compute])
--- a/src/core/ComputedAttribute.tsx
+++ b/src/core/ComputedAttribute.tsx
@@ -1,0 +1,27 @@
+import * as React from 'react'
+import { BufferAttribute, BufferGeometry } from 'three'
+
+type Props = {
+  compute: (geometry: BufferGeometry) => BufferAttribute
+  name: string
+}
+
+/**
+ * Used exclusively as a child of a BufferGeometry.
+ * Computes the BufferAttribute by calling the `compute` function
+ * and attaches the attribute to the geometry.
+ */
+export const ComputedAttribute: React.FC<Props> = ({ compute, name, ...props }) => {
+  const [bufferAttribute] = React.useState(() => new BufferAttribute(new Float32Array(0), 1))
+  const primitive = React.useRef<BufferAttribute>(null)
+
+  React.useLayoutEffect(() => {
+    if (primitive.current) {
+      // @ts-expect-error parent doesnt'e exist on BufferAttribute
+      const attr = compute(primitive.current.parent as BufferGeometry)
+      primitive.current.copy(attr)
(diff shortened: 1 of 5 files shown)


src/core/Sampler.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const group = React.useRef<Group>(null!)
+  const instancedRef = React.useRef<InstancedMesh>(null!)
+  const meshToSampleRef = React.useRef<Mesh>(null!)
+
+  React.useEffect(() => {
+    instancedRef.current =
+      instances?.current ?? (group.current!.children.find((c) => c.hasOwnProperty('instanceMatrix')) as InstancedMesh)
+
+    meshToSampleRef.current = mesh?.current ?? (group.current!.children.find((c) => c.type === 'Mesh') as Mesh)
+  }, [])
```
FarazzShaikh, line 81: We should add `[children,  mesh?.current, instances?.current]` as dependencies here. That will make the it re-sample if any of those change, this is needed if for example, we wanna change the base model using Leva like I do in the grass example.

src/core/Sampler.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+      instancedRef.current.setMatrixAt(i, dummy.matrix)
+    }
+
+    instancedRef.current.instanceMatrix.needsUpdate = true
+  }
+
+  React.useEffect(() => {
+    doSampling()
+  }, [])
```
FarazzShaikh, line 131: Same here, re: dependencies

src/core/Sampler.tsx
```diff
@@ -0,0 +1,138 @@
...
+   * It should mutate `transformPayload.dummy`.
+   *
+   * @see ( @todo add link to example )
+   *
+   * There is no need to update the dummy's matrix
+   */
+  transform?: TransformFn
+}
+
+export const Sampler: React.FC<Props> = ({ children, weight, transform, instances, mesh, ...props }) => {
```
FarazzShaikh, line 71: Is `React.FC` okay to use? I had read that it has a shot list of cons. Just personal preference, I'd change it to be
```ts
export const Sampler = ({ children, weight, transform, instances, mesh, ...props }: React.PropsWithChildren<Props & GroupProps>)
```

src/core/Sampler.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const meshToSampleRef = React.useRef<Mesh>(null!)
+
+  React.useEffect(() => {
+    instancedRef.current =
+      instances?.current ?? (group.current!.children.find((c) => c.hasOwnProperty('instanceMatrix')) as InstancedMesh)
+
+    meshToSampleRef.current = mesh?.current ?? (group.current!.children.find((c) => c.type === 'Mesh') as Mesh)
+  }, [])
+
+  const doSampling = () => {
```
FarazzShaikh, line 83: Nitpick: Put this in `useCallback` or move this outside the component?

src/core/ComputedAttribute.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * and attaches the attribute to the geometry.
+ */
+export const ComputedAttribute: React.FC<Props> = ({ compute, name, ...props }) => {
+  const [bufferAttribute] = React.useState(() => new BufferAttribute(new Float32Array(0), 1))
+  const primitive = React.useRef<BufferAttribute>(null)
+
+  React.useLayoutEffect(() => {
+    if (primitive.current) {
+      // @ts-expect-error parent doesnt'e exist on BufferAttribute
+      const attr = compute(primitive.current.parent as BufferGeometry)
```
FarazzShaikh, line 21: Did you mean `primitive.current.__r3f.parent`?

joshuaellis:
Seems to be an error in the codesandbox, not sure if this is because of an issue in the implementation of how the component should be used or the helper function?

image

src/core/ComputedAttribute.tsx
gsimone: I found that doesn't consistently work, for some reason @drcmda

src/core/Sampler.tsx
gsimone: Never ever had a problem with FC, do you have a link?

src/core/ComputedAttribute.tsx
FarazzShaikh: Ah well i though this might be what's breaking the CSB

src/core/Sampler.tsx
FarazzShaikh: https://stackoverflow.com/a/61547010

src/core/ComputedAttribute.tsx
drcmda: @gsimone @FarazzShaikh it should be __r3f.parent but it can well be that this is only fully consistent in v8. in v8 this will break, it won't write parent onto non scene objects, only __r3f.parent. but it's ok to merge if it works now.

FarazzShaikh:
Pasting in latest `ComputedAttribute` from this PR gives me the same error as the sandbox was getting

src/core/ComputedAttribute.tsx
gsimone: It doesn't work consinstently with any of the two, but I think I can be cheeky and have both methods, just taking the one that's defined 😅

gsimone: Until we move to v8, that is

gsimone:
> Pasting in latest `ComputedAttribute` from this PR into the grass demo gives me the same error as the sandbox was getting
> 
> 

Then it's surely the parent conundrum, I'm going to try a fix soon

gsimone:
@FarazzShaikh applied all feedback
@joshuaellis sandbox should work now

FarazzShaikh (approved):

PR #831, opened 2022-03-24
fix: changes import of MeshSurfaceSampler to `three-stdlib`

FarazzShaikh:
### What

Change import of MeshSurfaceSampler to `three-stdlib`

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


agcty:
still not working and throwing: "Cannot use import statement outside a module" 

should actually be just: 

`import { MeshSurfaceSampler } from 'three-stdlib'`
instead of 
`import { MeshSurfaceSampler } from 'three-stdlib/math/MeshSurfaceSampler'`

PR here: #833

PR #846, opened 2022-04-02
feat: view draft

drcmda:
#### View

<p>
  <a href="https://codesandbox.io/s/bp6tmc">Demo</a>
  <a href="https://codesandbox.io/s/1wmlew">Demo</a>
</p>

Views use gl.scissor to cut the viewport into segments. You tie a view to a tracking div which then controls the position and bounds of the viewport. This allows you to have multiple views with a single, performant canvas. These views will follow their tracking elements, scroll along, resize, etc.

It is advisable to re-connect the event system to a parent that contains both the canvas and the html content. This ensures that both are accessible/selectable and even allows you to mount controls or other deeper integrations into your view.

```tsx
<View
  /** The tracking element, the view will be cut according to its whereabouts */
  track: React.MutableRefObject<HTMLElement>
  /** Views take over the render loop, optional render index (1 by default) */
  index?: number
  /** If you know your view is always at the same place set this to 1 to avoid needless getBoundingClientRect overhead. The default is Infinity, which is best for css animations */
  frames?: number
  /** The scene to render, if you leave this undefined it will render the default scene */
  children?: React.ReactNode
/>
```

```jsx
const main = useRef()
const div = useRef()
return (
  <main ref={main}>
    <h1>Html content here</h1>
    <div ref={div} style={{ width: 200, height: 200 }} />
    <Canvas onCreated={(state) => state.events.connect(main.current)}>
      <View track={div}>
        <mesh />
        <OrbitControls />
      </View>
```

### Checklist

- [x] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/web/View.tsx
+++ b/src/web/View.tsx
@@ -1,0 +1,116 @@
+import * as React from 'react'
+import * as THREE from 'three'
+import { createPortal, useFrame, useThree } from '@react-three/fiber'
+
+const isOrthographicCamera = (def: any): def is THREE.OrthographicCamera =>
+  def && (def as THREE.OrthographicCamera).isOrthographicCamera
+const col = new THREE.Color()
+
+export type ContainerProps = {
+  scene: THREE.Scene
+  index: number
+  children?: React.ReactNode
+  frames: number
+  rect: React.MutableRefObject<DOMRect>
+  track: React.MutableRefObject<HTMLElement>
+}
+
+export type ViewProps = {
+  /** The tracking element, the view will be cut according to its whereabouts */
+  track: React.MutableRefObject<HTMLElement>
+  /** Views take over the render loop, optional render index (1 by default) */
+  index?: number
+  /** If you know your view is always at the same place set this to 1 to avoid needless getBoundingClientRect overhead */
+  frames?: number
+  /** The scene to render, if you leave this undefined it will render the default scene */
+  children?: React.ReactNode
+}
+
+function Container({ scene, index, children, frames, rect, track }: ContainerProps) {
+  const get = useThree((state) => state.get)
+  const camera = useThree((state) => state.camera)
+  const virtualScene = useThree((state) => state.scene)
+  const setEvents = useThree((state) => state.setEvents)
+
+  let frameCount = 0
+  useFrame((state) => {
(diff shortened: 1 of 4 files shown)


src/web/View.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export type ViewProps = {
+  /** The tracking element, the view will be cut according to its whereabouts */
+  track: React.MutableRefObject<HTMLElement>
+  /** Views take over the render loop, optional render index (1 by default) */
+  index?: number
+  /** If you know your view is always at the same place set this to 1 to avoid needless getBoundingClientRect overhead */
+  frames?: number
+  /** The scene to render, if you leave this undefined it will render the default scene */
+  children?: React.ReactNode
```
FarazzShaikh, line 26: Should we not include a `scene` and `camera` prop so that multiple windows can share a scene or camera?

FarazzShaikh:
With this setup transparency of uppermost View is not respected. See: https://codesandbox.io/s/view-skissor-51ozki?file=/src/App.js

Currently: 
image

Expected:
image

Fixed by:
```js
      state.gl.autoClear = false // This
      
      state.gl.setViewport(left, positiveYUpBottom, width, height)
      state.gl.setScissor(left, positiveYUpBottom, width, height)
      state.gl.setScissorTest(true)

      if (isOffscreen) {
        state.gl.getClearColor(col)
        state.gl.setClearColor(col, state.gl.getClearAlpha())
        state.gl.clear(true, true)
        return
      }

      state.gl.clearDepth() // This
      // When children are present render the portalled scene, otherwise the default scene
      state.gl.render(children ? virtualScene : scene, camera)
```

We should fix this in fiber itself as this is a problem even with regular portals.

src/web/View.tsx
drcmda: sure, would also be possible, you can add your changes to this pr if you want. with the current code it will share the default scene without children, but that arguably doesn't seem very useful and a scene prop would be more versatile.

drcmda:
@FarazzShaikh and i agreed to make a first merge and extend it for further use later, to get some feel for how it performs.

Issue #845, opened 2022-04-02
ScrollControls and hitBox


PooLP:
- `three` version: 0.139.2
- `@react-three/fiber` version: 7.0.27
- `@react-three/drei` version: 7.25.5
- `node` version:14.16.0
- `npm` (or `yarn`) version:6.14.11

### Problem description:

If the the app have a HTML header with height, this height shifts the hitBox of WebGL item.

codeSandBox : 
https://codesandbox.io/s/horizontal-tiles-forked-i4rv1x

On this based on @drcmda sample, each item hitBox is shifted
![Animation](https://user-images.githubusercontent.com/582733/161379591-c39762d1-5b6c-4222-aa86-f7df490f739e.gif)

If i remove ScrollControl, the hitBox is ok =>
https://codesandbox.io/s/horizontal-tiles-forked-504iuk

I'm a beginner, I don't know if it's a bug.

PooLP:
Hi, thanks 🥇 
But, the lastest version didn't work with React 16 and i wanted proposed a sample for PnP spfx sample repository (https://pnp.github.io/sp-dev-fx-webparts/). The SPFx framework allow a possibility to add functionality to MS SharePoint ans MS Teams, I thought it was interesting to add possibility to use React Three Fiber and Drei (and others) to this, but work only with React 16.

I'll wait (it's MS, so I'll wait a long time :D)

coopbri:
@PooLP did you figure out a solution for this? I am encountering the same issue (`ScrollControls` pointer event offset) on:

- `react`: 18.2.0
- `three`: 0.142.0
- `@react-three/fiber`: 8.0.27
- `@react-three/drei`: 9.16.0

If I discover a fix I will report it here.

AlexRAV:
Hi, I still have similar issues. Hitboxes of elements shifted down relative to their actual position by the height of header. If I'll remove header, everything is working. I can change position of header to absolute, but that's a hack, header shouldn't affect to these element. Is there any explanation for this behavior?
<img width="306" alt="image" src="https://user-images.githubusercontent.com/13298791/213445245-02a758e1-96f4-4749-a2bd-32e605015650.png">

AlexRAV:
I don't have scroll controls by the way, I only have a scene with Html component with transform prop.

PR #847, opened 2022-04-02
Bugfix: ScrollControls events

matiasperz:
### Why

I found out that ScrollControls was broken on the Y axes. Based on this example running the last version of `drei`: https://codesandbox.io/s/horizontal-tiles-forked-w3ux99

**The issue:** The tiles hover is being triggered even if you are not hovering them on the Y axes. And it was also triggering hover on the wrong coordinates on the drei's production storybook.

https://user-images.githubusercontent.com/43894343/161403840-aca9bec5-b48c-4b58-be93-46ec3c645090.mov

### What

I updated:
- The client coordinates axes variable on the offsetY calculation to be `clientY` instead of `clientX`
- The target element on both `x` and `y` offset calculations.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "@babel/preset-typescript": "^7.10.4",
     "@commitlint/cli": "^12.0.1",
     "@commitlint/config-conventional": "^12.0.1",
-    "@react-three/fiber": "^8.0.2",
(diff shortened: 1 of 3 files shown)


matiasperz:
Related issue #845

PR #862, opened 2022-04-17
chore: Updates three and stdlib to fix Line component

giulioz:
### Why

Right now using the Line component with the latest Three r139 yields a type error, since the `alphaWrite` prop is not present anymore.

image

### What

- `three-stdlib`, `three`, and `@types/three` are now to the latest version
- An import error was fixed in the Line stories

### Checklist

- [x] Ready to be merged


joshuaellis (approved):
LGTM

PR #864, opened 2022-04-18
fix: Type and story fixes for Segments

giulioz:
### Why

The current loading times of the Segments story are pretty slow, since it's calling `useFrame` for each object. Unfortunately this is slow in r3f, since it has to update the internal zustand state multiple times. I think a better example for performance can be shown by using an array of refs and a single useFrame.

During this process I also found that some user-facing types of the component were wrong, so I corrected them.

### What

(see above)

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/Segments.tsx
+++ b/src/core/Segments.tsx
@@ -14,19 +14,14 @@ type Api = {
   subscribe: (ref: React.RefObject<SegmentObject>) => void
 }
 
-type Segment = {
+type SegmentRef = React.RefObject<SegmentObject>
+type SegmentProps = Omit<JSX.IntrinsicElements['segmentObject'], 'start' | 'end' | 'color'> & {
   start: ReactThreeFiber.Vector3
   end: ReactThreeFiber.Vector3
   color?: ReactThreeFiber.Color
 }
 
-type SegmentRef = React.RefObject<Segment>
-type SegmentProps = Omit<JSX.IntrinsicElements['segmentObject'], 'start' | 'end' | 'color'> & Segment
-
 const context = React.createContext<Api>(null!)
-
-const arrColor = (color: Segment['color']) => (color instanceof THREE.Color ? color.toArray() : color)
-const arrPos = (pos: Segment['start']) => (pos instanceof THREE.Vector3 ? pos.toArray() : pos)
 
 const Segments = React.forwardRef<Line2, SegmentsProps>((props, forwardedRef) => {
   React.useMemo(() => extend({ SegmentObject }), [])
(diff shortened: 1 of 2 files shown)


src/core/Segments.tsx
```diff
@@ -11,24 +11,21 @@ type SegmentsProps = {
...
-type SegmentRef = React.RefObject<Segment>
-type SegmentProps = JSX.IntrinsicElements['segmentObject'] & Segment
-
 const context = React.createContext<Api>(null!)
 
-const arrColor = (color) => (color instanceof THREE.Color ? color.toArray() : color)
-const arrPos = (pos) => (pos instanceof THREE.Vector3 ? pos.toArray() : pos)
-
 const Segments = React.forwardRef<Line2, SegmentsProps>((props, forwardedRef) => {
+  React.useMemo(() => extend({ SegmentObject }), [])
```
joshuaellis, line 27: this doesn't feel right, why wouldn't we extend in the global space? cc @drcmda @CodyJasonBennett

src/core/Segments.tsx
```diff
@@ -37,12 +34,12 @@ const Segments = React.forwardRef<Line2, SegmentsProps>((props, forwardedRef) =>
   const [geometry] = React.useState(() => new LineSegmentsGeometry())
   const [resolution] = React.useState(() => new THREE.Vector2(512, 512))
 
-  const [positions] = React.useState(() => Array(limit * 6).fill(0))
-  const [colors] = React.useState(() => Array(limit * 6).fill(0))
+  const [positions] = React.useState<number[]>(() => Array(limit * 6).fill(0))
```
joshuaellis, line 37: was this not infered?

joshuaellis (changes requested):

joshuaellis:
Few Qs but nothing majro

src/core/Segments.tsx
giulioz: We do that also in the Instancing component

giulioz: https://github.com/pmndrs/drei/blob/bc6e3b0ddc5078cdd77e6385ffad0efa45dbf4d1/src/core/Instances.tsx#L39

giulioz: Furthermore, we do that for each instance

src/core/Segments.tsx
giulioz: No, because .fill doesn't change the return type of the array

joshuaellis: til

joshuaellis (approved):

PR #872, opened 2022-04-25
feat: add an abstraction component for marchingcubes

servinlp:
### Why

Was missing a component for `MarchingCubes` so I decided to make one.

### What

This PR contains a component for `MarchingCubes` from three.js. Threes demo can be found [here](https://threejs.org/examples/#webgl_marchingcubes).

In here you will find the following components `MarchingCubes`, `MarchingCube` and `MarchingPlane`. Here `MarchingCubes` is the wrapper that initializes the `MarchingCubes`. `MarchingCube` is a blob and `MarchingPlane` represents one of the walls.

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

NOTE: `three-stdlib` is being updated because in version 2.9.1 MarchingCubes hasn't been updated and causes errors.


FarazzShaikh:
That is really cool! 

Although i wonder if `MarchingCubes` is used widely enough to justify inclusion on Drei. As far as i understand it the goal of Drei is not to simply wrap `examples/jsm/` for React. it is to provide convenient abstractions for some commonly used element from `examples/jsm/` and to provide components that solve other commonlly had issues. 

I might be wrong though, what do you think @drcmda @gsimone? #963 falls in the same boat

drcmda:
i think it looks cool tbh, and useful. let's fix conflicting files and merge?

gsimone:
No strong opinion on this, but if it's not used it will probably go unmaintained and eventually be removed — THE CIRCLE OF LIFE — so whatever 🤷

drcmda:
true but without types and the format breaking out into new lines it's just 50loc, a primitive, some memo, looks pretty light to me.

PR #876, opened 2022-04-27
fix: html component breaks when toggling transform

### LINKED ISSUE #875: <Html/> component disappears if 'transform' property changes
- `three` version: 0.137.5
- `@react-three/fiber` version: 7.0.27
- `@react-three/drei` version: 8.20.2
- `node` version: CSB
- `npm` (or `yarn`) version: CSB

### Problem description:

If the `transform` property of the `<Html/>` component is initially true, changed to false, then back to true, the component ends up in a broken state where the HTML content is not visible.

### Relevant code:

[](https://codesandbox.io/s/r3f-drei-html-repro-l7t9wd?file=/src/App.js)

```js
  const [flag, setFlag] = useState(true);

  useEffect(() => {
    const intervalID = setInterval(() => {
      setFlag((value) => !value);
    }, 1000);
    return () => clearInterval(intervalID);
  }, []);

  return (
    <div className="App">
      <Canvas>
        <mesh>
          <boxBufferGeometry />
          <meshPhongMaterial />
        </mesh>
        {/* ❌  Shown once then disappears */}
        <Html
          as="div"
          visible={true}
          transform={flag}
          position={[-2, 0, 0]}
          className={"overlay"}
        >
          <h3>MarkerA</h3>
          <p>toggle:transform</p>
        </Html>
        {/* ✅  Toggles correctly */}
        <Html
          as="div"
          visible={flag}
          transform={true}
          position={[2, 0, 0]}
          className={"overlay"}
        >
          <h3>MarkerB</h3>
          <p>toggle:visible</p>
        </Html>
        <ambientLight args={[0xff0000]} intensity={0.1} />
        <directionalLight position={[0, 0, 5]} intensity={0.5} />
      </Canvas>
    </div>
  );
}

```

### Suggested solution:

I believe that changing the `transform` property "true → false → true" should be idempotent, leaving the component in its original state.

gsimone:
FIxes #875 

@drcmda I'm not sure why it was using layout effect before, this change doesn't seem to break the stories but maybe there's some stuff going on.

It was breaking because it was trying to render in a non existing root.

Notice that I put it like this because you can't re-initialize a root that's already mounted - gives a bunch of warnings.

**Demo of the fix 👉  https://drei-git-fix-html-transform-pmndrs.vercel.app/?path=/story/misc-html--html-transform-bug-st**

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -149,13 +149,19 @@ export const Html = React.forwardRef(
     const raycaster = useThree(({ raycaster }) => raycaster)
 
     const [el] = React.useState(() => document.createElement(as))
-    const root = React.useMemo(() => ReactDOM.createRoot(el), [el])
+    const root = React.useRef<ReactDOM.Root>()
     const group = React.useRef<Group>(null!)
     const oldZoom = React.useRef(0)
     const oldPosition = React.useRef([0, 0])
     const transformOuterRef = React.useRef<HTMLDivElement>(null!)
     const transformInnerRef = React.useRef<HTMLDivElement>(null!)
     const target = portal?.current ?? gl.domElement.parentNode
+
+    React.useEffect(() => {
+      root.current = ReactDOM.createRoot(el)
+
+      return () => root.current?.unmount()
+    })
 
     React.useEffect(() => {
       if (group.current) {
@@ -170,14 +176,14 @@ export const Html = React.forwardRef(
           if (prepend) target.prepend(el)
           else target.appendChild(el)
         }
+
         return () => {
           if (target) target.removeChild(el)
-          root.unmount()
         }
       }
     }, [target, transform])
 
-    React.useLayoutEffect(() => {
+    React.useEffect(() => {
       if (wrapperClass) el.className = wrapperClass
     }, [wrapperClass])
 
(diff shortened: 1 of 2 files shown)


gsimone:
I think this might be a problem in situations without frameloop, probably better to add a dedicated story for that

src/web/Html.tsx
```diff
@@ -149,14 +149,20 @@ export const Html = React.forwardRef(
...
-    const root = React.useMemo(() => ReactDOM.createRoot(el), [el])
+    const root = React.useRef<ReactDOM.Root>()
     const group = React.useRef<Group>(null!)
     const oldZoom = React.useRef(0)
     const oldPosition = React.useRef([0, 0])
     const transformOuterRef = React.useRef<HTMLDivElement>(null!)
     const transformInnerRef = React.useRef<HTMLDivElement>(null!)
     const target = portal?.current ?? gl.domElement.parentNode
 
+    React.useEffect(() => {
```
drcmda, line 160: wont this create a new root every render?

gsimone: Yeah, do we have a way to check if an element is already a reactDom root?
Otherwise maybe a simple check on the ref would work - assuming unmount also sets it back to undefined

```
if (!root.current) {
   ...
}
```

gsimone: The problem before was it was losing the root when re-rendering after a transform prop change.
So another thing could be manually disposing of the root only when it's no longer necessary, so still keeping the code in a useEffect but only unmounting if the root.current is defined AND transform is now false

donmccurdy:
Still having issues with #875 — anything I can do to help with this PR or another solution?

Issue #869, opened 2022-04-22
Image scale property typescript error


davidkhierl:
- `three` version: 0.139.2
- `@react-three/fiber` version: 8.0.12
- `@react-three/drei` version: 9.5.2
- `node` version: 16.14.2
- `npm` (or `yarn`) version: 1.22.15

### Problem description:

Type error when  setting scale prop

`Type 'number' is not assignable to type 'never'.`

![image](https://user-images.githubusercontent.com/25083033/164684825-57cf06e0-c013-44e0-b623-d895ae93c8ab.png)

PR #879, opened 2022-04-28
fix: expose Image transparent and opacity property

### LINKED ISSUE #871: Dynamically change image url.
When changing the Image URL (<Image url={url} />) in runtime it removes everything and after image load, rerenders, which is not expected. How can I solve this problem?

davidkhierl:
### Why

Cannot apply transparent and opacity to the `Image` material.

### What

Expose the `transparent` and `opacity` property

### Checklist

- [x] Documentation updated 
- [x] Storybook entry added
- [x] Ready to be merged

Closes #871


FarazzShaikh (approved):

davidkhierl:
> Great! Let’s add a storybook and I will address #871

done updating the docs and storybook for the transparency 🙌

gsimone (changes requested):

README.md
```diff
@@ -435,6 +435,8 @@ function Foo() {
     ref.current.material.zoom = ... // 1 and higher
     ref.current.material.grayscale = ... // between 0 and 1
     ref.current.material.color.set(...) // mix-in color
+    ref.current.material.transparent = ... // boolean, set this to true when applying opacity
+    ref.current.material.opacity = ... // between 0 and 1
   })
   return <Image ref={ref} url="/file.jpg" />
```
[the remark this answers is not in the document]
davidkhierl: Updated the example

FarazzShaikh:
@gsimone changing the URL during runtime seems to work fine. What exactly does 871 require? It it a new `texture` prop?

https://user-images.githubusercontent.com/55190601/165841682-3164989d-f07f-45e4-a7db-d6f5d33b6aa3.mp4

FarazzShaikh:
I assume thats what was required. Added `texture` prop to `<Image />` Now can be used:
```jsx
<Image texture={} /> // ✔
<Image url="" /> // ✔

<Image  /> // ❌
<Image texture={} url="" /> // ❌ This should also give a TS error however it wasnt doing so in my VSC, can you test?
```

src/core/Image.tsx
```diff
@@ -92,12 +106,62 @@ export const Image = React.forwardRef(
...
         {children}
       </mesh>
     )
   }
 )
+
+export const Image = React.forwardRef<THREE.Mesh, ImageProps>((props, ref) => {
+  if (props.url) return <ImageWithUrl {...props} ref={ref} />
+  else if (props.texture) return <ImageWithTexture {...props} ref={ref} />
+  else throw new Error('<Image /> requires a url or texture')
```
FarazzShaikh, line 166: Is this anti-pattern? Throwing an error in a component like that

davidkhierl:
@FarazzShaikh why is typescript complaining about this 
image
https://github.com/pmndrs/drei/issues/869

FarazzShaikh:
Probably because `scale` is also a property on `JSX.IntrinsicElements['mesh']`. I tried to fix it but ran out of energy. Will give it another go in the AM

davidkhierl:
I think instead of having two components `ImageWithUrl` `ImageWithTexture` it is much easier to maintain if we keep one component for the image, I tried this one and looks like have the similar behavior. let me know what your thoughts

image

image

FarazzShaikh:
The main thing is that if the user used a Texture we don’t want to re-trigger a suspense hit

Besides in your code if the user uses a texture, `useTexture` would be called twice. Once when the user created the texture and once within the Image component.

davidkhierl:
Oh okay understood! 😊

src/core/Image.tsx
drcmda: not that i know of. it will end up in an error boundary.

davidkhierl:
hmm what about having `ImageBase` as a child of both `ImageWithUrl` and `ImageWithTexture` would this work no?

```jsx
const ImageBase = React.forwardRef(
  (
    {
      children,
      color,
      segments = 1,
      scale = 1,
      zoom = 1,
      grayscale = 0,
      opacity = 1,
      texture,
      toneMapped,
      transparent,
      ...props
    }: Omit<ImageProps, 'url'>,
    ref: React.ForwardedRef<THREE.Mesh>
  ) => {
    extend({ ImageMaterial: ImageMaterialImpl })
    const gl = useThree((state) => state.gl)
    const planeBounds = Array.isArray(scale) ? [scale[0], scale[1]] : [scale, scale]
    const imageBounds = [texture!.image.width, texture!.image.height]
    return (
      <mesh ref={ref} scale={scale} {...props}>
        <planeGeometry args={[1, 1, segments, segments]} />
        <imageMaterial
          color={color}
          map={texture!}
          map-encoding={gl.outputEncoding}
          zoom={zoom}
          grayscale={grayscale}
          opacity={opacity}
          scale={planeBounds}
          imageBounds={imageBounds}
          toneMapped={toneMapped}
          transparent={transparent}
        />
        {children}
      </mesh>
    )
  }
)

const ImageWithUrl = React.forwardRef(({ url, ...props }: ImageProps, ref: React.ForwardedRef<THREE.Mesh>) => {
  const texture = useTexture(url!)
  return <ImageBase {...props} texture={texture} ref={ref} />
})

const ImageWithTexture = React.forwardRef((props: ImageProps, ref: React.ForwardedRef<THREE.Mesh>) => {
  return <ImageBase {...props} ref={ref} />
})
```

FarazzShaikh:
Ah yes that is a better way of doing things. Can you make the change?

davidkhierl:
is there's more we missing to be able to merge? I really wanted to start on my image fading animation 🤣

FarazzShaikh:
Awaiting approval from @gsimone

PR #882, opened 2022-04-29
Feature: Layer Component

TresAbhi:
### Why

3D projects usually require, in a lot of cases, laying objects on top of another. That's why, I made the `Layer` component that's meant to be used at the root of the `Canvas`, separating things like the GUI, Heads-up displays, grids, objects, etc.

The `Layer` component easily encapsulates this, relatively tricky feature, in a very simple API. It has one and only props (if you ignore children of course): layer. Look down below for an example.

### What

```
<Layer layer={1}>{ ... }</Layer> {/* above */}
<Layer layer={2}>{ ... }</Layer> (/* below */}
```

Check out the story in `Staging/Layer`.
Or, check out the same thing in [this sandbox](https://codesandbox.io/s/scene-based-layers-in-react-three-fiber-0eog7y).

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/.storybook/stories/Layer.stories.tsx
+++ b/.storybook/stories/Layer.stories.tsx
@@ -1,0 +1,39 @@
+import * as React from 'react'
+
+import { Setup } from '../Setup'
+
+import { Layer } from '../../src'
+
+export default {
+  title: 'Staging/Layer',
+  component: Layer,
+  decorators: [(storyFn) => <Setup>{storyFn()}</Setup>],
+}
+
+function LayerScene() {
+  return (
+    <>
+      <Layer layer={2}>
+        <mesh position={[0.75, 0.75, 0]}>
+          <boxGeometry />
+          <meshBasicMaterial color="blue" />
+        </mesh>
+      </Layer>
+      <Layer layer={1}>
+        <mesh position={[0, 0, 0]}>
+          <boxGeometry />
+          <meshBasicMaterial color="green" />
+        </mesh>
+      </Layer>
+      <Layer layer={0}>
+        <mesh position={[-0.75, -0.75, 0]}>
+          <boxGeometry />
+          <meshBasicMaterial color="red" />
+        </mesh>
+      </Layer>
+    </>
+  )
+}
+
+export const LayerSt = () => <LayerScene />
+LayerSt.storyName = 'Default'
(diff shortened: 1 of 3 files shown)


gsimone:
Layer is a meaningful name in three, so I'd change it. Also maybe this is already done che the <View. > component

TresAbhi:
I will make those changes ASAP. Sorry for the poor naming, it's going to be `<HeadsUpDisplay />` now.

TresAbhi:
Check out this [new sandbox](https://codesandbox.io/s/xenodochial-poincare-m228x4?file=/src/App.tsx) displaying a cleaner approach to the `HeadsUpDisplay` component.

One of the issues I'm having is that there is no guarantee that the end-user will have 1 `HeadsUpDisplay` with `priority` set to 1. If there are none, the default scene will never be rendered. If there are more than 1, the default scene will be rendered unnecessarily multiple times. No clue how to approach this issue.

However, I have also implemented a `RenderDefault` component, just in case the user cannot have any `HeadsUpDisplay` components with a priority of 1.

```typescript
/**
 * Renders `children` above the default scene, even if the priority is a
 * negative number.
 */
const HeadsUpDisplay: FC<HeadsUpDisplayProps> = ({ priority = 1, children }) => {
  const { gl: renderer, scene: defaultScene, camera } = useThree()
  const scene = new Scene()
  const portal = createPortal(children, scene, {
    events: { priority: priority + 1 }
  })

  useFrame(() => {
    if (priority === 1) {
      renderer.clear()
      renderer.render(defaultScene, camera)
    }

    renderer.clearDepth()
    renderer.render(scene, camera)
  }, priority)

  renderer.autoClear = false

  return <>{portal}</>
}
```


```typescript
/**
 * Use this component if the default scene disappears on usage of the
 * HeadsUpDisplay; caused because there are not HeadsUpDisplay components
 * with `1` priority.
 */
export const RenderDefault = () => {
  const { gl: renderer, scene, camera } = useThree()

  useFrame(() => renderer.render(scene, camera))

  return null
}
```

FarazzShaikh:
Link for the latest sandbox is broken, can you make sure its correct?

TresAbhi:
@FarazzShaikh https://codesandbox.io/s/0eog7y

PR #897, opened 2022-05-11
feature: adds range clamps to Float

clearly-outsane:
### Why

I wanted the ability to limit the floating of an object (on the y axis) to a specific range of positions

### What

I added the props minPosition and maxPosition to <Float />. Also added a storybook entry.

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/.storybook/stories/Float.stories.tsx
+++ b/.storybook/stories/Float.stories.tsx
@@ -1,6 +1,6 @@
 import React, { forwardRef, Suspense, useRef } from 'react'
 import * as THREE from 'three'
-import { withKnobs, number } from '@storybook/addon-knobs'
+import { withKnobs, number, array } from '@storybook/addon-knobs'
 import { useFrame } from '@react-three/fiber'
 
 import { Setup } from '../Setup'
@@ -21,11 +21,10 @@ function FloatScene() {
       <Suspense fallback={null}>
         <Float
           position={[0, 1.1, 0]}
-          minPosition={0}
-          maxPosition={number('Max Position', 4)}
+          floatingRange={[number('Min Floating Range', 0), number('Max Floating Range', 1)]}
           rotation={[Math.PI / 3.5, 0, 0]}
           rotationIntensity={number('Rotation Intensity', 4)}
-          floatIntensity={number('Float Intensity', 20)}
+          floatIntensity={number('Float Intensity', 2)}
           speed={number('Speed', 5)}
         >
           <mesh ref={cube}>
--- a/.storybook/stories/Float.stories.tsx
+++ b/.storybook/stories/Float.stories.tsx
@@ -21,7 +21,7 @@ function FloatScene() {
       <Suspense fallback={null}>
         <Float
           position={[0, 1.1, 0]}
-          floatingRange={[number('Min Floating Range', 0), number('Max Floating Range', 1)]}
+          floatingRange={[number('Min Floating Range', undefined), number('Max Floating Range', 1)]}
           rotation={[Math.PI / 3.5, 0, 0]}
           rotationIntensity={number('Rotation Intensity', 4)}
           floatIntensity={number('Float Intensity', 2)}
--- a/.storybook/stories/Float.stories.tsx
+++ b/.storybook/stories/Float.stories.tsx
@@ -1,0 +1,48 @@
+import React, { forwardRef, Suspense, useRef } from 'react'
+import * as THREE from 'three'
+import { withKnobs, number } from '@storybook/addon-knobs'
+import { useFrame } from '@react-three/fiber'
+
+import { Setup } from '../Setup'
+
+import { Float } from '../../src'
+
+export default {
+  title: 'Staging/Float',
+  component: Float,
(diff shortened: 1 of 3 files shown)


FarazzShaikh (approved):
Looks good to me. Let’s get another pair of eyes on this though. @gsimone?

FarazzShaikh:
Couple small things
- can we rename the props to `min` and `max`?
- can you quickly update the README to show the 2 new props too?

.storybook/stories/Float.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+function FloatScene() {
+  const cube = useRef()
+
+  return (
+    <>
+      <Suspense fallback={null}>
+        <Float
+          position={[0, 1.1, 0]}
+          minPosition={0}
+          maxPosition={number('Max Position', 4)}
```
gsimone, line 25: Maybe an array for `floatingRange={[ min, max ]}`? How does this work with intensity?

FarazzShaikh: the array sounds good `range={[min, max]}`.

I believe intensity acts just like it did, as a multiplier.

FarazzShaikh: Perhaps we can apply intensity after re-mapping? that would be more intuitive

clearly-outsane: Okay let me do that today! :3

clearly-outsane: I applied intensity in the re-mapping because the range basically means it won't float beyond those values. Like [2,-2] means the yPosition of that object would never go beyond 2 or -2.

FarazzShaikh: Nice! Did you push the changes? Don’t see them yet

clearly-outsane: Also the reason I made it two separate props is because I wanted them to be optional. Like you can only pass the minPosition as a prop and leave the maxPosition as default. I don't know how to do that if it was an array like floatingRange={[ min, max ]}

clearly-outsane: I applied floatIntensity after the mapping because ur right, its more intuitive as a multiplier

FarazzShaikh: With arrays you can do this:
```ts
THREE.MathUtils.mapLinear(
    yPosition, 
    -0.1, 
    0.1, 
    floatingRange?.[0] ?? -Infinity, 
    floatingRange?.[1] ?? Infinity
)
```

clearly-outsane: oh :0 let me do that then

clearly-outsane: so if they only want to set floatingRange[1] how do they pass that in as a prop ?

FarazzShaikh: They can do `floatingRange={[null, -1]}`

clearly-outsane: Yeah thought so, I've done that in the storybook example! Let me know if there's anything else I need to add :D

src/core/Float.tsx
```diff
@@ -1,24 +1,31 @@
...
 import { useFrame } from '@react-three/fiber'
 import mergeRefs from 'react-merge-refs'
+import * as THREE from 'three'
 
 export type FloatProps = JSX.IntrinsicElements['group'] & {
   speed?: number
   rotationIntensity?: number
   floatIntensity?: number
   children?: React.ReactNode
+  floatingRange?: [number | undefined | null, number | undefined | null]
```
FarazzShaikh, line 11: This is meesy. We can let it be `floatingRange?: [number?, number?]`

src/core/Float.tsx
```diff
@@ -1,24 +1,31 @@
...
 }
 
 export const Float = React.forwardRef<THREE.Group, FloatProps>(
-  ({ children, speed = 1, rotationIntensity = 1, floatIntensity = 1, ...props }, forwardRef) => {
+  (
+    { children, speed = 1, rotationIntensity = 1, floatIntensity = 1, floatingRange = [-1, 1], ...props },
```
FarazzShaikh, line 16: Let's have the default range be `[-Infinity, Infinity]` as to not cause breaking changes

FarazzShaikh (changes requested):

src/core/Float.tsx
clearly-outsane: ohh

src/core/Float.tsx
clearly-outsane: Wait if the expression only goes from [-0.1, 0.1] - would remapping it to [-Infinity, Infinity] cause issues?

FarazzShaikh: Ah yeah forgot this is not clamp. Anyways let’s make the default `+/- 0.1 * floatRange` So the default is such that there is no re-mapping

clearly-outsane: Yea that's what I've done in the recent commit! the default remapping is [-0.1, 0.1].

FarazzShaikh (approved):

PR #899, opened 2022-05-12
feat: add component for children position offsetting based on parent's AABB

sszczep:
### Why

There was no easy way to position the Html component relative to the parent's bounding box so I've created a new component offsetting the children position based on parent's size. 

The component's name and `anchor` parameter name is a subject to change. 

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/BBoxOffset.tsx
+++ b/src/core/BBoxOffset.tsx
@@ -29,9 +29,9 @@ export const BBoxOffset = ({ anchor, ...props }: BBOffsetProps) => {
       boundingBox.getSize(boundingBoxSize)
 
       ref.current.position.set(
-        boundingBoxSize.x * anchor[0],
-        boundingBoxSize.y * anchor[1],
-        boundingBoxSize.z * anchor[2]
+        parentRef.position.x + boundingBoxSize.x * anchor[0],
+        parentRef.position.y + boundingBoxSize.y * anchor[1],
+        parentRef.position.z + boundingBoxSize.z * anchor[2]
       )
     }
   })
--- a/src/core/BBoxOffset.tsx
+++ b/src/core/BBoxOffset.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { Group, Box3, Vector3 } from 'three'
+import { Box3, Vector3, Object3D } from 'three'
 import { useFrame, GroupProps } from '@react-three/fiber'
 
 export interface BBOffsetProps extends GroupProps {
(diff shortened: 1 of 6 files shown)


FarazzShaikh:
Pretty cool! Couple things:
- The `anchorY` slider doesn't seem to work in the Storybook
- The name `BBoxOffset` seems a bit too verbose for Drei. All other component names are more Implicit. I suggest `<Offset />` or `<Anchor />` Just my opnion tho.

Also, can you add a regular mesh being used with  `BBoxOffset` to the storybook? To show that its not just useful for HTML y'know

sszczep:
As of now, nesting meshes do not work as parent's bouding box expands with every children created. I'm still looking for a workaround.

src/core/BBoxOffset.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (ref.current?.parent?.parent) {
+      setParentRef(ref.current.parent)
+      ref.current.parent.parent.add(ref.current)
+    }
+  }, [])
+
+  // Do not create objects in loop.
+  // https://docs.pmnd.rs/react-three-fiber/advanced/pitfalls#don't-re-create-objects-in-loops
+  const boundingBox = new Box3()
+  const boundingBoxSize = new Vector3()
```
FarazzShaikh, line 26: Lets move these 2 variables outside the component or make them refs. That way the objects aren't initialized on every re-render.

src/core/BBoxOffset.tsx
```diff
@@ -0,0 +1,42 @@
...
+      boundingBox.setFromObject(parentRef)
+      boundingBox.getSize(boundingBoxSize)
+
+      ref.current.position.set(
+        boundingBoxSize.x * anchor[0],
+        boundingBoxSize.y * anchor[1],
+        boundingBoxSize.z * anchor[2]
+      )
+    }
+  })
```
FarazzShaikh, line 39: Why do this every frame? I think i'd be better to do this in the above `useEffect`. I've tested and it did work just fine. That way you cna also remove the `parentRef` state.

```js
  React.useEffect(() => {
    if (ref.current?.parent?.parent) {
      const parent = ref.current.parent
      ref.current.parent.parent.add(ref.current)

      boundingBox.setFromObject(parent)
      boundingBox.getSize(boundingBoxSize)

      ref.current.position.set(
        boundingBoxSize.x * anchor[0],
        boundingBoxSize.y * anchor[1],
        boundingBoxSize.z * anchor[2]
      )
    }
  }, [])
```

src/core/BBoxOffset.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react'
+import { Box3, Vector3, Object3D } from 'three'
```
FarazzShaikh, line 2: Lets make this `import * as THREE from 'three'` as all other components use that. Just consistancy

FarazzShaikh (changes requested):

src/core/BBoxOffset.tsx
sszczep: Sure, my bad

src/core/BBoxOffset.tsx
sszczep: Not really. The parent's AABB might change due to rotation, scaling, adding other meshes, etc.

src/core/BBoxOffset.tsx
sszczep: Sure thing

src/core/BBoxOffset.tsx
FarazzShaikh: Hmmm in that case, lets make `parentRef` a ref and not a state because useFrame will still get the latest ref value on each frame and it will save a re-render.

sszczep: Fixed

FarazzShaikh:
Sweet, @drcmda @gsimone Any ideas for names for this?

FarazzShaikh:
Sure `BBAnchor` works for me. @sszczep cab you rename and mark as ready for review? Will merge

sszczep:
> Sure `BBAnchor` works for me. @sszczep can you rename and mark as ready for review? Will merge

Done. Should be fine now

FarazzShaikh (approved):

drcmda:
awesome feature, thanks!

PR #900, opened 2022-05-12
feat: Add `<Text3D />`

FarazzShaikh:
### Why

Its used quite often by me and others.

### What

3D text component via `TextGeometry` without the friction of having to deal woth ThreeJS's traps in r3f

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


src/core/Text3D.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const Text3DBase = React.forwardRef<THREE.Mesh, React.PropsWithChildren<Text3DProps & { loader: FontLoader }>>(
+  ({ font, loader, size = 1, height = 0.2, bevelThickness = 0.1, bevelSize = 0.01, children, ...props }, ref) => {
+    const _font = React.useMemo(() => loader.parse(font as FontData), [font])
+
+    return (
+      <mesh ref={ref}>
+        <renamedTextGeometry
+          args={[
+            getTextFromChildren(children),
```
drcmda, line 60: imo this will re-create textgeo from scratch every render, this three class in particular is very expensive. it should memoize the arg as best as it can ...

src/core/Text3D.tsx
```diff
@@ -0,0 +1,96 @@
+import { extend, Node } from '@react-three/fiber'
+import * as React from 'react'
+import { suspend } from 'suspend-react'
+import * as THREE from 'three'
+import { TextGeometry, FontLoader, TextGeometryParameters } from 'three-stdlib'
+
+extend({ RenamedTextGeometry: TextGeometry })
```
drcmda, line 7: could this be inlined into the component as a useMemo(() => extend(...), []) ? most drei components should be doing it already. otherwise it creates a side effect that prevents tree shaking.

drcmda (changes requested):

FarazzShaikh:
Done, memoized args

PR #906, opened 2022-05-16
Wrap function type

kkirby:
### Why

This fixes the children type to be a function, or a `ReactNode`. Instead of a function with return type of `ReactNode | ReactNode`

Here's a Playground showcasing how the original type was invalid: [Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UAUBKBA+KAjA9jgNhAIYB2UAPtnoaQNwBQokUAQgsqhlQcSWpbh5169APSioAUQBO0nNPoBjHCQDOwKEQCMALljtg0gK4QGytRqIAmPXESd4mQyYZiJAORxTZ8pSvXYuqwGxqZ+Ftg2wfZojlDOYUA)

### What

Updated the Props type

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/Mask.tsx
+++ b/src/core/Mask.tsx
@@ -20,7 +20,7 @@ type Props = Omit<JSX.IntrinsicElements['mesh'], 'children'> & {
   /** If depth  of the masks own material will leak through, default: false */
   depthWrite?: boolean
   /** children must define a geometry, a render-prop function is allowed which may override the default material */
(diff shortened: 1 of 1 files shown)


drcmda:
i think that's now fixed with the other pr but thanks!

PR #919, opened 2022-05-24
fix(useHelper): allow ref value to be null

alexandernanberg:
### Why

With React 18 types you need to pass `null` to `useRef` for it to be a "MutableRefObject", this causes some issues with the `useHelper` hook currently.

```jsx
export function DirectionalLight(
  props: JSX.IntrinsicElements['directionalLight'],
) {
  const { debug } = useContext(LightContext)
  const ref = useRef<ThreeDirectionalLight>(null)
  const cameraRef = useRef<Camera | null>(null)

  useEffect(() => {
    cameraRef.current = ref.current?.shadow.camera ?? null
  }, [])

  // first arg fails TS checks with Argument of type 'false | MutableRefObject<Camera | null>' is not assignable to parameter of type 'MutableRefObject<Object3D<Event> | undefined> | Falsy'.
  useHelper(debug && cameraRef, CameraHelper)

  return <directionalLight ref={ref} {...props} />
}
```

### What

Allow `null` value of the ref

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/useHelper.tsx
+++ b/src/core/useHelper.tsx
@@ -10,7 +10,7 @@ type Constructor = new (...args: any[]) => any
 type Rest<T> = T extends [infer _, ...infer R] ? R : never
 
 export function useHelper<T extends Constructor>(
-  object3D: React.MutableRefObject<Object3D | undefined> | Falsey | undefined,
+  object3D: React.MutableRefObject<Object3D | null | undefined> | Falsey | undefined,
   helperConstructor: T,
   ...args: Rest<ConstructorParameters<T>>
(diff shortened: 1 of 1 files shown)


gsimone (approved):

steelx:
Im still facing this issue

mattrossman:
I also still experience this issue. It appears the change from this PR was undone in https://github.com/pmndrs/drei/commit/fe51e4f8eb9d309bcb92400636985b0362993953

Issue #938, opened 2022-06-08
Allow independant usage of AxisHead from GizmoViewport


aprehatny:
### Describe the feature you'd like:
The `AxisHead` component within the `GizmoViewport` file would be nice to utilize for other `AxesHelper` components or thinks of that nature. I was able to drop the `AxisHead` directly into my `axesHelper` component and it worked like a charm.

Allow access to library's private component `AxisHead`.
### Suggested implementation:
Export the `AxisHead` component from `src/core/GizmoViewport.tsx`

PR #939, opened 2022-06-08
feat: export AxisHead component from GizmoViewport

aprehatny:
### Why

The AxisHead component within the GizmoViewport file would be nice to utilize for other AxesHelper components or things of that nature. I was able to drop the AxisHead directly into my react-three-fiber `axesHelper` component and it worked out of the box.

### What
Allows access to library's private component AxisHead.

### Checklist
Exports the AxisHead component from src/core/GizmoViewport.tsx

### Relevant issue(s):
https://github.com/pmndrs/drei/issues/938

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/GizmoViewport.tsx
+++ b/src/core/GizmoViewport.tsx
@@ -43,7 +43,7 @@ function Axis({ scale = [0.8, 0.05, 0.05], color, rotation }: AxisProps) {
   )
 }
 
-function AxisHead({
+export function AxisHead({
   onClick,
   font,
   disabled,


FarazzShaikh:
Do you have a Sandbox i can take a look at as an example of an application of this?

aprehatny:
> Do you have a Sandbox I can take a look at as an example of an application of this?

I can get you a demo if absolutely necessary, but a really quick example use case would be so you can use the same things on the https://threejs.org/docs/index.html?q=axesHelper#api/en/helpers/AxesHelper element.

The axis helper can be utilized to serve different use cases than the gizmo helper, such as the need to represent xyz axis in a given space, and needing the reference points heads. 

image

FarazzShaikh:
IMO We should wrap this functionality within a `<AxesHelper />` component rather than expose `AxisHead` on its own which wouldnt be very useful

FarazzShaikh:
Closing this one, please open a new PR if you'd like to add a `<AxesHelper />` component to Drei

PR #951, opened 2022-06-17
feat: add center as an option for viewcube

rbackman:
### Why

I wanted the option to center the view cube on the bottom center instead of in the corner. while I was at it I made it so you could center on the sides as well. So now things like right-center or center-bottom work.

### What

I added more options to the alignment field of the viewCube. 

### Checklist

- [ x] Documentation updated <- there wasnt documentation to show the aligment flags
- [ x] Storybook entry added <- put some options in the GismoHelper.stories.tsx
- [ x] Ready to be merged

### DIFF
--- a/.storybook/stories/GizmoHelper.stories.tsx
+++ b/.storybook/stories/GizmoHelper.stories.tsx
@@ -15,7 +15,7 @@ export default {
   ],
 }
 
-const alignment = ['top-left', 'top-right', 'bottom-right', 'bottom-left']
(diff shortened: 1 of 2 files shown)


FarazzShaikh:
Why not just add all center options: 
- [x] bottom-center 
- [ ] left-center
- [x] right-center
- [ ] top-center

rbackman:
Ah yeah All those should work too. I went ahead and added all the centering options to the story.

> Why not just add all center options:
> 

> Why not just add all center options:
> 
> * [x]  bottom-center
> * [ ]  left-center
> * [x]  right-center
> * [ ]  top-center

Issue #947, opened 2022-06-13
Environment Ground Projection - Has Top Mirrored plane


theproj3ct:
Hey there, when using Ground Projection and Orbit Controls
I notice there is a cloned plane on top which appears if the angle is low enough, this prevents me from using ground projection. 
it gets worse the larger the radius.

Add
```
minAzimuthAngle={-Math.PI}
maxAzimuthAngle={Math.PI}
minPolarAngle={0.75}
maxPolarAngle={1.5}
```
to OrbitControls in this demo:

https://codesandbox.io/s/0c5hv9?file=/src/App.js

See mirrored plane appears on top as you move angle down.

https://user-images.githubusercontent.com/42445493/173346274-f85dc7fe-526d-47ce-9106-168ddf28fb42.mov

Possible to remove this plane on top?

Thanks

zphr:
I can confirm this bug and submitted a PR to fix it here #957

FarazzShaikh:
Fixed In `9.14.3` via #957

theproj3ct:
@FarazzShaikh @zphr 
Thanks for the help, the issue is resolved. Many thanks

PR #957, opened 2022-06-24
fix: the ground rendering of the Environment component

zphr:
### Why

As referenced in issue #947 the ground rendering of the Environment is broken if the camera enters the radius and looks up.

### What

My solution is to use a back facing check in the disk intersection function to avoid using the wrong intersection point (the one behind the camera if it looks up).

### Checklist

- [X] Ready to be merged


FarazzShaikh:
Brilliant! Thanks!

PR #963, opened 2022-06-30
feat: add abstraction for CatmullRomCurve3

### LINKED ISSUE #456: CatmullRomCurve3 spline in Drei
Is there any chance CatmullRomCurve3 will be added to drei? It's far better than the Cubic or Quadratic curves.

haf-decent:
### Why

I was working on a project using CatmullRomCurve3 instances and I created a custom abstraction component. Since abstractions exist for CubicBezierCurve3 and QuadraticBezierCurve3, but not CatmullRomCurve3, I figured this component may be useful to others. Resolves #456 

### What

Added new abstraction for CatmullRomCurve3, `CatmullRomLine`, based on `CubicBezierLine` and `QuadraticBezierLine`

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### Other

Within the `CatmullRomLine` abstraction, I added a vertexColor interpolation calculation which automatically generates an interpolated vertexColor array dynamically based on the number of `segments` specified in props. If this interpolation is considered useful, and seeing as the `segments` approach is also used in the other Line abstractions, this calculation could be extracted into some sort of utility function or hook.

### DIFF
--- a/.storybook/stories/Line.stories.tsx
+++ b/.storybook/stories/Line.stories.tsx
@@ -1,11 +1,11 @@
 import * as React from 'react'
 import { Vector3 } from 'three'
 import { GeometryUtils } from 'three-stdlib/utils/GeometryUtils'
-import { withKnobs, number, color, boolean } from '@storybook/addon-knobs'
+import { withKnobs, number, color, boolean, select } from '@storybook/addon-knobs'
 
 import { Setup } from '../Setup'
 
-import { Line, OrbitControls, QuadraticBezierLine, CubicBezierLine } from '../../src'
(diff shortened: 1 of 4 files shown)


gsimone:
@FarazzShaikh the only difference in my mind is that - allegedly - a curve abstraction is more useful generally than marching cubes, which are a rare thing to use generally. So i'd be more than happy to merge this, for example, rather than the marching cubes one ( we can merge both, just exposing my thought process! )

FarazzShaikh:
Yeah fair enough, i agree the curve can be more useful than Marching Cubes

drcmda:
> ( we can merge both, just exposing my thought process! )

let's do, we just need to build demos around these so that people get to see the potential. perhaps at a later stage we can find way to vote features out if we suspect they're not doing well.

PR #981, opened 2022-07-12
fix: View component supports non-full screen use cases

### LINKED ISSUE #944: `<View />` Component assumes that the canvas is full screen
- `three` version: r139
- `@react-three/fiber` version: 8.x (any compatible with createPortal state "enclave")
- `@react-three/drei` version: 9.1.3 (any including View component)
- `node` version: 16
- `npm` (or `yarn`) version: 3.2.0

### Problem description:

The nifty new View component seems to bake-in the assumption that the canvas is fullscreen. If the r3f canvas is not fullscreen, the `View`s within it are offset from the top/left by the same amount that the whole canvas is.

### Relevant code:

I've modified one of the `<View />` examples to demonstrate this.

https://codesandbox.io/s/view-skissor-forked-k46ley?file=/src/App.js

### Suggested solution:

When calculating the scissor viewport in the `<View />` component, the calculations for bottom and left should factor in the canvas element's own bounding rect as well as the "track" elements bounding rect.

Not sure what the best way to go about getting the canvas's bounding rect--it's no longer exposed on the `state.size` property like it used to be in older r3f versions (now it's just height, width. Used to include `top` and `left`).

kmannislands:
### Why

Resolves #944 

### What

Since the [upstream fix to @react-three/fiber](https://github.com/pmndrs/react-three-fiber/pull/2339) has been released in v8.1.0, this takes advantage of the position information to get `View` working as expected in non-fullscreen use cases.

Additionally, a warning is logged (using console.warn) if an older version of react-three-fiber is detected. I couldn't find any existing logging infrastructure for this type of dev mode warning at a glance so I opted to keep it simple. Please advise if there is a better approach here.

### Checklist

- [X] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -1844,7 +1844,13 @@ useBVH(mesh)
 
 Views use gl.scissor to cut the viewport into segments. You tie a view to a tracking div which then controls the position and bounds of the viewport. This allows you to have multiple views with a single, performant canvas. These views will follow their tracking elements, scroll along, resize, etc.
 
-It is advisable to re-connect the event system to a parent that contains both the canvas and the html content. This ensures that both are accessible/selectable and even allows you to mount controls or other deeper integrations into your view.
+It is advisable to re-connect the event system to a parent that contains both the canvas and the html content.
+This ensures that both are accessible/selectable and even allows you to mount controls or other deeper
+integrations into your view.
+
+> Note that `@react-three/fiber` newer than `^8.1.0` is required for `View` to work correctly if the
+> canvas/react three fiber root is not fullscreen. A warning will be logged if drei is used with older
+> versions of `@react-three/fiber`.
 
 ```tsx
 <View
--- a/src/web/View.tsx
+++ b/src/web/View.tsx
@@ -47,21 +47,28 @@ export type ViewProps = {
   children?: React.ReactNode
 }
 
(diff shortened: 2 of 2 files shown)


kmannislands:
Handling seems correct with latest react-three-fiber: https://codesandbox.io/s/view-skissor-forked-cxyur1?file=/src/App.js

Confirmed that the warning message is logged with older versions but the bottom two views have gone missing: https://codesandbox.io/s/view-skissor-forked-249ngx

I'll investigate

FarazzShaikh:
The bug in the old version box is that the `isOffscreen` calculation requires the `bottom` value to be before it is converted to the `positiveYUpBottom` form

https://github.com/pmndrs/drei/blob/c7fa011b67b2837918aa6c4c009b585b07c8940d/src/web/View.tsx#L44-L46

Fixed box: https://codesandbox.io/s/view-skissor-forked-gvhfg1?file=/src/View.tsx:2788-2912

kmannislands:
Implemented a speculative fix. I think I'll add a story (`Abstractions/View` ?) for this when I have a moment to make informal testing easier.

itsthekeming:
Any update on this?

joshuaellis (approved):

PR #1001, opened 2022-08-03
fix: selection state missing when dragging transforms pointerup

m1911star:
### Why

bug: selection state loss when dragging transforms pointer up

I think it's better to keep that state

**Before**

https://user-images.githubusercontent.com/4948120/182520782-9df2385d-66f6-4cb2-bf9a-fdc56a53c48d.mov

**After**

https://user-images.githubusercontent.com/4948120/182520856-7b55ef33-de3d-4afa-818b-35350c84f4b6.mov

### What

The problem is that `onClick` triggered improperly, so I use `onPointerDown` `onPointerUp` `onPointerMove` to simulate `onClick`

### Checklist

- [x] Documentation updated (no need for that)
- [x] Storybook entry added
- [x] Ready to be merged


src/web/Select.tsx
```diff
@@ -37,12 +37,35 @@ export function Select({
     },
     []
   )
-  React.useEffect(() => void onChange?.(active), [active])
+  const [isPointerDown, setDown] = React.useState(false)
+  const [dragging, setDragging] = React.useState(false)
+  const onPointerMove = React.useCallback(() => {
+    setDragging(isPointerDown)
```
drcmda, line 43: i feel a little uneasy about setStating react 60 times a second here. that will have performance implications. is there a better way? can we use refs?

```jsx
const dragging = useRef(false)

...
dragging.current = true
```

drcmda (changes requested):

src/web/Select.tsx
m1911star: Hi, thanks for your review.

> can we use refs?

Yes, it will be better

> is there a reason for this?

Because when you drag transforms, the onClick will still be triggered. So I make it a flag to judge whether current event is a `click` or `drag`

m1911star: fixed

Issue #894, opened 2022-05-09
useGLTF returns incomplete type


krispya:
- `three` version: 0.139.2
- `@react-three/fiber` version: 8.0.12
- `@react-three/drei` version: 9.6.2
- `node` version: 14.17.3
- `npm` (or `yarn`) version: 8.9.0

useGLTF returns type `GLTF` which is defined from the [declare file ](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/three/examples/jsm/loaders/GLTFLoader.d.ts) as:

```ts
export interface GLTF {
    animations: AnimationClip[];
    scene: Group;
    scenes: Group[];
    cameras: Camera[];
    asset: {
        copyright?: string | undefined;
        generator?: string | undefined;
        version?: string | undefined;
        minVersion?: string | undefined;
        extensions?: any;
        extras?: any;
    };
    parser: GLTFParser;
    userData: any;
}
```

But it is missing the `nodes` and `materials` properties. I'm not sure if this is an issue on the DefinitelyTyped side or something we can fix on the Drei side.

system32uwu:
Updates? We can safely patch the package to overwrite this for the time being right?

system32uwu:
guess you can just use [this](https://github.com/pmndrs/gltfjsx) instead:

`npx gltfjsx model.gltf --types`

rgbjoy:
I'd love this fixed

trebeljahr:
I think there's at least one more issue related to this namely #1159 so a few people seem to run into this. 

would love to help fix it, but I don't know enough about GLTF formats to come up with good types for what nodes and materials could potentially be, because I guess it depends on the GLTF file and what it contains?

trebeljahr:
ok, looked at the closed issues too and it seems like there was a whole discussion around this already #469

trebeljahr:
specifically this comment seems to answer it nicely: 

https://github.com/pmndrs/drei/issues/469#issuecomment-882054348

PR #1002, opened 2022-08-04
fix: add proper generic for useGLTF

miker2049:
This still requires on ugly cast, not sure if its r3f problem yet.

### Why
Trying to address #894 

### What
In order for r3f's `useLoader` to return `GLTF & ObjectMap`, a second generic is needed that extends GLTF.  

But this then causes an error in the first parameter for `useLoader`, which is why there is a temporary `as unknown` typecast, and also why this is a draft.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/core/useGLTF.tsx
+++ b/src/core/useGLTF.tsx
@@ -1,7 +1,7 @@
 import { Loader } from 'three'
 // @ts-ignore
-import { GLTFLoader, DRACOLoader, MeshoptDecoder } from 'three-stdlib'
-import { useLoader } from '@react-three/fiber'
+import { GLTFLoader, DRACOLoader, MeshoptDecoder, GLTF } from 'three-stdlib'
+import { LoaderResult, useLoader } from '@react-three/fiber'
 
 let dracoLoader: DRACOLoader | null = null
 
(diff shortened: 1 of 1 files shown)


miker2049:
Not clear this if anything up here is going to change.

https://github.com/pmndrs/react-three-fiber/pull/930

miker2049:
If [this](https://github.com/pmndrs/react-three-fiber/pull/2421) goes through, I can remove the cast.

miker2049:
Necessary fix in r3f is merged in master.  Once we have that, this will be ready.

https://github.com/pmndrs/react-three-fiber/pull/2421

Annieday:
Hi, thanks for fixing the typing.
Do you have an estimation on when this can get into the release?
I need this fix for upgrading my project to react 18.

drcmda:
@CodyJasonBennett @joshuaellis could anyone confirm?

src/core/useGLTF.tsx
```diff
@@ -1,7 +1,7 @@
 import { Loader } from 'three'
 // @ts-ignore
-import { GLTFLoader, DRACOLoader, MeshoptDecoder } from 'three-stdlib'
-import { useLoader } from '@react-three/fiber'
+import { GLTFLoader, DRACOLoader, MeshoptDecoder, GLTF } from 'three-stdlib'
+import { LoaderResult, useLoader } from '@react-three/fiber'
```
joshuaellis, line 4: Where have you used LoaderResult?

joshuaellis (changes requested):

src/core/useGLTF.tsx
miker2049: Oops nowhere! Fixed now.

joshuaellis (approved):

PR #1005, opened 2022-08-09
feat: Add `useSurfaceSampler` hook

FarazzShaikh:
### Why

Adds `useSurfaceSampler` as it is sometimes useful to have sampler result as a buffer.

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -768,6 +768,7 @@ You can either pass a Mesh and InstancedMesh as children:
 <Sampler
   weight={"normal"} // the name of the attribute to be used as sampling weight
   transform={transformPoint} // a function that transforms each instance given a sample. See the examples for more.
+  count={16} // Number of samples
 >
   <mesh>
     <sphereGeometry args={[2]} />
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ The `native` route of the library **does not** export `Html` or `Loader`. The de
           <li><a href="#useintersect">useIntersect</a></li>
           <li><a href="#useboxprojectedenv">useBoxProjectedEnv</a></li>
           <li><a href="#useTrail">useTrail</a></li>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1423,6 +1425,22 @@ useFrame(() => {
...
+```js
+const buffer = useSurfaceSampler(
+  mesh, // Mesh to sample
+  count, // [Optional] Number of samples (default: 16)
+  transform, // [Optional] Transformation function. Same as in `<Sampler />`
+  weight, // [Optional] Same as in `<Sampler />`
+  instancedMesh // [Optional] Instanced mesh to scatter
+)
+```
+
```
CodyJasonBennett, line 1443: Implementation looks fine but watch out for the typo -- it would be `InstancedMesh`.

src/core/Sampler.tsx
```diff
@@ -128,12 +128,47 @@ export const Sampler = ({
...
+
+export const Sampler = ({
+  children,
+  weight,
+  transform,
+  instances,
+  mesh,
+  count = 16,
+  ...props
+}: React.PropsWithChildren<Props & GroupProps>) => {
```
CodyJasonBennett, line 158: This is something we could hint at via linter but you might want to prefer using named functions (e.g. `function Sampler(...) {}`) so they show up properly in browsers' profilers and react devtools. Arrow functions show up as anonymous.

```jsx
// Anonymous
const Sampler = (props) => {}
// Named
function Sampler(props) {}
// Named, with forward ref (hypothetical)
const Sampler = React.forwardRef(function Sampler(props, ref) {})
```

FarazzShaikh: I see, sure will change

PR #1008, opened 2022-08-18
fix(Stage): wrap the environment with suspense

dineshsalunke:
### Why

Environment in the Stage component isn't wrapped in Suspense, so changing the environment prop on it causes the 
whole stage component flash till the new environment is loaded.

Ideally it would be good if we can keep the previous environment till the new one is loaded, thoughts on this ?

### What

Wrapped the Environment component in Suspense and added fallback prop on the stage component

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

If the draft is ok then I will update the doc with the fallback prop details if required


drcmda:
keeping the previous would be possible with useTransition

drcmda:
@dineshsalunke keeping the current one would be possible if the prop is a transition.

PR #1009, opened 2022-08-18
refactor: react-merge-refs => useImperativeHandle

CodyJasonBennett:
### What

Removes `react-merge-refs` in favor of React's built-in `useImperativeHandle`.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
     "lodash.pick": "^4.4.0",
     "meshline": "^2.0.4",
     "react-composer": "^5.0.3",
-    "react-merge-refs": "^1.1.0",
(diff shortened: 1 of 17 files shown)


CodyJasonBennett:
I'm not going to be able to unravel the typing woes around Instances etc. Leaving this as a draft for now.

drcmda:
if you want just do any and let someone else figure it out, there are places where i did that bc it just blocked everything. useImperativeHandle makes so much sense here instead of react-use-refs, and rur has even caused problems in the past because their later version breaks compat, so removing that would be awesome.

PR #1015, opened 2022-08-23
add missing } to close example ScrollControls function

Kevinjchang98:
### Why

ScrollControls' example code has a missing } in the Foo() function.

### What

Add missing } to allow for more convenient copy and pasting

### Checklist

- [x] Documentation updated
- [x] Ready to be merged


drcmda:
it's not really missing :) many example are like that, it just shows you the critical code, sometimes it also leaves jsx elements open.

PR #1025, opened 2022-08-31
feat: add pivot-controls

STINGNAILS:
### Why

We needed a convenient gizmo for both translation and rotation, which doesn't necessarily transform the object directly (but still can), but may provide transform matrices like local, global, and their delta changes. And looks better for that as well.

### What

We (with @drcmda ) have implemented a gizmo which does this, with some customization (i.e. axis colors, axis length (absolute and in pixels), axis width (in pixels only). Will link an up-to-date codesandbox later.

### Checklist

- [x] Link codesandbox
- [x] Storybook
- [x] Docs
- [x] Ready to be merged


src/core/index.ts
```diff
@@ -35,6 +35,7 @@ export * from './PointerLockControls'
 export * from './FirstPersonControls'
 
 // Gizmos
+export { PivotControls } from './pivotControls'
```
joshuaellis, line 38: This should follow the same pattern as other exports please.

src/core/GizmoViewport.tsx
```diff
@@ -109,7 +109,7 @@ export const GizmoViewport = ({
   hideAxisHeads,
   disabled,
   font = '18px Inter var, Arial, sans-serif',
-  axisColors = ['#ff3653', '#0adb50', '#2c8fdf'],
+  axisColors = ['#ff2060', '#20df80', '#2080ff'],
```
joshuaellis, line 112: Whats the reason for this change?

src/core/pivotControls/AxisArrow.tsx
```diff
@@ -0,0 +1,154 @@
...
+    userData,
+  } = React.useContext(context)
+
+  const camControls = useThree((state) => state.controls as any)
+  const objRef = React.useRef<THREE.Group>(null!)
+  const clickInfo = React.useRef<{ clickPoint: THREE.Vector3; dir: THREE.Vector3 } | null>(null)
+  const [isHovered, setIsHovered] = React.useState(false)
+
+  const onPointerDown = React.useCallback(
+    (e: any) => {
```
joshuaellis, line 58: This should be correctly typed.

joshuaellis: Also why are we wrapping these in `useCallback`?

src/core/pivotControls/AxisArrow.tsx
```diff
@@ -0,0 +1,154 @@
...
+    [camControls, onDragEnd]
+  )
+
+  const onPointerOut = React.useCallback((e: any) => {
+    e.stopPropagation()
+    setIsHovered(false)
+  }, [])
+
+  const { cylinderLength, coneWidth, coneLength, matrixL } = React.useMemo(() => {
+    const coneWidth_ = fixed ? (lineWidth / scale) * 1.2 : scale / 20
```
joshuaellis, line 103: is there a reason you've appended underscores to the end of variable names? we don't do this in other files.

joshuaellis (changes requested):
I've done a quick first pass, I think you've over engineered your event handlers imo but happy to be told why you need to do it such that way. Also we should be avoiding using the type `any` we have specific types for thinks like PointerEvents, they should be used.

src/core/GizmoViewport.tsx
drcmda: i did this, the colors in gizmo were off, saturation wasn't accurate among the rgb pairs. but its almost imperceptible but good opportunity to clean up.

src/core/pivotControls/AxisArrow.tsx
drcmda: its problematic, i think fiber has wrong types. e.target is not defined by `ThreeEvent<PointerEvent>`. should we fix it here? `ThreeEvent<PointerEvent & { target: { capture..... }>`

src/core/pivotControls/AxisArrow.tsx
drcmda: will fix

src/core/pivotControls/AxisArrow.tsx
drcmda: yeah, this is more complex than i thought. if i change the event type it wont fit into the fiber element props. the correct approach imo is `ThreeEvent<PointerEvent>` and a `// @ts-ignore` until fiber is fixed upstream.

joshuaellis (approved):

PR #933, opened 2022-06-05
feat: add support for multiple active modes to TransformControls (#927)

Glavin001:
### Why

Enable multiple/all modes at the same time for `TransformControls`. See #927

### What

Demo: https://drei-git-fork-glavin001-feat-927-all-transformcontrols-pmndrs.vercel.app/?path=/story/controls-transformcontrols--transform-controls-all-modes-story

Modes:

| 🆕 All | Translate | Rotate | Scale |
| --- | --- | --- | --- |
| image | image | image | image |

### Checklist

- [x] Get 👍 from @sketchpunk it's OK to add to Drei
- [x] Merge in dependent Pull Request
  - https://github.com/sketchpunklabs/manipulator3d/pull/1
  - [x] manipulator3d is published to `npm`
- [x] Add support for other `TransformControls` features (i.e. from `transformOnlyPropNames`)
  - [x] Works by default
  - Snap
    - ~~Translation Snap~~
    - ~~Rotation Snap~~
  - ~~Show axis~~
  - [Events](https://threejs.org/docs/?q=transform#examples/en/controls/TransformControls)
    - [x] change
    - ~~mouseDown~~
    - ~~mouseUp~~
    - [x] objectChange
  - [~~Methods~~](https://threejs.org/docs/?q=transform#examples/en/controls/TransformControls)
  - [~~Properties~~](https://threejs.org/docs/?q=transform#examples/en/controls/TransformControls)
- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

For future Pull Requests:

- [ ] Add support for other `TransformControls` features (i.e. from `transformOnlyPropNames`)
  - Snap
    - [ ] Translation Snap
    - [ ] Rotation Snap
  - [ ] Show axis
  - [Events](https://threejs.org/docs/?q=transform#examples/en/controls/TransformControls)
    - [ ] mouseDown
    - [ ] mouseUp
  - [ ] [Methods](https://threejs.org/docs/?q=transform#examples/en/controls/TransformControls)
  - [ ] [Properties](https://threejs.org/docs/?q=transform#examples/en/controls/TransformControls)

PR #1040, opened 2022-09-05
feat: add hit areas to pivotcontrols

iuriiiurevich:
Added invisible meshes for raycaster picking (as in [`TransformControls`](https://github.com/mrdoob/three.js/blob/68daccedef9c9c325cc5f4c929fcaf05229aa1b3/examples/jsm/controls/TransformControls.js#L1129)).
Picking gizmos by their thin lines is not really convenient, especially for touch devices.

It can be tweaked and improved in the future, but for now that is kind of must-have feature.

Is it ok?

Related: https://github.com/pmndrs/drei/pull/933#issuecomment-1226039814


drcmda:
@iuriiiurevich do you have a demo of it?

iuriiiurevich:
@drcmda https://codesandbox.io/s/object-gizmo-controls-forked-invozu?file=/src/App.js

PR #1045, opened 2022-09-07
feat: pivotcontrols better hit areas

iuriiiurevich:
### Why

There were some problems like this one:
Trying to pick the slider, an arrow is picked instead
(Trying to pick the slider, an arrow is picked instead)

### What

Before:
Before
After:
After


iuriiiurevich:
Demo: https://codesandbox.io/s/object-gizmo-controls-forked-mjr2js

drcmda:
i fixed it, didnt see your pr. though i use cylinders with even top and bottom radius, seems alright now.

iuriiiurevich:
@drcmda 
Imo, larger hit areas are better, it is noticeable no touch devices.
For example, that's how hit areas look in Three's TransformControls:
TransformControls

drcmda:
hit areas for axes now are the same as rotators. it could also make sense at the user level to test for mobile and make linewidth thicker and the control bigger.

PR #1071, opened 2022-10-04
fix(RenderTexture): use texture JSX props

CodyJasonBennett:
Properly extends `RenderTexture` props to match a native element, namely `attach`.


CodyJasonBennett:
Upgraded Vercel CI to Node 16 so previews should work now.

PR #1075, opened 2022-10-04
feat: expose component impl, props

### LINKED ISSUE #937: Export `OrbitControlsImpl`
https://github.com/pmndrs/drei/blob/b86b46ff01344ce92e49f694ece815d29a464450/src/core/OrbitControls.tsx#L4

`OrbitControls` component requires a `ref` prop of type `OrbitControls` from `three-stdlib`.
As `three-stdlib` isn't listed as a peer dependency, consumers may not have `three-stdlib` as a direct dependency, which means doing something similar as a consumer will cause linting errors, like so

```tsx
import { OrbitControls as OrbitControlsImpl } from 'three-stdlib';
...
const orbitControlsRef = useRef<OrbitControlsImpl>(null);
...
<OrbitControls
  ref={orbitControlsRef}
   ...
/>
```
ESLint complains on the import: 

> 'three-stdlib' should be listed in the project's dependencies. Run 'npm i -S three-stdlib' to add it
eslint [import/no-extraneous-dependencies](https://github.com/import-js/eslint-plugin-import/blob/v2.26.0/docs/rules/no-extraneous-dependencies.md)

Can we forward export this type, please?

CodyJasonBennett:
Resolves #937.

### DIFF
--- a/src/core/ArcballControls.tsx
+++ b/src/core/ArcballControls.tsx
@@ -5,6 +5,8 @@ import { ArcballControls as ArcballControlsImpl } from 'three-stdlib'
 
 import type { Event, OrthographicCamera, PerspectiveCamera } from 'three'
 
+export { ArcballControlsImpl }
+
 export type ArcballControlsProps = Omit<
   ReactThreeFiber.Overwrite<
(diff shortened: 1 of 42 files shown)


drcmda:
it a good change, lots of work, if the last few conflicts can be fixed let's merge?

PR #1090, opened 2022-10-10
feat: Add SpotLightShadows

FarazzShaikh:
### Why

Shadows are a very good way to add some life and atmosphere to a scene. 

image

### What

SpotLightShadows adds a shadow-casting plane with an alpha map in front of the spotlight. It keeps the plane correctly parented and oriented. An optional `shader` prop can be used to modify/add effects to the shadow map like wind-distortion. You can also use SDFs to "manufacture" an alpha map.

If the `shader` prop is present, the shader is rendered to a render target which is then used as the Alpha Map on the shadow casting plane.

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


drcmda (approved):

drcmda:
sorry it took so long @FarazzShaikh this is awesome!

PR #1091, opened 2022-10-11
feat: add Svg component

RodrigoHamuy:
### Why

So is easy to render SVG stuff.

### What

Added an `Svg` r3f component that can render SVG data in three.

Is a wrapper around the [svg loader](https://threejs.org/examples/?q=sv#webgl_loader_svg) demo. 

[Storybook PR link](https://drei-git-fork-rodrigohamuy-feat-svg-pmndrs.vercel.app/?path=/story/abstractions-svg--svg-st)

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/Svg.tsx
+++ b/src/core/Svg.tsx
@@ -57,7 +57,7 @@ export const Svg = forwardRef<Object3D, SvgProps>(function R3FSvg(
                 </mesh>
               ))}
             {!skipStrokes &&
-              path.userData?.style.stroke &&
+              path.userData?.style.stroke !== undefined &&
               path.userData.style.stroke !== 'none' &&
               path.subPaths.map((_subPath, s) => (
                 <mesh key={s} geometry={strokeGeometries[p]![s]}>
--- a/src/core/Svg.tsx
+++ b/src/core/Svg.tsx
@@ -1,7 +1,7 @@
 import { Object3DProps, useLoader } from '@react-three/fiber'
 import * as React from 'react'
 import { forwardRef, Fragment, useEffect, useMemo } from 'react'
-import { Color, DoubleSide, Object3D } from 'three'
+import { DoubleSide, Object3D } from 'three'
 import { SVGLoader } from 'three-stdlib'
 
 export interface SvgProps extends Object3DProps {
--- a/src/core/Svg.tsx
+++ b/src/core/Svg.tsx
@@ -14,10 +14,10 @@ export interface SvgProps extends Object3DProps {
 }
 
 export const Svg = forwardRef<Object3D, SvgProps>(function R3FSvg(
-  { src: url, skipFill, skipStrokes, strokesWireframe, fillWireframe },
+  { src, skipFill, skipStrokes, strokesWireframe, fillWireframe, ...props },
   ref
 ) {
-  const svg = useLoader(SVGLoader, !url.startsWith('<svg') ? url : `data:image/svg+xml;utf8,${url}`)
+  const svg = useLoader(SVGLoader, !src.startsWith('<svg') ? src : `data:image/svg+xml;utf8,${src}`)
 
   const shapeGroups = useMemo(
     () => (skipFill ? [] : svg.paths.map((path) => SVGLoader.createShapes(path))),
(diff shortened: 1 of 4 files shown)


src/core/Svg.tsx
```diff
@@ -0,0 +1,100 @@
...
+              new MeshBasicMaterial({
+                color: new Color()
+                  .setStyle(path.userData!.style.fill === 'none' ? '#000000' : path.userData!.style.fill)
+                  .convertSRGBToLinear(),
+                opacity: path.userData!.style.fillOpacity,
+                transparent: true,
+                side: DoubleSide,
+                depthWrite: false,
+                wireframe: fillWireframe,
+              })
```
CodyJasonBennett, line 42: Maybe prefer to map this into JSX. This can leak memory when updated.

RodrigoHamuy: @CodyJasonBennett the problem with using JSX is that at the the shapes may be sharing the same material, so keeping them in JSX will mean that we will be creating more materials than needed.

`SVGLoader.createShapes( path )` can return (and it does in some of the example files) more than one shape per path/material.

As an alternative, I can add a `useEffect` return function to dispose old materials, so you may still have the best performance and we fix the memory leak.

Let me know what you think.

RodrigoHamuy: Ok, updated. All the rendering issues on some of the SVG files are now gone and... is so much nicer to read in JSX!

The only downside is that now every shape uses its own material, but I guess that's ok?

The stroke geometry I don't think is possible to convert to JSX, so instead I'm disposing it with `useEffect`.

If you are worried about the missing opportunity of reusing materials, I can apply the same strategy as with the stroke geometry and dispose it with JSX.

RodrigoHamuy: After testing all the three.js sample shapes, the following have issues, but I don't think is something to do with this PR and instead is because `three-stdlib` `SVGLoader` is out of date compared to the one from `three` `examples` folder.

- Energy: throws error `Uncaught TypeError: Cannot read properties of null (reading 'boundingSphere')
    at Frustum.intersectsObject`
- Style CSS inside def: should be yellow but is black
- Round join: the geometry seems off

CodyJasonBennett: I can update three-stdlib and you can pump it in this PR if that unblocks you here.

src/core/Svg.tsx
```diff
@@ -0,0 +1,79 @@
...
+  )
+
+  useEffect(() => {
+    return () => strokeGeometries.forEach((group) => group && group.map((g) => g.dispose()))
+  }, [strokeGeometries])
+
+  return (
+    <object3D ref={ref} {...props}>
+      <object3D scale={[1, -1, 1]}>
+        {svg.paths.map((path, p) => (
```
joshuaellis, line 41: you should not be using the `index` as the key for a map. Does the path not have a UUID?

joshuaellis (changes requested):

src/core/Svg.tsx
RodrigoHamuy: It doesn't block this PR, it will just make the SVG loader support more SVG formats. So we can merge this anyway and when `three-stdlib` is updated and `drei` bumps the latest version things should just work.

src/core/Svg.tsx
RodrigoHamuy: No, `SVGResultPaths` does not have a UUID, I tried. But in all fairness, is quite unlikely that the Svg will change in a way that could make `keys` useful in that way? 

On the other hand, I updated the Svg props to receive Material props so you can have a bit more control over them instead of exposing just the wireframe prop.

drcmda:
seems usefull, is it done?

src/core/Svg.tsx
joshuaellis: `SVGResultPaths` extends `ShapePath` which has `currentPath` could that be used as the `key`?

RodrigoHamuy: @joshuaellis `currentPath` is an object. No type definition so is hard to tell, but I did test it and I believe what it is returning is a [Path](https://threejs.org/docs/?q=path#api/en/extras/core/Path) instance. [Here is](https://github.com/mrdoob/three.js/blob/dev/src/extras/core/ShapePath.js#L21) when is initialised.

RodrigoHamuy:
> seems usefull, is it done?

@drcmda it is ready to go imho. Not sure if I should resolve the change requests or the change request authors are in charge of deciding that.

The only thing I could say to maybe add if you want is ability to pass the material type (Phong/Standard, etc) to use and other stuff like render order, but I'm not sure how to do that and keep a nice JSX style. I guess if that's wanted we can always do another PR?

src/core/Svg.tsx
joshuaellis: ah, dang. Well thanks for doing all that :)

joshuaellis (approved):

joshuaellis:
> The only thing I could say to maybe add if you want is ability to pass the material type (Phong/Standard, etc) to use but I'm not sure how to do that and keep a nice JSX style. I guess if that's wanted we can always do another PR?

lets keep it simple for the interest of tracking, upgrades can be made in other PRs. 👍🏼 great work!

PR #1096, opened 2022-10-13
feat: add enabled prop to PresentationControls

benekoehler:
### Why
really nice job with those helpers, i really like your projects!

easily disabling the controls by props wasn't possible before.
i am using a early return to disable the controls. is there a better way?
this pr removes the cursor as well as skips some calculations while dragging if enabled is set to false.

I went ahead and added a story for this component as well :)

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/web/PresentationControls.tsx
+++ b/src/web/PresentationControls.tsx
@@ -5,7 +5,7 @@ import { a, useSpring } from '@react-spring/three'
 import { useGesture } from '@use-gesture/react'
 import { SpringConfig } from '@react-spring/core'
 
-type Props = {
+export type PresentationControlProps = {
   snap?: Boolean | SpringConfig
   global?: boolean
   cursor?: boolean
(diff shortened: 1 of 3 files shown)


src/web/PresentationControls.tsx
```diff
@@ -46,14 +48,15 @@ export function PresentationControls({
   const [spring, api] = useSpring(() => ({ scale: 1, rotation: rInitial, config }))
   React.useEffect(() => void api.start({ scale: 1, rotation: rInitial, config }), [rInitial])
   React.useEffect(() => {
-    if (global && cursor) gl.domElement.style.cursor = 'grab'
-  }, [global, cursor, gl.domElement])
+    if (global && cursor && enabled) gl.domElement.style.cursor = 'grab'
```
joshuaellis, line 51: doesn't this need a cleanup in this case?

joshuaellis (changes requested):

src/web/PresentationControls.tsx
benekoehler: that is true!

drcmda:
is it ready @benekoehler ?

joshuaellis (approved):

PR #1110, opened 2022-10-25
fix: Trail position when inside transformed parent;

### LINKED ISSUE #1109: Trail's world position breaks when parent has transfroms
- `three` version:
- `@react-three/fiber` version:
- `@react-three/drei` version:
- `node` version:
- `npm` (or `yarn`) version:

### Problem description:

When trail is within a parent with trasnfrom like so:
```jsx
<group rotation={...} position={...}>
    <Trail local={false} />
</group>
```
Then the parent's position applies an offset to the Trail even with the `local` flag set to false. This leads to incorrect tracking of the anchor. 

Broken Box: https://codesandbox.io/s/gallant-surf-2hvij1?file=/src/App.js

### Suggested solution:

We nee to cancle the parent's transfroms on the trail, or better yet, we can mount the trail to the scene directly via portals.

FarazzShaikh:
### Why

 Fixes #1109

### What

Portals the Trail directly into the scene causing it to effectively ignore parent trasnforms

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/Trail.tsx
+++ b/src/core/Trail.tsx
@@ -1,4 +1,4 @@
-import { useFrame, useThree } from '@react-three/fiber'
+import { createPortal, useFrame, useThree } from '@react-three/fiber'
 import * as React from 'react'
 import { ColorRepresentation, Group, Object3D, Vector2, Vector3 } from 'three'
 // @ts-ignore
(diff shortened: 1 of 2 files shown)


src/core/Trail.tsx
```diff
@@ -159,7 +160,7 @@ export const Trail = React.forwardRef<MeshLine, React.PropsWithChildren<TrailPro
 
   return (
     <group>
-      <mesh ref={forwardRef} geometry={geo} material={mat} />
+      {createPortal(<mesh ref={forwardRef} geometry={geo} material={mat} />, scene)}
```
FarazzShaikh, line 163: @drcmda Is this a memory leak?

PR #1117, opened 2022-10-29
feat: add types to keyboard controls

RodrigoHamuy:
### Why

- `KeyboardControlsEntry` wasn't exported, so it wasn't easy to define this outside the component.
- `useKeyboardControls()` always returned `any` because TS couldn't resolve the function, which is not fun for TS devs.

### What

- Added `KeyboardControls` story: `/?path=/story/controls-keyboardcontrols--keyboard-controls-st`
- `KeyboardControlsEntry` is now exported and is optionally generic.
- `useKeyboardControls()` now returns the correct types for each case, because the function overloading declarations have been added.
- `useKeyboardControls()` now supports generics, so you can pass which keys you are expected to use.

For example, consider you have a `Controls` enum, you can now pass that as a generic and get the correct types for all case scenarios:

```tsx
enum Controls {
  forward = 'forward',
  left = 'left',
  right = 'right',
  back = 'back',
}
const [sub, get] = useKeyboardControls<Controls>()
const left = useKeyboardControls<Controls>(st => st.left);
console.log(`left: ${left}`);
useFrame(() => console.log(`back: ${get().back}`))
useEffect(() => sub(st => console.log(`forward: ${st.forward}`)), [])

get().up // will throw TS error
```

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/.storybook/stories/KeyboardControls.stories.tsx
+++ b/.storybook/stories/KeyboardControls.stories.tsx
@@ -1,0 +1,72 @@
+import { useFrame } from '@react-three/fiber'
+import * as React from 'react'
+import { useMemo, useRef } from 'react'
+import { MathUtils, Mesh, Vector3 } from 'three'
+import { Cone, KeyboardControls, KeyboardControlsEntry, useKeyboardControls } from '../../src'
+import { Setup } from '../Setup'
+
+export default {
+  title: 'Controls/KeyboardControls',
+  decorators: [
+    (storyFn) => (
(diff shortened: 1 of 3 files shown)


drcmda:
thanks! :)

RodrigoHamuy:
Wow, that was quick @drcmda , thanks!! Really sorry but spotted one more TS issue where it was returning `any` (the selector), but it should be fixed in this PR https://github.com/pmndrs/drei/pull/1118

PR #1118, opened 2022-10-29
feat: add types to keyboard controls

RodrigoHamuy:
### Why

- `KeyboardControlsEntry` wasn't exported, so it wasn't easy to define this outside the component.
- `useKeyboardControls()` always returned `any` because TS couldn't resolve the function, which is not fun for TS devs.

### What

- Added `KeyboardControls` story: `/?path=/story/controls-keyboardcontrols--keyboard-controls-st`
- `KeyboardControlsEntry` is now exported and is optionally generic.
- `useKeyboardControls()` now returns the correct types for each case, because the function overloading declarations have been added.
- `useKeyboardControls()` now supports generics, so you can pass which keys you are expected to use.

For example, consider you have a `Controls` enum, you can now pass that as a generic and get the correct types for all case scenarios:

```tsx
enum Controls {
  forward = 'forward',
  left = 'left',
  right = 'right',
  back = 'back',
}
const [sub, get] = useKeyboardControls<Controls>()
const left = useKeyboardControls<Controls>(st => st.left);
console.log(`left: ${left}`);
useFrame(() => console.log(`back: ${get().back}`))
useEffect(() => sub(st => console.log(`forward: ${st.forward}`)), [])

get().up // will throw TS error
```

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/.storybook/stories/KeyboardControls.stories.tsx
+++ b/.storybook/stories/KeyboardControls.stories.tsx
@@ -1,0 +1,72 @@
+import { useFrame } from '@react-three/fiber'
+import * as React from 'react'
+import { useMemo, useRef } from 'react'
+import { MathUtils, Mesh, Vector3 } from 'three'
+import { Cone, KeyboardControls, KeyboardControlsEntry, useKeyboardControls } from '../../src'
+import { Setup } from '../Setup'
+
+export default {
+  title: 'Controls/KeyboardControls',
+  decorators: [
+    (storyFn) => (
+      <Setup cameraPosition={new Vector3(0, 10, 0)} lights={true}>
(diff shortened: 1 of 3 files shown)


RodrigoHamuy:
About [this](https://github.com/pmndrs/drei/pull/1118/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R356), that `;` typo is being added by husky when I run `commit -m "docs: fix typo"`, so not sure how to get rid of it :smile: . I actually didn't touch anything near that section!

PR #1125, opened 2022-10-31
docs: fix KeyboardControls example type error

brentyi:
### Why

A pair of brackets was missing :slightly_smiling_face: 

### What

Fixed the type (which could also just be removed / inferred, but perhaps being explicit here is more readable?); also added a `React.` prefix to make the example consistent with the other call to `useMemo()` in the docs.

### Checklist

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,7 +471,7 @@ enum Controls {
   jump = 'jump',
 }
 function App() {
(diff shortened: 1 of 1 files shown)


drcmda:
thanks! we had it in already in a prev pr :)

PR #1126, opened 2022-11-01
fix: fixes Html element moving with scrollControls closes #1048

### LINKED ISSUE #1048: ScrollControls scroll Html element even outside of <Scroll> in Typescript
Hi everyone!

First of all, thanks for this amazing library, I have some much fun working with it.

- `three` version:
- `@react-three/fiber`: ^8.6.0
- `@react-three/drei`: ^9.23.4
- `node` version: v18.9.0
- `npm` version: 8.19.1

### Problem description:

I recently found a new (I guess) issue with the `ScrollControls`.  As explained in the title, if I put a `Html` element inside the `ScrollControls`, this `Html` element will move according to the scroll position (which is not expected because it is not inside a `Scroll` element).

For some reasons, I was able to reproduce this bug only in Typescript (it works as expected in pure JS).

### Relevant code:

Here is a minimal code to repro:

```js
import { Html, ScrollControls } from '@react-three/drei';
import { Canvas } from '@react-three/fiber';
import React from 'react';
import './App.css';

function App() {
  return (
      <Canvas style={{ position: 'fixed', top: 0, left: 0 }}>
        <ScrollControls pages={6}>
          <Html style={{ background: 'green' }}>
            <p style={{ color: 'red' }}>This is a test</p>
          </Html>
        </ScrollControls>
      </Canvas>
  );
}

export default App;
```

### Suggested solution:

I'm not sure if it is the best solution, but I at least found a temporary one:
- add a `wrapperClass` to your `Html` element with `position: fixed !important`

I also found a potential interesting hint, it seems that for the same code (example above) the component tree is not the same for Typescript and pure JS.

TomasGonzalez:
### Why

Html component is currently having issues when used along the Scroll controls, resolves #1048 

### What

I modified the `target` `DOMElement selection conditional` to prioritize the parent element returned by `gl.domElement.parentNode` instead of the element returned by `events.connected`

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged


joshuaellis:
Can you explain further detail why you've done the changes you've made please?

TomasGonzalez:
> Can you explain in further detail why you've done the changes you've made please?

When you set your `Html` element to transform, the way this works is that the element gets attached to a "target" DOM element, this target was being selected in this order `portal.current || events.connected || gl.domElement.parentNode` `portal.current` is an arbitrary value sent as a prop, so unless that was explicitly passed it would use the `events.connected` as default. The issue with that is that the DOM element returned by `events.connected` collides with `scrollControls` as it's position is modified by it. removing `events.connected` from this condition, defaults to `gl.domElement.parentNode` which is the root element of the page, this fixes the issue because now the DOM element we are using to render the `Html` component is not a child of any arbitrary Element. And I see no case where it should actually be, and even if there was a case we could just assign it manually using the `portal` prop.

drcmda:
im not sure this is a fix. imo it breaks.

if you have a scroll container you cannot use client coordinates, you need offset

```jsx
<Canvas eventPrefix="offset"
```

and it would also make sense to make the event source a shared parent, this is what would break here

```jsx
<Canvas eventSource={document.getElementById("root")}>
```

TomasGonzalez:
> 

Ok, that makes sense, I guess that In order to fix it I would have to refactor the way ScrollControls work instead? Prevent it to transform the root? and act on a child node? or I could also add a flag, here in Html component that allows the `dev` to chose whether or not to attach to the root or the event connected?

What do you suggest?

drcmda:
good question @TomasGonzalez did you figure something out?

TomasGonzalez:
> good question @TomasGonzalez did you figure something out?

I'll give some thought to this this week, please don't close the PR yet.

TomasGonzalez:
@drcmda I'm sorry but I've got no time to fully fix this issue by the method I mentioned https://github.com/pmndrs/drei/pull/1126#issuecomment-1300835495 but I found a very simple work around, As you just have to pass `{current: gl.domElement.parentNode}` as a `portal` property (obtained from the useThree hook).
I have verified that this works just fine, and already have recommended it to other devs that have asked for help, and found this opened issue. So basically I reverted my changes to the Html components and updated the readme to add that pice of information. I hope this is a good enough fix until I can dedicate some time to elegantly fixe this.

TomasGonzalez:
@drcmda @joshuaellis 👀

LinyanZ:
It seems that this work around prevents you from scrolling when your mouse is over the Html element (If you apply `fullscreen` you cannot scroll at all). Did I miss anything?

`
import { Html, ScrollControls } from "@react-three/drei";
import { Canvas, useThree } from "@react-three/fiber";

const HtmlContent = () => {
  const { gl } = useThree();
  return (
    <Html transform portal={{ current: gl.domElement.parentNode }}>
      Test
    </Html>
  );
};

function App() {
  return (
    <Canvas style={{ width: "100vw", height: "100vh" }}>
      <ScrollControls page={2}>
        <HtmlContent />
      </ScrollControls>
    </Canvas>
  );
}

export default App;
`

TomasGonzalez:
> It seems that this work around prevents you from scrolling when your mouse is over the Html element (If you apply `fullscreen` you cannot scroll at all). Did I miss anything?
> 
> ```
> import { Html, ScrollControls } from "@react-three/drei";
> import { Canvas, useThree } from "@react-three/fiber";
> 
> const HtmlContent = () => {
>   const { gl } = useThree();
>   return (
>     <Html transform portal={{ current: gl.domElement.parentNode }}>
>       Test
>     </Html>
>   );
> };
> 
> function App() {
>   return (
>     <Canvas style={{ width: "100vw", height: "100vh" }}>
>       <ScrollControls page={2}>
>         <HtmlContent />
>       </ScrollControls>
>     </Canvas>
>   );
> }
> 
> export default App;
> ```

The work around simply prioritized the Dom element that was used in versions where it used to work... I don't see how this could mess up the scroll (If it was working already)

bryndsey:
I've also been experiencing issues with the `Html` component interfering with normal scrolling within a `ScrollControl`.

My understanding about what's happening is that when using a portal to the DOM element mentioned in the workaround, events that occur within the `Html` don't propagate to the Three scene anymore. This affects not only scroll, but also pointer events that should go to any parent elements of the `Html` (e.g. if the `Html` is a child of mesh, the mesh will no longer receive click events when the `Html` is clicked).

There might be a way to correct this using the `eventSource` of the `Canvas`, but I wasn't able to figure out how to do that.

However, I did figure out another workaround that seemed to generally get the behavior I was expecting. Rather than using `gl.domElement.parentNode` (or `parentElement`), I did the following:

```
import { Html, ScrollControls, useScroll } from "@react-three/drei";
import { Canvas, useThree } from "@react-three/fiber";

const Scene = () => {
  const scrollData = useScroll();
  return (
    // Portal to a fixed-position container created by the ScrollControl component
    <Html transform portal={{ current: scrollData.fixed }}>
      Test
    </Html>
  );
};

function App() {
  return (
    <Canvas>
      <ScrollControls pages={2}>
        <Scene />
      </ScrollControls>
    </Canvas>
  );
}
```

With this, event propagation seems to work as expected again, and it handles "nested" scrolling properly (i.e. if the HTML content can scroll, you can scroll within it without moving the ScrollControl until reaching the top/bottom of the scroll, and then the ScrollControl receives the scroll again).

The only issue I've seen so far is that it seems to act very strangely on Firefox - the HTML content will jump around while scrolling. It works fine on Chrome and Safari though.

Here is a [codesandbox](https://codesandbox.io/s/happy-shockley-hmyfqn?file=/src/App.tsx) that demonstrates this.

rahulsharma1902:
` <Html transform portal={{ current: scrollData.fixed }}
                 scale={50}
                >`


`<Html
            portal={{ current: gl.domElement.parentNode }}
            scale={100}
            transformGroup
            position={ref.current?.position}
          > `
          
          
I try using both of them, but I am getting an issue. While scrolling, they are not fixed at there position. how I can resolve this issue ?

`        <ScrollControls hideScrollbar={true}
           pages={35} distance={1} damping={0.5}
           >
          <Newmodel isMobile={isMobile} />

          </ScrollControls>`

PR #1128, opened 2022-11-01
fix(Instances): use parent raycast side

AlaricBaraou:
### Why

The Instance component wasn't using the side of the parent material for its raycast.

### What

Added the side of the material before raycating.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged


drcmda:
@CodyJasonBennett looks like next deploy is down, do you know what this could be?

PR #1131, opened 2022-11-02
feat: ScreenSpace

antokhio:
### Why

New abstraction that removes camera position and rotation to children.
Gives ability to have 3d objects in the HUD of the camera.

### What
Added /src/core/ScreenSpace.tsx
Added info to readme.md
Added to src/core/index.ts
Added .storybook/stories/ScreenSpace.stories.tsx

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The `native` route of the library **does not** export `Html` or `Loader`. The de
           <li><a href="#text3d">Text3D</a></li>          
           <li><a href="#positionalaudio">PositionalAudio</a></li>
           <li><a href="#billboard">Billboard</a></li>
-          <li><a href="#screenspace">Billboard</a></li>
+          <li><a href="#screenspace">ScreenSpace</a></li>
           <li><a href="#effects">Effects</a></li>
           <li><a href="#gradienttexture">GradientTexture</a></li>
           <li><a href="#edges">Edges</a></li>
--- a/README.md
+++ b/README.md
@@ -930,9 +930,7 @@ Adds a `<group />` that always faces the camera.
 
 #### ScreenSpace
 
-<p>
-  <a href="https://codesandbox.io/s/nifty-shape-nt1h8b"><img width="20%" src="https://codesandbox.io/api/v1/sandboxes/nifty-shape-nt1h8b/screenshot.png" alt="Demo"/></a>
-</p>
+[![](https://img.shields.io/badge/-storybook-%23ff69b4)](https://drei.pmnd.rs/?path=/story/abstractions-screenspace--screen-space-story)
 
 Adds a `<group />` that aligns objects to screen space.
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ The `native` route of the library **does not** export `Html` or `Loader`. The de
           <li><a href="#text3d">Text3D</a></li>          
           <li><a href="#positionalaudio">PositionalAudio</a></li>
           <li><a href="#billboard">Billboard</a></li>
+          <li><a href="#screenspace">Billboard</a></li>
           <li><a href="#effects">Effects</a></li>
           <li><a href="#gradienttexture">GradientTexture</a></li>
           <li><a href="#edges">Edges</a></li>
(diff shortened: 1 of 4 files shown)


src/core/ScreenSpace.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Group } from 'three'
+import mergeRefs from 'react-merge-refs'
+import { useFrame } from '@react-three/fiber'
+
+export type ScreenSpaceProps = {
+  depth?: number
+} & JSX.IntrinsicElements['group']
+
+export const ScreenSpace = React.forwardRef<Group, ScreenSpaceProps>(({ children, depth = -1, ...rest }, ref) => {
+  const localRef = React.useRef<Group>(null)
```
joshuaellis, line 11: you can use `null!` because `group` will always be rendered, this avoiding you optionalChaining in `useFrame`.

joshuaellis (changes requested):

src/core/ScreenSpace.tsx
antokhio: Hi, changed to:
`const localRef = React.useRef<Group>(null!)`

src/core/ScreenSpace.tsx
```diff
@@ -0,0 +1,22 @@
...
+  useFrame(({ camera }) => {
+    localRef.current.quaternion.copy(camera.quaternion)
+    localRef.current.position.copy(camera.position)
+  })
+  return (
+    <group ref={mergeRefs([ref, localRef])} {...rest}>
+      <group position-z={-depth}>{children}</group>
+    </group>
+  )
+})
```
CodyJasonBennett, line 22: Do note that these camera transforms are in local space, so this can break if the camera is parented in some way. Alternatively, you can portal children into the camera and keep a single group for depth:

```js
import { createPortal, useThree } from '@react-three/fiber'

export const ScreenSpace = React.forwardRef<THREE.Group, ScreenSpaceProps>(function ScreenSpace(
  { children, depth = -1, ...rest },
  ref,
) {
  const camera = useThree((state) => state.camera)

  return (
    <>
      {createPortal(
        <group {...rest} ref={ref} position-z={-depth}>
          {children}
        </group>,
        camera,
      )}
    </>
  )
```

README.md
```diff
@@ -927,6 +928,20 @@ Adds a `<group />` that always faces the camera.
...
+
+Adds a `<group />` that aligns objects to screen space.
+
+```jsx
+<ScreenSpace
+  depth={1} // Distance from camera
+>
+  <Box>I'm in screen space</Box>
+</ScreenSpace>
+```
```
CodyJasonBennett, line 943: Wouldn't this be view space, rather? What is a use-case for this?

antokhio: This normally referred to screen space, since if the depth=0 you should be exactly at screen plane, the use case, if you want for instance drag an object from canvas in to the scene, you draw it in screen space while it not intersecting with anything...  You can use this for previewing of some buffers also... 

The way we have it vvvv in dx11, is a layer/group/shader modifier, that overrides camera semantics for underlaying shaders, so you can have one camera in the scene another one for let's say ui objects.... 

But i have no clue how to write to semantics in three js, so easy addition...

drcmda:
would it make sense to extend billboard for this? `<Billboard screen>` or something like that? asking because billboard does almost the same, except for the position: https://github.com/pmndrs/drei/blob/master/src/core/Billboard.tsx

also, `depth`, what does this do? is this not the same as position.z?

src/core/ScreenSpace.tsx
antokhio: Hi, i need to test both options, but can do that next week prolly 

The point of using:
```
localRef.current.quaternion.copy(camera.quaternion)
localRef.current.position.copy(camera.position)
```
Was that it supports the Html tag.
Also to take a note there is this approach witch is kind`a resonable:
https://discourse.threejs.org/t/how-to-build-a-hud-in-a-single-scene-with-a-single-camera/16108/2
```
camera.add(yourobject3d);
```

antokhio:
> would it make sense to extend billboard for this? `<Billboard screen>` or something like that? asking because billboard does almost the same, except for the position: https://github.com/pmndrs/drei/blob/master/src/core/Billboard.tsx
> 
> also, `depth`, what does this do? is this not the same as position.z?

The depth is actually -position.z in this case it’s distance from camera…

drcmda:
in this case @antokhio @CodyJasonBennett good to merge?

antokhio:
Hi, there was no particular decision in the end, because there are many other ways to achieve same functionality, currently it’s needs a decision feedback `if this type of exposed functionality would be good for library` and then run few tests on other approaches… 

Sorry it’s prolly my bad since I promised to test them… In the end prolly merge is to early or  merge and fix after.

p.s. will run other options tests this week and let you know

antokhio:
@drcmda @CodyJasonBennett just came across few issues on stackoverflow where this tech would be applicable, (screen space stencil), I think should merge, improve after by request.

PR #1134, opened 2022-11-04
fix: put PivotControls annotation class prop in PivotContext

timmevandermeer:
`PivotControls` accept a `annotationClass` prop, but the component does not actually place this into the `PivotContext`. This results in styles for the `PivotControls` label / annotation not being applied. This PR fixes that.

### Why
- Custom styles for label don't work, while the prop is documented.

### What
- Actually used the prop in the `PivotControls` component by adding it to the context that is created for the control (it was already present in the type definition).

### Checklist
- [x] Ready to be merged


drcmda:
thanks @timmevandermeer !

PR #1145, opened 2022-11-13
fix(Text): downstream tree-shaking & *BufferGeometry deprecation

CodyJasonBennett:
Fixes withstanding issues with troika-three-text and tree-shaking as well as recent three deprecations of `PlaneBufferGeometry` => `PlaneGeometry`. I've had to vendor the package since this is a breaking change there.


drcmda:
could we ask troika for an update rather? if its breaking they could release a major. afraid that once we move vendors here keeping this up to date will become an obstacle.

CodyJasonBennett:
I've been looking into fixing this in troika-three-text itself, it would bump three peer deps to r125+. There are IIFEs all over the codebase, that would have to be fixed separately for tree-shaking.

CodyJasonBennett:
https://github.com/protectwise/troika/pull/225, https://github.com/protectwise/troika/pull/224.

drcmda:
@CodyJasonBennett peer deps to r125+ imo is no problem, that's ancient.

PR #1155, opened 2022-11-21
feat: add useTrailTexture + story

ffdead:
### Why

New hook that can be used to generate a texture with a pointer trail. Useful for controlling interactive displacement etc in shaders.

### What

https://user-images.githubusercontent.com/420472/203173825-b83bdef4-9191-4199-9986-51ff48826254.mp4

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/.storybook/stories/useTrailTexture.stories.tsx
+++ b/.storybook/stories/useTrailTexture.stories.tsx
@@ -1,0 +1,41 @@
+import * as React from 'react'
+import { withKnobs, number } from '@storybook/addon-knobs'
+
+import { Setup } from '../Setup'
+
+import { useTrailTexture } from '../../src'
+
+export default {
+  title: 'misc/useTrailTexture',
+  component: useTrailTexture,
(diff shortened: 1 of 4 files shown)


drcmda:
thanks @ffdead, super useful! i would merge now and add make a few change to it, i think for instance, that the useMemo triggers every time due to config being a new object every render. maybe also making the size a little bit bigger by default?

ffdead:
Cool thanks! Good catch reg. the `useMemo` 👌 

No strong opinion on the size. It can possibly get a bit heavy to render if the maxAge is large, and it seems a smaller texture is fine in many cases. Depends on the use-case.

drcmda:
i also made props a partial and spread them inside the class, this should allow us to only give it a subset. https://github.com/pmndrs/drei/commit/5fc8a5e617c5b93bc26b1cacb7980bc66aec9082

ffdead:
Woops noticed the documentation ended up under `Loaders` in the Readme 🙊

PR #1168, opened 2022-12-05
feat: Correct occlusion on `<Html />`

### LINKED ISSUE #1129: Real occlusion with HTML elements
I noticed when using the HTML component the occlude option is 0 or 1, means if any object is in front of the element it disappears fully, rather than getting masked. 

https://codesandbox.io/s/mixing-html-and-webgl-w-occlusion-forked-nngn7o

I know it is possible to have masked occlusion with two simultaneous renderers, CSSRenderer with alpha for the HTML and the WebGLRenderer for 3D meshes; like this example: 
https://codepen.io/trusktr/pen/RjzKJx (not fiber)

FarazzShaikh:
### Why

Fixed #1129 

### What

This PR adds correct occlusion to `<Html />`. This method uses the fact that materials with `0` alpha make a "hole" in the canvas if alpha is enabled on it. This means that if we parent a plane to the `<Html />` with `alpha: 0`, and make the HTML be underneeth the canvas, we can achive correct occlusion of the HTML. 

A limitation of this method is that HTML with transparancy or things that arent rectangular like div with a border radius, will not be occluded with those extra properties taken into account. So text is not well suited to this type of occluson

### API
The occlusion technique in this PR will be refered to as "blending" and the legacy impl as "raycast"

```js
<Html /> // No occlusion

<Html occlude /> // Occlusion with "blending"
<Html occlude="blending" /> // Occlusion with "blending"

<Html occlude="raycast" /> // Occlusion with "raycasting"
<Html occlude={[ref, ref, ...]} /> // Occlusion with "raycasting"
```

This technique supports both `transform` and no `transform`

```js
<Html occlude transform /> // Works
<Html occlude transform={false} /> // Works 
```

A custom material can be provided for the backing plane to enable lighting/shadows 

```js
<Html 
    occlude 
    material={
         <meshPhysicalMaterial
            blending={NoBlending} // This option is needed in order for it to work on materials except ShaderMaterial
            opacity={0.1} // The opacity value, the more this is, the more of the backing plane (and thus lighitng) is visiblle
            ...
          />
    }
/> 
```

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


FarazzShaikh:
Just need to update readme

gsimone:
Why not make this into another component since it requires so many changes to html? Html is getting bigger and bigger, I'm afraid it's going to get out of control soon 😆 

I'd make it something like

```jsx
const RegularHtml => HtmlWeHaveNow
const NewHtml => TheNewOne

const Html = ({ specialVersion }) => {

   if (specialVersion) {
		return <RegularHtml />
   }

	return <NewHtml />
}
```

gsimone:
The correct way would be separating the logic for positioning for reusability, but feel free to merge if you don't think it's worth it 👍

FarazzShaikh:
Merging this, will refactor HTML shortly to make things more managable

PR #1213, opened 2023-01-02
Fix/1209-meshreflector

xinaesthete:
### Why
Using a GUI to tweak values of `MeshReflectorMaterial` was causing it to crash, and it appears resources would also be leaked as it went out of scope etc.

### What
Changed to `useState()` for the FBOs, `depthTexture` and `blurpass`, and rather than making new instances when properties change, mutate those in place.

Added some property getters and setters to `BlurPass` to facilitate that.

Added `dispose()` to `BlurPass` and `MeshReflectorMaterial`, haven't done much in the way of testing to verify they behave correctly.

Also haven't done very extensive testing of things like changing the resolution dynamically, but it certainly helps when tweaking sliders etc.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged

Probably could be merged as is. May want to consider using `useFBO` since it's there, but I was procrastinating about how to treat `depthTexture` in that case. I suppose the advantage would be that as well as being a bit more idiomatic, it'd mean `dispose()` was handled somewhat more naturally.

Maybe some of the code that's in the `useMemo()` could be in a `useEffect()` instead, can't really see how it'd make much difference. I'm fairly inexperienced with React.

### DIFF
--- a/src/core/MeshReflectorMaterial.tsx
+++ b/src/core/MeshReflectorMaterial.tsx
@@ -86,7 +86,28 @@ export const MeshReflectorMaterial = React.forwardRef<MeshReflectorMaterialImpl,
     const [q] = React.useState(() => new Vector4())
     const [textureMatrix] = React.useState(() => new Matrix4())
     const [virtualCamera] = React.useState(() => new PerspectiveCamera())
-
+    const parameters = {
+      minFilter: LinearFilter,
+      magFilter: LinearFilter,
+      encoding: gl.outputEncoding,
+      type: HalfFloatType,
+    }
+    //incremental change before useFBO()?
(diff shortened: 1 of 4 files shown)


drcmda:
could you use useFBO in that? yes it was made before it existed

xinaesthete:
Yes, actually I ended up duplicating the code in the repository where I'm actually using this, and making some other tweaks to the shader (so that the interaction of shadows with the reflection behaves differently), as well as using useFBO. I'll move the useFBO bit over. The only reason I didn't do that straight away was to do with thinking it might also make sense to change useFBO depth buffer management or something, but I didn't want to make too many changes.

PR #1229, opened 2023-01-12
Fix GradientTexture not working when document is undefined

merwaaan:
### Why

GradientTexture generates an error when used in a web worker because document is not defined.

### What

Switched to OffscreenCanvas so that it works in both contexts.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [X] Ready to be merged


drcmda:
OffscreenCanvas isn't in safari though (?), that would exclude practically half of mobile.

merwaaan:
@drcmda We could do like threejs does: check for the availability of OffscreenCanvas and use it if possible.

https://github.com/mrdoob/three.js/blob/88e8954b69377dad4ad1ceaf1a383f3536e88e5a/src/renderers/webgl/WebGLTextures.js#L29

PR #1231, opened 2023-01-16
feat(Line): support for LineSegments2

promontis:
### Why

Line currently only renders a ThreeJS Line2. There is no support for LineSegments2.

### What

Adds a segments prop to allow switching between ThreeJS Line2 and LineSegments2

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged


drcmda (approved):

martini-airsquire:
The newly added segments prop (boolean) is conflicting with CatmullRomLine's own segments prop (number), making it unable to set CatmullRomLine's segments, so it will always be default segments - 20, suggest rename either one to solve this bug

PR #1237, opened 2023-01-21
feat(Controls): adding CameraControls

abernier:
### Why

A [CameraControls](https://github.com/yomotsu/camera-controls) implementation component for drei.

See also this comment: https://github.com/yomotsu/camera-controls/issues/262#issuecomment-1399305886

### What

A drei `<CameraControls />` heavily inspired by the already present `<OrbitControls />`

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


package.json
```diff
@@ -132,11 +133,15 @@
...
     "@react-three/fiber": ">=8.0",
+    "camera-controls": ">=1.0",
     "react": ">=18.0",
     "react-dom": ">=18.0",
     "three": ">=0.137"
   },
   "peerDependenciesMeta": {
+    "camera-controls": {
+      "optional": true
+    },
```
CodyJasonBennett, line 144: This should be a dependency and nothing else. Do you anticipate forking the library?

src/core/CameraControls.tsx
```diff
@@ -0,0 +1,48 @@
...
+import type { PerspectiveCamera, OrthographicCamera } from 'three'
+
+import * as React from 'react'
+import { forwardRef, useMemo, useEffect } from 'react'
+import { extend, useFrame, useThree, ReactThreeFiber, EventManager } from '@react-three/fiber'
+
+import CameraControlsImpl from 'camera-controls'
+
+CameraControlsImpl.install({ THREE })
+extend({ CameraControlsImpl })
```
CodyJasonBennett, line 11: These break tree-shaking, so I'd try to move them into the component as a render-effect like we do in R3F's canvas: https://github.com/pmndrs/react-three-fiber/blob/b06ff7478f53166e5cf65402a43347f6b32d453c/packages/fiber/src/web/Canvas.tsx#L52-L55

This may be a problem elsewhere in Drei, so apologies if you're just following convention.

package.json
abernier: I defined `camera-controls` as an optional peer-dep so a user that don't need camera-controls, do not get it installed when `npm i @react-three/drei` -- according to https://github.com/npm/rfcs/issues/221

abernier: disclaimer, I'm not sure about that

CodyJasonBennett: This would be tree-shaken out though. peer dependencies are for dependencies you anticipate to be installed in user-space, and that would make it no longer an implementation detail of this component.

src/core/CameraControls.tsx
abernier: >  I'd try to move them into the component as a render-effect like we do in R3F's canvas: https://github.com/pmndrs/react-three-fiber/blob/b06ff7478f53166e5cf65402a43347f6b32d453c/packages/fiber/src/web/Canvas.tsx#L52-L55
apart from that (I'm not sure to understand the fix you suggest)

Would defining a `subsetOfTHREE` object like explained here https://github.com/yomotsu/camera-controls#important

```ts
const subsetOfTHREE = {
	Vector2   : Vector2,
	Vector3   : Vector3,
	Vector4   : Vector4,
	Quaternion: Quaternion,
	Matrix4   : Matrix4,
	Spherical : Spherical,
	Box3      : Box3,
	Sphere    : Sphere,
	Raycaster : Raycaster,
	MathUtils : {
		DEG2RAD: MathUtils.DEG2RAD,
		clamp: MathUtils.clamp,
	},
};

CameraControls.install( { THREE: subsetOfTHREE } );
```

could help?

I can easily do this, if you confirm ;)

package.json
abernier: ok I will make `camera-controls` a simple dep

src/core/CameraControls.tsx
CodyJasonBennett: It wouldn't solve the tree-shaking issue -- that has to be moved out of global/module scope, but it'll pull in a lot less of three.js when the component is imported. If the author doesn't anticipate this subset changing between versions, it would be safe to do, yeah.

abernier: better with your snippet now, I will make this change

package.json
abernier: ok done

src/core/CameraControls.tsx
abernier: ok done https://github.com/pmndrs/drei/pull/1237/files#diff-a0d895145d4d77f878604f353413ea046de8817965da0f2dad997985d06fe842R22-R25

package.json
```diff
@@ -57,11 +57,13 @@
     "@babel/runtime": "^7.11.2",
     "@react-spring/three": "^9.3.1",
     "@use-gesture/react": "^10.2.0",
+    "camera-controls": "^1.37.6",
     "detect-gpu": "^5.0.5",
     "glsl-noise": "^0.0.0",
     "lodash.clamp": "^4.0.3",
     "lodash.omit": "^4.5.0",
     "lodash.pick": "^4.4.0",
+    "maath": "^0.5.1",
```
abernier, line 66: I don't know why yarn has re-ordered the "maath" dep 🤷‍♂️

I can manually re-ordered it if you want

CodyJasonBennett: No worries, it tries to alphabetically sort.

CodyJasonBennett (approved):
Tree-shaking looks good to me here and in `camera-controls`. Same with types, all good here.

PR #1238, opened 2023-01-22
fix(CameraControls): taking advantage on new released `.connect` method

abernier:
### Why / What

Now that [camera-controls](https://github.com/yomotsu/camera-controls/releases/tag/v1.38.0) lib implements `connect`/`disconnect` methods, I'm following [@drcmda recommendation](https://github.com/yomotsu/camera-controls/issues/262#issuecomment-1399431569) and update the `CameraControls` drei component to take advantage on

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/src/core/CameraControls.tsx
+++ b/src/core/CameraControls.tsx
@@ -34,14 +34,15 @@ export const CameraControls = forwardRef<CameraControlsImpl, CameraControlsProps
   const explCamera = camera || defaultCamera
   const explDomElement = (domElement || events.connected || gl.domElement) as HTMLElement
 
-  const cameraControls = useMemo(() => new CameraControlsImpl(explCamera, explDomElement), [explCamera, explDomElement])
+  const cameraControls = useMemo(() => new CameraControlsImpl(explCamera), [explCamera])
 
   useFrame((state, delta) => {
     if (cameraControls.enabled) cameraControls.update(delta)
   }, -1)
 
   useEffect(() => {
-    return () => void cameraControls.dispose()
+    cameraControls.connect(explDomElement)
+    return () => void cameraControls.disconnect()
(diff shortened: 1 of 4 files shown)


src/core/CameraControls.tsx
```diff
@@ -41,7 +41,8 @@ export const CameraControls = forwardRef<CameraControlsImpl, CameraControlsProps
   }, -1)
 
   useEffect(() => {
-    return () => void cameraControls.dispose()
+    cameraControls.connect(explDomElement)
+    return () => void cameraControls.disconnect()
```
yomotsu, line 45: @abernier I think we could keep `dispose()` at cleanup as same as other controls in drei. `disconnect()` may remain some features of CameraControls instance.
What do you think?
```suggestion
    return () => void cameraControls.dispose()
```

abernier: in my understanding, r3f will take care of calling `.dispose` thanks to https://docs.pmnd.rs/react-three-fiber/api/objects#disposal

Here it is more about "re-connecting" (`disconnect` + `connect`) once any useEffect deps changes

@drcmda or @CodyJasonBennett can you confirm?

CodyJasonBennett: Connect/disconnect should do and undo effectful stuff like listeners. I'd expect `dispose` to not be reversible, repeatable, or cheap to do at runtime.

abernier: Thank you @CodyJasonBennett, so in my mind this looks ok to me 👍🏻

yomotsu: @abernier 
That makes sense. Thank you for considering it.

PR #1239, opened 2023-01-23
fix: `<Html />` race conditions canvas z-index

### LINKED ISSUE #1228: `<Html>` 'blending' occlude mode is not working when having multiples `<Html>` component that doesn't have the 'blending' mode
- `three` version: 0.148.0
- `@react-three/fiber` version: 8.9.1
- `@react-three/drei` version: 9.51.1
- `node` version: 16.18.1
- `npm` (or `yarn`) version: 8.19.2

### Problem description:

When having `<Html>` component with `occlude='blending'` and adding another one that doesn't have an occlude props set, it removes the blending properties from the first component. 

### Suggested solution:

One workaround is to always have `<Html>` component with `occlude='blending'` but it gives a white background. 
But I can't use this solution, I'm only using text from the HTML (and the `<Text>` component from Drei seems to not accept emojis)

Any thoughts about this? 
Thanks!

FarazzShaikh:
### Why

Fixes #1228 by making the canvas always under the `<Html />` parents.

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -193,16 +193,10 @@ export const Html = React.forwardRef(
 
     React.useLayoutEffect(() => {
       const el = gl.domElement as HTMLCanvasElement
-      if (occlude) {
-        el.style.zIndex = `${Math.floor(zIndexRange[0] / 2)}`
-        el.style.position = 'absolute'
-        el.style.pointerEvents = 'none'
-      } else {
-        el.style.zIndex = null!
-        el.style.position = null!
-        el.style.pointerEvents = null!
-      }
-    }, [occlude, isRayCastOcclusion])
+      el.style.zIndex = `${Math.floor(zIndexRange[0] / 2)}`
+      el.style.position = 'absolute'
+      el.style.pointerEvents = 'none'
+    }, [])
 
     React.useLayoutEffect(() => {
       if (group.current) {


cyrus2281:
This MR seems to broke the z-index.
I'm using the `Html` inside a `primitive` , inside a `PresentationControls`, and now I can no longer interact with the model since this verison.

```js
      <PresentationControls
        global
        rotation={[0.13, 0.1, 0]}
        polar={[-0.4, 0.2]}
        azimuth={[-1, 0.75]}
        config={{ mess: 2, tension: 400 }}
        snap={{ mess: 4, tension: 400 }}
      >
        <Float rotationIntensity={0.4}>
          <primitive object={scene} position-y={-1.2} rotation-x={0.13}>
            <Html>
                I'm broken ☹
            </Html>
          </primitive>
        </Float>
      </PresentationControls>
```

FarazzShaikh:
@cyrus2281 I cant repro, please provide a minimal CodeSanbox with the issue and i can look into it.

cyrus2281:
Here is a simple CodeSandbox.
https://codesandbox.io/s/thirsty-marco-n560lo?file=/src/Experience.js

The issue is not you can't interact with the object, your mouse cursor should change to pointer when you hover the page and you should be able to drag and move the camera perspective

For some reason the dragging doesn't work inside the sandbox, but you should still be able to see the mouse cursor change

FarazzShaikh:
You box doesnt work all the way bck to v9.54 which has nothing to do with this PR  https://codesandbox.io/s/gifted-dream-3xcq8d

What version was it working in?

FarazzShaikh:
he issue is `<PresentaitonControls />` its been broken since `9.46.5` 

Broken: https://codesandbox.io/s/bouncy-watch-forked-3l1wfp (9.46.5)
Working: https://codesandbox.io/s/bouncy-watch-forked-8oib5p (9.46.4)

firtina-lyfta:
FYI this makes the canvas pointer events none in all cases which breaks events like onclick for objects inside the canvas, even if there are small HTML objects above it. I have to fix my drei version to before this PR to fix bugs in my system.

FarazzShaikh:
@firtina-lyfta Please provide an example CodeSanbox. 

Here is it working, https://codesandbox.io/s/intelligent-kalam-n5ogxe?file=/src/App.js

The events are bound to `rootState.events.conneced` which by default is the nearest parent of the Canvas and the HTML, this measn that it doesn't matter if the canvas has no pointer events, the events are not actully bound to the canvas.

If you have a fix please pull latest master and PR the fix along side demonstrating the issue it fixes.

firtina-lyfta:
@FarazzShaikh thank you for the quick response.

I investigated a bit and I think I may have discovered an edge case.

It's when you use a createRoot and render instead of using `<Canvas/>`.

Here's a loom to demonstrate: https://www.loom.com/share/976b4e7ab5834596902e5d1326008157

And the codesandbox: https://codesandbox.io/s/nameless-water-r0xeki?file=/src/App.js

In the codesandbox, the version for drei is not the latest, you can see it working as expected. When you switch the drei version to latest, it stops working, orbit controls or the sphere click doesn't work any more, if the`<Html/>` is in the scene.

gsimone:
@FarazzShaikh This always puts some possibily irrelevant and very high zIndex on the canvas, it broke our use case since we had some UI on top with a zIndex design system that's not compatible with the randomly immense number we get ( 88k ).

PR #1240, opened 2023-01-23
fix: `GL_INVALID_OPERATION` on AccumulativeShadows

### LINKED ISSUE #1236: Accumulative Shadows: "GL_INVALID_OPERATION" console warnings
When using the <AccumulativeShadows> (and <RandomizedLight>) component, chrome spits out around 60 console warnings:

`[.WebGL-0000409A00C1A200] GL_INVALID_OPERATION: Active draw buffers with missing fragment shader outputs.`

I assume it's to do with how render targets for the shadowmap baking are used but that's just off the top of my head, and not really having a good grasp on what exactly the various components actually do under the hood ;)

This happens on Chrome/Chromium only though, not on Firefox. It's not a big deal, just figured I'd drop it here :)

Seems to be related: https://stackoverflow.com/questions/57535605/three-js-custom-shader-error-on-chrome-gl-invalid-operation-active-draw-buffers

FarazzShaikh:
### Why

Fixes #1236. The FS in DiscardMaterial just needed its output set to anything before discarding. 

### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/materials/DiscardMaterial.tsx
+++ b/src/materials/DiscardMaterial.tsx
@@ -1,3 +1,7 @@
 import { shaderMaterial } from '../core/shaderMaterial'
 
-export const DiscardMaterial = shaderMaterial({}, 'void main() { }', 'void main() { discard; }')
+export const DiscardMaterial = shaderMaterial(
+  {},
+  'void main() { }',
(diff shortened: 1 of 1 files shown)


drcmda:
oh no, i had no idea. thank you for figuring that out!

PR #1261, opened 2023-01-30
fix: set return type to null as a fn w/ return type 'void' is not a valid JSX element

rajput-hemant:
### Checklist

- [ ] Documentation updated
- [ ] Storybook entry added
- [x] Ready to be merged


drcmda:
it's strange because undefined is an allowed return type.

rajput-hemant:
@drcmda

drcmda:
that issue, TS is so weird sometimes. could you quickly remove the semicolon? then let's merge.

rajput-hemant:
@drcmda done 😇

Issue #189, opened 2020-12-04
useGLTF with Draco: TypeError: WebAssembly.instantiate() - with fix


nathansearles:
Started seeing the following errors today. After some digging it looks like it's related to [this issue](https://github.com/google/draco/issues/653) with the Draco decoders.

```
failed to asynchronously prepare wasm: TypeError: WebAssembly.instantiate(): 
Import #2 module="wasi_unstable" error: module is not an object or function
```

```
Uncaught (in promise) RuntimeError: abort(TypeError: WebAssembly.instantiate(): 
Import #2 module="wasi_unstable" error:module is not an object or function). 
Build with -s ASSERTIONS=1 for more info.
```

**Quick fix**

Thankfully we can point to other decoders. Doing the following resolved the issue.

```
const { nodes, materials } = useGLTF(
  "draco_model.glb",
  "https://www.gstatic.com/draco/versioned/decoders/1.4.0/"
);
```

Another options is to use decoder version "https://www.gstatic.com/draco/versioned/decoders/1.3.6/".

Hope this helps someone else out there tearing their hair out this evening!

deanssmart:
I have had this exact issue... my site was working the other day and then it kept just throwing up the WebAssembly.instantiate() error. Thanks for the fix it has worked for me 

` const { scene } = useLoader(GLTFLoader, "scene.gltf", draco("https://www.gstatic.com/draco/versioned/decoders/1.4.0/"));`

alex-shortt:
It's been spotty for a while but recently it seems like the cdn began working again, can anyone else confirm this?

nathansearles:
The Draco decoder URL (https://www.gstatic.com/draco/v1/decoders/) that Drei uses appears to be working again.

You can test it by going to: https://www.gstatic.com/draco/v1/decoders/draco_decoder.js

For now I'd still use:

```
const { nodes, materials } = useGLTF(
  "draco_model.glb",
  "https://www.gstatic.com/draco/versioned/decoders/1.4.0/"
);
```

drcmda:
i've made a small patch pinning the version

PR #1268, opened 2023-02-04
Update Draco decoders path

clementroche:
Use the latest Draco version
https://github.com/google/draco/releases/tag/1.5.5

### DIFF
--- a/src/core/useGLTF.tsx
+++ b/src/core/useGLTF.tsx
@@ -15,7 +15,7 @@ function extensions(useDraco: boolean | string, useMeshopt: boolean, extendLoade
         dracoLoader = new DRACOLoader()
       }
       dracoLoader.setDecoderPath(
-        typeof useDraco === 'string' ? useDraco : 'https://www.gstatic.com/draco/versioned/decoders/1.4.3/'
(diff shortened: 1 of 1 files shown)


drcmda:
did you try it, can you confirm it work? the old version was deliberate because draco had breaking changes in patch versions.

clementroche:
Can you link where the discussion happened, i was not aware ? @drcmda

drcmda:
here is one, we used draco indiscriminately without pinning back then https://github.com/pmndrs/drei/issues/189

we could also ping @donmccurdy, don, do you think 1.5.x is safe with current threejs?

donmccurdy:
I'm not aware of breaking API changes — there were some problems with the Draco CDN during release rollouts, but I think those were accidental. This is a newer version of the Draco API than three.js is using, so I haven't tested it, but would expect that if it seems to be working fine in your tests then it should be OK.

drcmda:
thanks! in that case, everything's working fine, tested it against local binaries and remote.

donmccurdy:
Will upgrade three.js in next release:

- https://github.com/mrdoob/three.js/pull/25475

We're updating DRACOLoader to use newer Draco APIs there, but previous versions of DRACOLoader work with the newest Draco library too.

PR #1293, opened 2023-02-15
Billboard: use camera world rotation

timohausmann:
### Why

In my scene the camera is nested inside an animated group. The Billboard component only takes the local camera quaternion into account, so as soon as a parent object rotation comes into play, the Billboard effect breaks.  

### What

Inset of copying the local rotation of camera.quaternion, I suggest using camera.[getWorldQuaternion()](https://threejs.org/docs/#api/en/core/Object3D.getWorldQuaternion).

### Checklist

- [x] Ready to be merged


neftaly:
I also had this issue, and came up with the same solution.
https://codesandbox.io/s/green-snow-gt7hcw

PR #1303, opened 2023-02-20
fix(MeshRefractionMaterial): minimize shader bandwidth

### LINKED ISSUE #1297: Reduce varying count in MeshRefractionMaterial

MeshRefractionMaterial currently uses 16 varying slots.  My 2019 MacBook Pro only supports 15, so the program fails to link.  At quick glance, several of the varyings are unnecessary and could be removed, which would also increase performance.

### Suggested implementation:

- projectionMatrixInv: pass in as uniform
- viewMatrixInv: pass in as uniform
- vInstanceMatrix: I'm not familiar with the reason that `instanceMatrix` is an attribute, not a uniform, in three, so this is maybe less of a question for this repo.  if it were a uniform instead, this could be removed.  otherwise, this is unnecessary if !USE_INSTANCING, so the logic could be switched around to only define + use it if there's instancing.
- vColor: similarly, this looks like it only has to be varying due to instanceColor.  but if !USE_INSTANCING_COLOR, it is unnecessary.  the varying could be hidden behind `#ifdef USE_INSTANCING_COLOR` and the multiplication by `uniform vec3 color` could be moved to the fragment shader, which would be one fewer varying if !USE_INSTANCING_COLOR

CodyJasonBennett:
Fixes #1297 by removing unneeded varyings and inlining expressions where applicable.


src/materials/MeshRefractionMaterial.tsx
```diff
@@ -21,36 +21,30 @@ export const MeshRefractionMaterial = shaderMaterial(
   #ifndef USE_COLOR
     uniform vec3 color;
   #endif
+  #ifdef USE_INSTANCING_COLOR
+    varying mat4 vInstanceMatrix;
+  #endif
+
+  varying vec3 vColor;
```
gydence, line 28: this can also be conditionally defined behind `#ifdef USE_INSTANCING_COLOR`, and perhaps renamed to `vInstanceColor`.  in the fragment shader, you can multiply by the uniform `color`, and then only multiply by `vInstanceColor` ifdef USE_INSTANCING_COLOR

src/materials/MeshRefractionMaterial.tsx
```diff
@@ -130,10 +124,15 @@ export const MeshRefractionMaterial = shaderMaterial(
...
+      mat4 modelMatrixInverse = inverse(modelMatrix * vInstanceMatrix);
+    #else
+      mat4 modelMatrixInverse = inverse(modelMatrix);
+    #endif
+
     vec2 uv = gl_FragCoord.xy / resolution;
-    vec3 directionCamPerfect = (projectionMatrixInv * vec4(uv * 2.0 - 1.0, 0.0, 1.0)).xyz;
-    directionCamPerfect = (viewMatrixInv * vec4(directionCamPerfect, 0.0)).xyz;
+    vec3 directionCamPerfect = (inverse(projectionMatrix) * vec4(uv * 2.0 - 1.0, 0.0, 1.0)).xyz;
+    directionCamPerfect = (inverse(viewMatrix) * vec4(directionCamPerfect, 0.0)).xyz;
```
gydence, line 135: `inverse` in a fragment shader should be avoided.  I believe these can be calculated in the [react component](https://github.com/pmndrs/drei/blob/master/src/core/MeshRefractionMaterial.tsx) using a beforeRender callback like [MeshReflectorMaterial](https://github.com/pmndrs/drei/blob/master/src/core/MeshReflectorMaterial.tsx).  the tradeoff is that people using the pure threejs version of the material will need to compute it themselves, but that's already true for a number of the other properties

this would be doubly helpful since we're also using inverse(viewMatrix) in the vertex shader still

src/materials/MeshRefractionMaterial.tsx
```diff
@@ -130,10 +124,15 @@ export const MeshRefractionMaterial = shaderMaterial(
   #endif
   
   void main() {
-    mat4 modelMatrixInverse = inverse(modelMatrix * vInstanceMatrix);
+    #ifdef USE_INSTANCING
+      mat4 modelMatrixInverse = inverse(modelMatrix * vInstanceMatrix);
```
gydence, line 128: I missed this before but this inverse in the fragment shader is also bad...modelMatrixInverse could be moved to the vertex shader and then you wouldn't have to pass vInstanceMatrix at all (except it would be always passing the varying instead of conditionally passing it...probably still worth it to avoid the inverse)

gydence (changes requested):
thank you again for this quick PR!  I left a few suggestions.  I had started a branch for this but I'm still getting familiar with react so I never could have done it this quickly, and with a test to boot!

src/materials/MeshRefractionMaterial.tsx
CodyJasonBennett: I'm assuming that `color` otherwise is inherited from vertex colors, so I reverted this change in 1673401a1502b3623a0305c5cbb6514e6812a6f6. If that's not so, I can add it back in.

CodyJasonBennett: Looks like [`color_vertex`](https://github.com/mrdoob/three.js/blob/master/src/renderers/shaders/ShaderChunk/color_vertex.glsl.js) isn't used, so this I'll assume it wasn't intentional for now and add back in this change.

src/materials/MeshRefractionMaterial.tsx
```diff
@@ -16,41 +16,40 @@ export const MeshRefractionMaterial = shaderMaterial(
...
     vec4 transformedNormal = vec4(normal, 0.0);
     vec4 transformedPosition = vec4(position, 1.0);
     #ifdef USE_INSTANCING
-      vInstanceMatrix = instanceMatrix;
       transformedNormal = instanceMatrix * transformedNormal;
       transformedPosition = instanceMatrix * transformedPosition;
+    #endif
+
+    #ifdef USE_INSTANCING
+      vModelMatrixInverse = inverse(modelMatrix * vInstanceMatrix);
```
gydence, line 42: for some additional savings you could pass in `inverse(modelMatrix)` as a uniform too 😄

gydence (approved):
from 16 to 6/7 varying slots and way fewer inverses!

Issue #429, opened 2021-06-03
mixer from useAnimations doesn't set _root property


yaroslavnikiforov:
- `three` version: 0.129.0
- `@react-three/fiber` version: 6.2.2
- `@react-three/drei` version: 5.3.1
- `node` version: 14.15.0
- `yarn` version: 1.22.10

### Problem description:

The `_root` property of the mixer returned by `useAnimations` hook is always undefined. This doesn't allow you to use methods of AnimationMixer, for example `mixer.clipAction()` as it relies on this property.

### Relevant code:
Just try to call `mixer.clipAction(clips[0])` in useEffect from this example:
[https://codesandbox.io/s/gltf-animations-pecl6](https://codesandbox.io/s/gltf-animations-pecl6)

joshuaellis:
Please provide a minimal reproduction of the issue.

yaroslavnikiforov:
Here is a demo:
[https://codesandbox.io/s/gltf-animations-forked-ukf1b?file=/src/Model.js](https://codesandbox.io/s/gltf-animations-forked-ukf1b?file=/src/Model.js)

JDansercoer:
@joshuaellis @yaroslavnikiforov I'm facing this issue as well. These are my versions

* three version: 0.134.0
* @react-three/fiber version: 7.0.17
* @react-three/drei version: 7.19.2

Any idea?

lexa18:
Confirm this problem. I faced the same situation.

GoreStarry:
Have the same issue.

PR #1307, opened 2023-02-22
fix(useAnimations): passing root to mixer init if it is Object3D

casualWaist:
### Why

Trying to create a new AnimationClips in useAnimations mixer fails because mixer is instantiated with no reference to the root scene as described in issue #429

### What

passed the root scene to the constructor function after and instanceof check

### Checklist

- [x] Ready to be merged

This is my first ever pull request, the fix worked for me but I'm not sure if this is breaking some intentional design or I'm missing something else. Be GeNtLe :)


drcmda:
sorry for the long delay.

i think useState is the wrong place, it should probably be a useLayoutEffect that sets the property? otherwise useState would be stale and couldn't react to changing the root from one to another?

casualWaist:
Okay, I think I see. You're saying if the root of the mixer changes via a direct call then the actualRef will be out of sync with the mixer and the useEffect cleanup won't work? I admit I didn't think to do any tests with multiple objects interacting with the mixer. Sorry, I will revisit this.

casualWaist:
I did some testing and I'm more confused. I don't understand when or why the root object would change as other objects can be passed as optional root for a new action or create a separate mixer. The class doesn't seem to have a function that changes it, even unchacheRoot didn't change it, so, I don't know. Please let me know what I should do. Should I delete the pull request?

Issue #969, opened 2022-07-06
<Mask> doesn't work with <Effects>


leeuwisarvin:
I'm currently using the <Mask> component which works great! Although if I'm adding the <Effects> component in my scene the <Mask> stops working. 

Does anyone have a solution for issue?

drcmda:
i think it works but you need to switch on the depth buffer. for instance react-postprocessing has `<EffectComposer depthBuffer ... />`.

plushviedev4:
> i think it works but you need to switch on the stencil buffer. for instance react-postprocessing has `<EffectComposer stencilBuffer ... />`, and drei/Effects has `<Effects stencilBuffer .../>`. without this it cannot use stencil on which mask is based. i know this isn't really straight forward but some of these things are highly intricate internally. if you want to add a small note to the docs please go ahead.

Thanks it's working adding stencilBuffer attribute to EffectComposer and in Bloom in my case

PR #1322, opened 2023-02-27
Add note about Masks and stencilBuffer

Loque-:
Add a quick note about Masks requiring stencilBuffer being set to true on effect composers.

### Why

I lost a bit of time not knowing stencilBuffer needed to be applied to effect composers in order to use Mask.

### What

I have just added [the note I saw in issue 969](https://github.com/pmndrs/drei/issues/969) to the readme (really appreciated the issue and feedback <3)

### Checklist

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -2864,6 +2864,12 @@ return (
     <meshStandardMaterial {...stencil} />
 ```
 
+If you are using postprocessing effects you may need to ensure the composer has the stencilBuffer attribute set.
+
+```jsx 
+<Effects stencilBuffer .../>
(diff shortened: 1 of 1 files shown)


drcmda:
could this be worded differently? by threejs-default it's off. this makes it sound like it's something needed for effect composition.

> If you are using postprocessing effects you may need to ensure the composer has the stencilBuffer attribute set.

it should be more like if postpro + stencil is wanted it needs to be set

Issue #522, opened 2021-08-22
GizmoHelper doesn't respect camera up direction


Irev-Dev:
### versions
![image](https://user-images.githubusercontent.com/29681384/130343272-4e800052-2adc-4306-a73d-05ec99493473.png)

### Problem description:

Setting the camera up position to something other than the default (e.g. `[0, 0, 1]`) works until the gizmo is used at which point it snaps back to the default up direction.

Steps to reproduce 
- go to https://codesandbox.io/s/gizmohelper-doesnt-respect-camera-axis-yr7sd?file=/src/App.js
- DONT click the viewport gizmo yet, use orbit controls and see that z axis is up ( cat should be facing up in littlest tokyo)
![image](https://user-images.githubusercontent.com/29681384/130343378-c3a730c8-d672-44a6-8742-877eb3edaed3.png)
- now click use the Gizmo, notice the up axis has gone back to the default but orbitControls as still working as if the up were still [0,0,1] 
![image](https://user-images.githubusercontent.com/29681384/130343890-8899edbe-2e24-4073-8a8e-332f91a55ad8.png)

I hope that describes the problem, its pretty obvious one you try it. Here's a video (at the end you can see I drag the mouse horizontally but the orbit doesn't match)

https://user-images.githubusercontent.com/29681384/130343560-b395f7a1-6037-4e60-aefd-f82b5f827a87.mov

I had a very quick go at debugging but am really not sure what I'm doing, I could see in https://github.com/pmndrs/drei/blob/master/src/core/GizmoHelper.tsx the line
`mainCamera.up.set(0, 1, 0).applyQuaternion(q1).normalize()`
which I thought might be related but am not sure.

Having a lot of fun using r3f :)

joshuaellis:
GizmoHelper came in as a contribution, as such we'll need the community to resolve issues like this. Thank you for the report!

JulesBlm:
This bug seems to have been introduced in pull request #496 ([commit](https://github.com/pmndrs/drei/commit/a2befcd2aca8d2ccdd4be36a120562e69ee19d98)). I tried removing the `.set(0, 1, 0)` in `mainCamera.up.set(0, 1, 0).applyQuaternion(q1).normalize()`  [line 102 in GizmoHelper.tsx](https://github.com/pmndrs/drei/blob/74073db27b62cfb79b8a95135f2ed708cb5b9284/src/core/GizmoHelper.tsx#L102) but that did not work.

cjimmy:
[...]

PR #1330, opened 2023-03-04
fix: ensure `<GizmoHelper />` rotations respsect camera up direction

mdtusz:
### Why

This change resolves an issue when using the GizmoHelper with a non-standard "up" vector for your world (e.g. +Z up rather than the default +Y). By setting the up vector when computing each tweened rotation for the animation, the world up can be presevered and avoid the flicker to set the camera position up after the animation completes. 

To demo this change, create a scene with a camera that has `up={[0, 0, 1]}` using a gizmo to view the behaviour where the view "snaps" to it's final position and does not obey the up vector during the animation. 

While this resolves this issue, the `GizmoViewport` will still be presented in the default world orientation (i.e. with +Y up) which should be addressed in a separate PR. 

Relates to #522 

### What

When computing the quaternion rotation to apply, the `mainCamera` up vector is included to ensure the rotation remains upright. 

### Checklist

- [ ] Storybook entry added
- [ ] Ready to be merged

### DIFF
--- a/src/core/GizmoHelper.tsx
+++ b/src/core/GizmoHelper.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { useFrame, useThree } from '@react-three/fiber'
-import { Camera, Group, Intersection, Matrix4, Object3D, Quaternion, Raycaster, Scene, Texture, Vector3 } from 'three'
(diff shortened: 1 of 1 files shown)


drcmda:
thanks!

PR #1342, opened 2023-03-14
Fix points raycast

promontis:
There are two issues:

1. `_position` is the world position. In the original code, _position is retrieved from the bufferattribute which is local position

2. the matrix that is applied in the original code is the parent matrixWorld. This is also in line with the line 58.

### DIFF
--- a/src/core/Points.tsx
+++ b/src/core/Points.tsx
@@ -65,7 +65,7 @@ export class PositionPoint extends THREE.Group {
     if (rayPointDistanceSq < localThresholdSq) {
       const intersectPoint = new THREE.Vector3()
       _ray.closestPointToPoint(this.position, intersectPoint)
-      intersectPoint.applyMatrix4(parent.matrixWorld)
+      intersectPoint.applyMatrix4(this.matrixWorld)
       const distance = raycaster.ray.origin.distanceTo(intersectPoint)
       if (distance < raycaster.near || distance > raycaster.far) return
       intersects.push({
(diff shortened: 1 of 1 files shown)


src/core/Points.tsx
```diff
@@ -60,12 +60,12 @@ export class PositionPoint extends THREE.Group {
...
     const localThresholdSq = localThreshold * localThreshold
-    const rayPointDistanceSq = _ray.distanceSqToPoint(_position)
+    const rayPointDistanceSq = _ray.distanceSqToPoint(this.position)
 
     if (rayPointDistanceSq < localThresholdSq) {
       const intersectPoint = new THREE.Vector3()
-      _ray.closestPointToPoint(_position, intersectPoint)
-      intersectPoint.applyMatrix4(this.matrixWorld)
+      _ray.closestPointToPoint(this.position, intersectPoint)
+      intersectPoint.applyMatrix4(parent.matrixWorld)
```
drcmda, line 68: this imo imo not sure. the function of the positionpoint class is to render a headless group with its own matrix which is then used to position the instance. the parents matrix would be the instance itself (THREE.Points) but that one is the same for all points.

promontis: You are only talking about line 68 right?

drcmda:
@promontis is this done?

promontis:
This version is better than the current version, so you might want to merge it in. 

I'm currently running in a couple of problems with this code, namely:
- points will disappear because the boundingBox/Sphere is not updated
- when we zoom in (like x100), the raycast hits too wide

PR #1360, opened 2023-03-25
feat(web): add optional hideable scrollbar to scrollcontrols

### LINKED ISSUE #637: Add an option to hide the scrollbar in Scroll Controls
### It would be great if we can have an option maybe something like a prop that can Hide the scrollbar without losing functionality in ScrollControls

Doesn't seem like there could be any drawbacks on this, but maybe someone can suggest?

I managed to do it by changing the file ScrollControls.js  at line 76

>    el.style.top = '0px';
    el.style.left = '15px'; // changed it here, to hide the scroll bar and retain functionality
    fixed.style.position = 'sticky';

CaptEmulation:
### Why

Fixes #637

### What

Sure, one could just know the right style to apply here in order to hide the scrollbar (I do it myself) but this gives users a nice option.

### Checklist

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/src/web/ScrollControls.tsx
+++ b/src/web/ScrollControls.tsx
@@ -24,6 +24,8 @@ export type ScrollControlsProps = {
    *  then a maxSpeed of e.g. 3 which will clamp the speed to 3 units per second, it may now
    *  take much longer than damping to reach the target if it is far away. Default: Infinity */
   maxSpeed?: number
+  /** hideScrollbar optionally hides the scrollbar */
+  hideScrollbar?: boolean
   enabled?: boolean
   style?: React.CSSProperties
   children: React.ReactNode
@@ -55,6 +57,7 @@ export function ScrollControls({
   enabled = true,
   infinite,
   horizontal,
+  hideScrollbar,
   pages = 1,
   distance = 1,
   damping = 0.25,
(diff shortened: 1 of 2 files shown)


src/web/ScrollControls.tsx
```diff
@@ -108,7 +111,7 @@ export function ScrollControls({
     el.style[horizontal ? 'overflowX' : 'overflowY'] = 'auto'
     el.style[horizontal ? 'overflowY' : 'overflowX'] = 'hidden'
     el.style.top = '0px'
-    el.style.left = '0px'
+    el.style.left = hideScrollbar ? '15px' : '0px'
```
drcmda, line 114: shouldn't it be reversed?

drcmda (changes requested):

src/web/ScrollControls.tsx
CaptEmulation: really sorry but been busy with some unexpected work. Going to get back to you as soon as I can with an example, but may be another couple of weekends

CaptEmulation:
going ahead and closing since I currently have no need for this change

rickyzhangca:
need this for the immersive ux without seeing scrollbar jumping around, especially in infinite scrolls...

PR #1364, opened 2023-03-27
fix: don't import example subpaths for compat

### LINKED ISSUE #1363: Nextjs: Import from drei causes `SyntaxError: Cannot use import statement outside a module`
- `three` version: `0.150.1`
- `@react-three/fiber` version: `8.12.0`
- `@react-three/drei` version: `9.58.4`
- `next` version: `13.2.4`
- `react` version: `18.2.0`
- `node` version: `latest`
- `npm` (or `yarn`) version: `latest`

### Problem description:

It seems that importing anything from drei causes the following error:

```node
SyntaxError: Cannot use import statement outside a module

Call Stack
<unknown>
file:/project/sandbox/node_modules/three-stdlib/utils/BufferGeometryUtils.js (1)
...
```

Here is a small csb showing the issue: https://codesandbox.io/p/sandbox/nervous-andras-l4pkwg

CodyJasonBennett:
Fixes #1363

### DIFF
--- a/src/core/RoundedBox.tsx
+++ b/src/core/RoundedBox.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react'
 import { Mesh, Shape, ExtrudeGeometry } from 'three'
 import { NamedArrayTuple } from '../helpers/ts-utils'
-import { toCreasedNormals } from 'three-stdlib/utils/BufferGeometryUtils'
+import { toCreasedNormals } from 'three-stdlib'
 
 const eps = 0.00001
 


abernier:
oopsie 😬

src/core/RoundedBox.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react'
 import { Mesh, Shape, ExtrudeGeometry } from 'three'
 import { NamedArrayTuple } from '../helpers/ts-utils'
-import { toCreasedNormals } from 'three-stdlib/utils/BufferGeometryUtils'
```
abernier, line 4: @CodyJasonBennett I did this :| but can you explain to me why this was a bad idea? (so I won't make the same error next time) thx

CodyJasonBennett: Drei doesn't transpile subpaths AFAIK. We can fix this with `package#exports` there but that would be a breaking change. For now, we just import from the root.

PR #1437, opened 2023-05-09
`Facemesh` v2

abernier:
### Why / What

`<Facemesh>` is now looking in some direction:

image

We can for example gets the direction with the ref-api:

```tsx
meshRef.current.localToWorld(new THREE.Vector3(0, 0, -1))
```

https://github.com/pmndrs/drei/assets/76580/279c4b44-d2f8-4b7b-8b08-0cc3b1ee0a2c

### Checklist

- [x] Documentation updated
- [x] Storybook entry added
- [x] Ready to be merged


src/core/Facemesh.tsx
```diff
@@ -41,17 +45,15 @@ export const Facemesh = React.forwardRef<FacemeshApi, FacemeshProps>(
   ) => {
     const outerRef = React.useRef<THREE.Group>(null)
     const meshRef = React.useRef<THREE.Mesh>(null)
-    const [geometry] = React.useState(() => new THREE.BufferGeometry().setIndex(FacemeshDatas.TRIANGULATION))
 
     const [sightDir] = React.useState(new THREE.Vector3())
```
krispya, line 51: This should be a callback.
const [sightDir] = React.useState(() => new THREE.Vector3())

src/core/Facemesh.tsx
```diff
@@ -41,17 +45,15 @@ export const Facemesh = React.forwardRef<FacemeshApi, FacemeshProps>(
...
     const [sightDirQuaternion] = React.useState(new THREE.Quaternion())
```
krispya, line 52: This should be a callback.
const [sightDirQuaternion] = React.useState(() => new THREE.Quaternion())

src/core/Facemesh.tsx
```diff
@@ -41,17 +45,15 @@ export const Facemesh = React.forwardRef<FacemeshApi, FacemeshProps>(
...
 
-    const [defaultLookAt] = React.useState(new THREE.Vector3(0, 0, -1))
-    const [origin] = React.useState(new THREE.Vector3(0, 0, 0))
-
-    const [lineGeometry] = React.useState(() => new THREE.BufferGeometry().setFromPoints([origin, defaultLookAt]))
+    React.useEffect(() => {
+      meshRef.current?.geometry.setIndex(FacemeshDatas.TRIANGULATION)
+    }, [])
 
     const [a] = React.useState(new THREE.Vector3())
```
krispya, line 60: This should be a callback.

src/core/Facemesh.tsx
```diff
@@ -41,17 +45,15 @@ export const Facemesh = React.forwardRef<FacemeshApi, FacemeshProps>(
...
     const [b] = React.useState(new THREE.Vector3())
```
krispya, line 61: This should be a callback.

src/core/Facemesh.tsx
```diff
@@ -60,6 +62,9 @@ export const Facemesh = React.forwardRef<FacemeshApi, FacemeshProps>(
     const [ac] = React.useState(new THREE.Vector3())
```
krispya, line 64: This should be a callback.

src/core/Facemesh.tsx
```diff
@@ -60,6 +62,9 @@ export const Facemesh = React.forwardRef<FacemeshApi, FacemeshProps>(
...
     const [bboxSize] = React.useState(new THREE.Vector3())
```
krispya, line 65: This should be a callback.

krispya (commented):
It's better to stick all these vectors into a useState callback. But also, instead of putting each in their own useState callback you can put them in one like this:
```js
const [store] = useState(() => ({ vecA: new Vector3(), vecB: new Vector3(), vecC: new Vector3() }))
store.vecA.set(1,1,1)
```

src/core/Facemesh.tsx
abernier: @CodyJasonBennett also [already advised](https://discord.com/channels/740090768164651008/740093168770613279/1105616773472059474) me to use what he called a lazy setter, but can you explain why: is this for performance/non-blocking reasons? In my mind, executing `new Vector3()` was instant, isn't it?

krispya: Yes and it will execute instantly whenever the component renders. Putting it in the callback will make sure it only executes once. :p

abernier: Maybe I'm wrong, but in my mind, wether `initialState` is an expression or a function:
> This argument is ignored after the initial render.

-- [`useState` doc](https://react.dev/reference/react/useState#parameters:~:text=This%20argument%20is%20ignored%20after%20the%20initial%20render)

krispya: I’d have to sit down to really test it, but what I can say just looking at the code is that even if it is ignoring the initial value after the first render, you are still executing a function as a parameter each render. So you are eating the cost of creating a new object which then gets ignored. If it is a callback then it just doesn’t call the function in the first place.

krispya: It gets worst with a Buffer, which will keep allocating memory without any cleanup.

abernier: ok, I understand that: in effective, the result of `new Vector3()` should be ignored but still executed, so better to put it into a function: `() => new Vector3()` so it is ignore AND not executed ;)

abernier: thank you @krispya for this explanation

abernier:
I leave it with multiple states for now, but great technique to "pack" them together ;)

PR #1454, opened 2023-05-20
fix(MeshRefractionMaterial): work across unique meshes

CodyJasonBennett:
Ensures `MeshRefractionMaterial` is reusable between unique meshes. Before, it was order dependent.

https://codesandbox.io/s/k6uiie

| Torus Last | Diamond Last |
|--------|----|
| image | image |

### DIFF
--- a/src/core/MeshRefractionMaterial.tsx
+++ b/src/core/MeshRefractionMaterial.tsx
@@ -2,7 +2,7 @@ import * as THREE from 'three'
 import * as React from 'react'
 import { useLayoutEffect, useMemo, useRef } from 'react'
 import { extend, ReactThreeFiber, useThree, useFrame } from '@react-three/fiber'
-import { MeshBVH, SAH } from 'three-mesh-bvh'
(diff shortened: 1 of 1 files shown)


src/core/MeshRefractionMaterial.tsx
```diff
@@ -69,8 +69,9 @@ export function MeshRefractionMaterial({
     const geometry = (material.current as any)?.__r3f?.parent?.geometry
     // Update the BVH
     if (geometry) {
+      ;(material.current as any).bvh = new MeshBVHUniformStruct()
       ;(material.current as any).bvh.updateFrom(
-        new MeshBVH(geometry.toNonIndexed(), { lazyGeneration: false, strategy: SAH })
+        new MeshBVH(geometry.clone().toNonIndexed(), { lazyGeneration: false, strategy: SAH })
```
CodyJasonBennett, line 74: This is necessary since `toNonIndexed` is destructive and will deopt rendering.

PR #1493, opened 2023-06-05
feat: useProgress() - additional return value: itemsMap

GabeDahl:
### Why

useProgress() state does not update as fast as Three's DefaultLoadingManager.
As a result, the `item` value may skip many items that get loaded.
Users who want a complete list (for display or verification) of loaded items would have to hijack the `DefaultLoadingManager` callbacks, leaving the useProgress hook inoperable.

### What

Added a raw map outside the React/Zustand batched state updates to keep up with DefaultLoadingManager callbacks.
The actual hook/state value updates are still batched by React/Zustand to preserve render performance, but now they include a comprehensive map of loaded items and status.
I also added some descriptions of the hook return values to answer repeat questions.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -2474,7 +2474,15 @@ A convenience hook that wraps `THREE.DefaultLoadingManager`'s progress status.
 
 ```jsx
 function Loader() {
-  const { active, progress, errors, item, loaded, total } = useProgress()
+  const {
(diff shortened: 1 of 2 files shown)


drcmda:
would this leak memory over time? could it be a weakmap?

Issue #1507, opened 2023-06-13
I can not import drei


richi-coder:
What is the solution to this ?

- `three` version:  0.153.0
- `@react-three/fiber` version: 8.13.2
- `@react-three/drei` version: 9.74.12
- `node` version: 19.6.0
- `npm` (or `yarn`) version: 9.4.0
- `vite` version: 4.3.9

### Problem description:

Once react-three/drei is installed, I get an error in my react-vite project when using this import:

### Relevant code:

```
import { Stats, OrbitControls } from "@react-three/drei"
```
The error is: *Failed to resolve entry for package "@react-three/drei". The package may have incorrect main/module/exports specified in its package.json.*

Please help!
My first time trying react-three/drei

wass08:
I'm also facing this issue with Vite on a new project.
Copying the `package.json` content from a working project with `rm -rf node_modules yarn.lock` doesn't solve this issue. 
A real mystery

zgid123:
I think it is because the new build does not contain the `index.js`, it only contains the `index.cjs.js`, so the `module` cannot be import since it is `"module": "index.js",`

jksntaylor:
I manually changed the module path to ```"module": "index.cjs.js"``` and it seems to be working

zgid123:
no, don't do that, cjs is `commonjs`, but `vite` is using `esm` as I remember, I think you should wait for a new deployment or use the version `9.74.11` at the moment

wass08:
Does it work for you with drei version `9.74.11`?
I have this error as soon as I import anything from `drei`:
```
The plugin "vite:dep-scan" was triggered by this import

    src/App.jsx:1:30:
      1 │ import { OrbitControls } from "@react-three/drei";
```

I tried with those package versions without success:
```
{
    "@react-three/drei": "^9.74.11",
    "@react-three/fiber": "^8.13.0",
    "@types/three": "^0.152.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "three": "^0.153.0"
  }
```

zgid123:
with your package json and yarn.lock is deleted, yarn will install the latest version `9.74.12`, can you try

```json
{
    "@react-three/drei": "9.74.11",
    "@react-three/fiber": "^8.13.0",
    "@types/three": "^0.152.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "three": "^0.153.0"
  }
```

helgee:
[...]

PR #1509, opened 2023-06-13
fix: drop mediapipe vendor, pin to RC release with CJS builds

CodyJasonBennett:
Continues #1506, and uses an RC release with CJS compat. Since inlining it for transpile, this seems to have broken builds with #1507.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.11.2",
-    "@mediapipe/tasks-vision": "^0.10.1",
+    "@mediapipe/tasks-vision": "0.10.2-rc2",
     "@react-spring/three": "~9.6.1",
     "@use-gesture/react": "^10.2.24",
     "camera-controls": "^2.4.2",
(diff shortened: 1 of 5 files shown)


src/core/FaceLandmarker.tsx
```diff
@@ -34,6 +34,7 @@ export function FaceLandmarker({
   const opts = JSON.stringify(options)
 
   const faceLandmarker = suspend(async () => {
+    const { FilesetResolver, FaceLandmarker: FaceLandmarkerImpl } = await import('@mediapipe/tasks-vision')
```
CodyJasonBennett, line 37: This might need to be moved to an effect for Next.

package.json
```diff
@@ -55,7 +55,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.11.2",
-    "@mediapipe/tasks-vision": "^0.10.1",
+    "@mediapipe/tasks-vision": "0.10.2-rc2",
```
CodyJasonBennett, line 58: This can be changed back to `~0.10` or `^0.10` once `0.10.2` is released.

PR #1545, opened 2023-06-29
semantic-release optims

abernier:
### Why / What

- using local devDependencies `semantic-release` package (rather than 3rd-party `cycjimmy/semantic-release-action`)
- simplifying the workflow: only 1 job now

### Checklist

- [x] ~~Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example))~~
- [x] ~~Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/Example.stories.tsx))~~
- [ ] Ready to be merged

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,42 +16,14 @@ jobs:
   main:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout repo
-        uses: actions/checkout@v3
-      - name: Setup node
-        uses: actions/setup-node@v3
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
           cache: 'yarn'
-      - name: Install deps
-        # this runs a build script so there is no dedicated build
-        run: yarn install
-      - name: Run storybook build script
-        run: yarn build-storybook
-  release:
-    needs: main
-    runs-on: ubuntu-latest
-    if: ${{ github.repository == 'pmndrs/drei' && contains('refs/heads/master,refs/heads/beta,refs/heads/alpha',github.ref) && github.event_name == 'push' }}
-    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -16,42 +16,14 @@ jobs:
...
           cache: 'yarn'
-      - name: Install deps
-        # this runs a build script so there is no dedicated build
-        run: yarn install
-      - name: Run storybook build script
-        run: yarn build-storybook
-  release:
-    needs: main
-    runs-on: ubuntu-latest
-    if: ${{ github.repository == 'pmndrs/drei' && contains('refs/heads/master,refs/heads/beta,refs/heads/alpha',github.ref) && github.event_name == 'push' }}
```
abernier, line 33: there is actually no need to check:
- `contains('refs/heads/master,refs/heads/beta,refs/heads/alpha',github.ref)` because semantic-release won't release any branch that is not declared in `release.config.js`.`branches`
- `github.event_name == 'push'` because semantic-release will ignore PR: `"This run was triggered by a pull request and therefore a new version won't be published."`

.github/workflows/release.yml
```diff
@@ -16,42 +16,14 @@ jobs:
...
-            @semantic-release/git
-          branches: |
-            [
-              'master',
-              {name: 'beta', prerelease: true},
-              {name: 'alpha', prerelease: true}
-            ]
+      - run: yarn install
+      - run: yarn build-storybook
+      # semantic-release skips not configured branches(see: release.config.js) or pull-requests
```
abernier, line 25: we can securely execute `yarn release` here, because semantic-release will not make a release for PRs or not configured pushed-branches

Issue #391, opened 2021-05-09
ContactShadows disappear at positive values of position-y


romarybi:
- `three` version: 0.123.0
- `@react-three/fiber` version:6.0.12
- `@react-three/drei` version: 4.1.8

### Problem description:

ContactShadows disappear when the y-position is set to any positive value. For instance position={[0,0,0]} works fine, the shadows are displayed, but at position={[0,1,0]}, there is nothing. All negative values seem to work, the problem only occurs at positive values. Am I missing something or is this a bug?

joshuaellis:
Can you provide a code sandbox to show the issue please?

drcmda:
it doesn't sound like a bug to me, all this does is point a camera up and film with a really short far value. you need to adjust the "far" props or it cuts off. all these smoke and mirror effects are really hard to control.

romarybintsev:
Here is a sandbox, adjusting the far prop doesn't fix the issue. 
https://codesandbox.io/s/contact-shadows-0ul6z

CodyJasonBennett:
This is because the scene is being clipped by the near frustum plane. I added a `near` prop in #1556 you can use to customize this behavior.

https://codesandbox.io/s/m5t7x9

PR #1547, opened 2023-06-29
fix: enable positive Y positioning in ContactShadows

NordlingDev:
closes issue #391

### Why

There was no way to set positive Y position in `ContactShadows` as the position of the internal `OrthographicCamera` wasn't compensating the distance of the shadow's position.

This fix resolves issue #391.

### What

The solution is to use the value of incoming position Y property (if defined) and set that on the Z position of the `OrthographicCamera`.

### Checklist

- [x] Storybook entry slightly updated ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/ContactShadows.stories.tsx))
- [x] Ready to be merged

### DIFF
--- a/.storybook/stories/ContactShadows.stories.tsx
+++ b/.storybook/stories/ContactShadows.stories.tsx
@@ -24,14 +24,14 @@ function ContactShadowScene({ colorized }: any) {
         <meshBasicMaterial color="#2A8AFF" />
       </Sphere>
       <ContactShadows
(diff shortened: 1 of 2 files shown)


drcmda:
sorry i had to revert, it breaks contact shadows - some reports of that in bruno simons threejs journey

Screenshot 2023-07-03 at 14 28 56


Screenshot 2023-07-03 at 14 29 01

PR #1564, opened 2023-07-07
fix: gizmoHelper throw some error when use CameraControls, add support

luocongqiu:
### Why

GizmoHelper throw some error when use CameraControls as default Controls,  focusPoint.current is undefined

### What

support CameraControls

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/GizmoHelper.tsx
+++ b/src/core/GizmoHelper.tsx
@@ -1,3 +1,4 @@
+import CameraControls from 'camera-controls'
 import * as React from 'react'
 import { useFrame, useThree } from '@react-three/fiber'
 import { Camera, Group, Matrix4, Object3D, Quaternion, Vector3 } from 'three'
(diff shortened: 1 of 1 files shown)


src/core/GizmoHelper.tsx
```diff
@@ -115,8 +118,23 @@ export const GizmoHelper = ({
...
+              defaultControls.updateCameraUp()
+              defaultControls.setLookAt(
+                mainCamera.position.x,
+                mainCamera.position.y,
+                mainCamera.position.z,
+                focusPoint.current.x,
+                focusPoint.current.y,
+                focusPoint.current.z
+              )
+            } else if (isOrbitControls(defaultControls)) {
```
drcmda, line 134: this imo would now tie it to either cc or oc, but all other controls would break

drcmda (changes requested):

FindDefinition:
We should use method in camera controls instead of set camera position directly.
see this [codesandbox](https://codesandbox.io/s/objective-newton-dmp7tt) with my change, this pull request will cause problem if we click XYZ in sequence. check the problem in this [codesandbox](https://codesandbox.io/s/elated-swanson-477zln) 
In example above, we don't use `CameraControls` in drei because the `CameraControls` in drei prevent update during `controls.enabled === false`. `CameraControls` should be disabled during animation.

PR #1550, opened 2023-06-30
fix: interop Zustand for CJS

CodyJasonBennett:
Fixes #1544.

PR #1537, opened 2023-06-24
rollup3

abernier:
### Why / What

Upgrading to rollup v3

### Checklist

- [x] ~~Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example))~~
- [x] ~~Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/Example.stories.tsx))~~
- [ ] Ready to be merged

PR #1570, opened 2023-07-12
fix: revert Rollup 3 => 2

CodyJasonBennett:
Reverts #1537 since Rollup 3 is not correctly interoping default exports (see #1550, #1551). Rather than do this for all dependencies, I'm reverting until we have a safer process.


src/core/useProgress.tsx
```diff
@@ -1,9 +1,5 @@
 import { DefaultLoadingManager } from 'three'
-import zustand from 'zustand'
-
-// TODO: upgrade to v4
-// https://github.com/pmndrs/drei/pull/1550
-const create = ((zustand as any).default || zustand) as typeof zustand
```
abernier, line 6: oh, this is the reason why? (`.default`)

PR #1585, opened 2023-07-22
fix: Enhance contact shadows

FarazzShaikh:
### What

Fixes the following issues in ContactShadows
- Handles case where `autoClear: false` is set on `<Canvas />`
- Respects `castShadow` property on meshes
   - Useful for excluding meshes from Contact Shadows, this is not possible in the current implimentation. 
   - This also makes the shadows less expensive
- Respects customDepthMaterial to allow for shaders to affect the shadow. See attached box and video. 
   - Notice the contact shadow mirroring into alpha falloff and displacement of the mesh. This is because a customDepthMaterial is provided.
   - Using customDepthMaterial means that the distance attentuation feature will be left up to the user to impliment within their customDepthMaterial.
- Refactors some of the code to improve readibility
- Clean up resources allocated in useMemo

Box: https://codesandbox.io/s/black-leftpad-jmfyjm

https://github.com/pmndrs/drei/assets/55190601/f718dc01-519a-4070-b99a-f22e1c8d0990

https://github.com/pmndrs/drei/assets/55190601/fab07b32-3ad6-4f3a-822d-a4476d689101

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/ContactShadows.tsx
+++ b/src/core/ContactShadows.tsx
@@ -6,9 +6,6 @@ import * as THREE from 'three'
 import { useFrame, useThree } from '@react-three/fiber'
 import { HorizontalBlurShader, VerticalBlurShader } from 'three-stdlib'
 
-// @ts-ignore
-import { FullScreenQuad } from 'three/examples/jsm/postprocessing/Pass'
-
 function isMesh(obj: THREE.Object3D): obj is THREE.Mesh {
   return (obj as THREE.Mesh).isMesh
 }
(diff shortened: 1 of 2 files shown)


src/core/ContactShadows.tsx
```diff
@@ -118,34 +122,49 @@ export const ContactShadows = React.forwardRef(
...
-        ref.current.visible = false
-        scene.background = null
-        scene.overrideMaterial = depthMaterial
+        count++
 
         gl.setRenderTarget(renderTarget)
-        gl.render(scene, shadowCamera.current)
+        gl.clear()
+
+        scene.traverse((obj) => {
```
drcmda, line 134: but wouldn't it be very expensive, given a large scene, to do this every frame?

PR #1593, opened 2023-08-01
fix(useAnimations): ensure API result updates with `clips` and `root`

### LINKED ISSUE #1481: useAnimations is not updating on clip effects
- `three` version: `0.153.0`
- `@react-three/fiber` version: `8.13.0`
- `@react-three/drei` version: `9.70.1`
- `node` version: `18.3.0`
- `yarn` version: `1.22.19`

### Problem description:

I have a use-case where I need to be able to load and play animation clips from different source files (_.glb_). I am loading the source files by updating and passing the `path` parameter to the `useGLTF` hook.

I can confirm that changing the `path` produces a side effect that updates the loaded scene, as expected. However, the updated `animations` clip array does not appear to trigger an update to the object returned by `useAnimations`.

It seems the `useAnimations` hook will only refresh the cache of animations if the component is unmounted ([useAnimations.tsx#L42](https://github.com/pmndrs/drei/blob/c959e0d4232f541130dcd25aced22c6f9fb63138/src/core/useAnimations.tsx#L42)). Is this the intended behavior? If so, are there recommended ways for swapping animations without re-mounting a component?

### Relevant code:

Below is just a test to demonstrate what I am experiencing. Every 3 seconds, the `path` to the animation file changes. I am testing with 2 files that have a different amount of animation clips. I expect the `isEqual` output to _always_ be `true` and the sorted arrays to be the same.

What actually happens is the `animations` array changes with every side effect, but the `names` array _never_ changes after the component is initially mounted.

```tsx
const ANIMATIONS = [`animations_01.glb`, `animations_02.glb`]

const TestPage = () => {
    const [activeIndex, setActiveIndex] = useState(0)
    useEffect(() => {
        const interval = setInterval(() => {
            setActiveIndex(activeIndex => (activeIndex + 1) % ANIMATIONS.length)
        }, 3000)
        return () => clearInterval(interval)
    }, [])

    return (
        <Canvas>
            <Suspense fallback={null}>
                <AnimationChild path={ANIMATIONS[activeIndex]} />
            </Suspense>
        </Canvas>
    )
}

const AnimationChild = ({ path }: { path: string }) => {
    const { scene, animations } = useGLTF(path)
    const { names } = useAnimations(animations, scene)

    useEffect(() => {
        const sortedAnimations = animations.map(animation => animation.name).sort()
        const sortedNames = names.sort()

        console.log({
            sortedAnimations,
            sortedNames,
            isEqual: sortedAnimations.join('') === sortedNames.join(''),
        })
    }, [animations, names])

    return <primitive object={scene} />
}
```

### Suggested solution:

I think effects on the `clips` property should cause an update to the object returned by `useAnimations` during any render, not just in the cleanup function. It should be the responsibility of parent components to decide when the clips cache should be refreshed.

### LINKED ISSUE #1548: useAnimations does not update actions
- `three` version: 0.152.2
- `@react-three/fiber` version: 8.13.0
- `@react-three/drei` version: 9.68.6
- `node` version: v18.13.0
- `yarn` version: 1.22.19

### Problem description:
`useAnimations` does not update `actions` after updating `animations` (after changing the model source of `useGLTF`)

### Relevant code:

```ts
const ref = useRef<THREE.Group | null>(null);
const { animations } = useGLTF(mapModel);
const { actions } = useAnimations(animations, ref);
  
return <primitive ref={ref} object={scene} />;
```

arcturus3:
### Why

Previously `actualRef` did not update if a new `root` argument was passed. Additionally, the `clips`, `names`, and `actions` of the API did not update if a new `clips` argument was passed.

Resolves #1481, resolves #1548.

### What

Correctly update the API and ref in appropriate effects.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/useAnimations.tsx
+++ b/src/core/useAnimations.tsx
@@ -21,8 +21,7 @@ export function useAnimations<T extends AnimationClip>(
   const [mixer] = React.useState(() => new AnimationMixer(undefined as unknown as Object3D))
   React.useLayoutEffect(() => void ((mixer as any)._root = actualRef.current), [mixer, root])
   const lazyActions = React.useRef({})
-  // defined because api creation happens both in state initialization and clips update effect
-  const makeApi = () => {
+  const api = React.useMemo<Api<T>>(() => {
     const actions = {} as { [key in T['name']]: AnimationAction | null }
     clips.forEach((clip) =>
       Object.defineProperty(actions, clip.name, {
@@ -39,11 +38,9 @@ export function useAnimations<T extends AnimationClip>(
       })
     )
     return { ref: actualRef, clips, actions, names: clips.map((c) => c.name), mixer }
-  }
-  const [api, setApi] = React.useState<Api<T>>(makeApi)
+  }, [clips])
   useFrame((state, delta) => mixer.update(delta))
   React.useEffect(() => {
-    setApi(makeApi())
     const currentRoot = actualRef.current
     return () => {
       // Clean up only when clips change, wipe out lazy actions and uncache clips
--- a/src/core/useAnimations.tsx
+++ b/src/core/useAnimations.tsx
@@ -18,7 +18,7 @@ export function useAnimations<T extends AnimationClip>(
   const [actualRef] = React.useState(() => (root ? (root instanceof Object3D ? { current: root } : root) : ref))
   // eslint-disable-next-line prettier/prettier
   const [mixer] = React.useState(() => new AnimationMixer(undefined as unknown as Object3D))
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
     actualRef.current = root && (root instanceof Object3D ? root : root.current)
     ;(mixer as any)._root = actualRef.current
   }, [root])
--- a/src/core/useAnimations.tsx
+++ b/src/core/useAnimations.tsx
@@ -16,11 +16,13 @@ export function useAnimations<T extends AnimationClip>(
 ): Api<T> {
   const ref = React.useRef<Object3D>()
   const [actualRef] = React.useState(() => (root ? (root instanceof Object3D ? { current: root } : root) : ref))
+  React.useEffect(() => void (actualRef.current = root && (root instanceof Object3D ? root : root.current)), [root])
   // eslint-disable-next-line prettier/prettier
   const [mixer] = React.useState(() => new AnimationMixer(undefined as unknown as Object3D))
   React.useLayoutEffect(() => void ((mixer as any)._root = actualRef.current), [mixer, root])
   const lazyActions = React.useRef({})
-  const [api] = React.useState<Api<T>>(() => {
+  // defined because api creation happens both in state initialization and clips update effect
+  const makeApi = () => {
     const actions = {} as { [key in T['name']]: AnimationAction | null }
     clips.forEach((clip) =>
       Object.defineProperty(actions, clip.name, {
@@ -37,9 +39,11 @@ export function useAnimations<T extends AnimationClip>(
       })
     )
     return { ref: actualRef, clips, actions, names: clips.map((c) => c.name), mixer }
-  })
+  }
+  const [api, setApi] = React.useState<Api<T>>(makeApi)
   useFrame((state, delta) => mixer.update(delta))
   React.useEffect(() => {
+    setApi(makeApi())
     const currentRoot = actualRef.current
     return () => {
       // Clean up only when clips change, wipe out lazy actions and uncache clips
--- a/src/core/useAnimations.tsx
+++ b/src/core/useAnimations.tsx
@@ -16,10 +16,12 @@ export function useAnimations<T extends AnimationClip>(
 ): Api<T> {
   const ref = React.useRef<Object3D>()
   const [actualRef] = React.useState(() => (root ? (root instanceof Object3D ? { current: root } : root) : ref))
-  React.useEffect(() => void (actualRef.current = root && (root instanceof Object3D ? root : root.current)), [root])
   // eslint-disable-next-line prettier/prettier
   const [mixer] = React.useState(() => new AnimationMixer(undefined as unknown as Object3D))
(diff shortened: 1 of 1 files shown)


src/core/useAnimations.tsx
```diff
@@ -16,11 +16,13 @@ export function useAnimations<T extends AnimationClip>(
 ): Api<T> {
   const ref = React.useRef<Object3D>()
   const [actualRef] = React.useState(() => (root ? (root instanceof Object3D ? { current: root } : root) : ref))
+  React.useEffect(() => void (actualRef.current = root && (root instanceof Object3D ? root : root.current)), [root])
```
CodyJasonBennett, line 19: This needs to be a `useLayoutEffect` for L22.

src/core/useAnimations.tsx
```diff
@@ -37,9 +39,11 @@ export function useAnimations<T extends AnimationClip>(
       })
     )
     return { ref: actualRef, clips, actions, names: clips.map((c) => c.name), mixer }
-  })
+  }
+  const [api, setApi] = React.useState<Api<T>>(makeApi)
   useFrame((state, delta) => mixer.update(delta))
   React.useEffect(() => {
+    setApi(makeApi())
```
CodyJasonBennett, line 46: Perhaps this should be `useMemo` instead. Cleanup can stay in an effect which should address #1594.

arcturus3: Agreed, especially because it eliminates the superfluous render caused by `setApi`. I just pushed this change. But how is #1594 addressed as is?

CodyJasonBennett: I see we'll still need that PR.

src/core/useAnimations.tsx
arcturus3: Yeah, do you see a reason for using a layout effect on L22? If not I'll just move L19 and L22 into the same effect.

CodyJasonBennett: Layout effects happen before the first frame is rendered whereas regular effects are async and will be at least a frame late. You can combine them in any case if you want.

src/core/useAnimations.tsx
```diff
@@ -18,9 +18,12 @@ export function useAnimations<T extends AnimationClip>(
   const [actualRef] = React.useState(() => (root ? (root instanceof Object3D ? { current: root } : root) : ref))
   // eslint-disable-next-line prettier/prettier
   const [mixer] = React.useState(() => new AnimationMixer(undefined as unknown as Object3D))
-  React.useLayoutEffect(() => void ((mixer as any)._root = actualRef.current), [mixer, root])
+  React.useEffect(() => {
```
CodyJasonBennett, line 21: This should still be a `useLayoutEffect` so it's not a frame late. Otherwise LGTM.

arcturus3: My understanding of `useLayoutEffect` is that it runs after the component renders (just like `useEffect`), but before the browser paints. And in this case, it's not necessary to have `actualRef.current` and `mixer._root` updated for the browser paint, only for the next time they're referenced as the result of another effect or component render. Since this should be guaranteed by `useEffect` and [it's preferred over `useLayoutEffect` when possible](https://react.dev/reference/react/useLayoutEffect), I think it's the better choice. Let me know if I'm missing something here, and either way, I'm happy to switch it back to `useLayoutEffect` just in case if you'd like.

CodyJasonBennett: The main thing to consider here is whether this needs to be updated in this effect to ensure a correct result on first render. If that is not relevant or is guaranteed some other way (like via a constructor), then we should keep `useEffect` since React is better able to schedule this as it's async and can be scheduled after many frames whereas `useLayoutEffect` is synchronous and will block work at commit.

arcturus3: Right, I think the initial render will behave the same whether we go with `useEffect` or `useLayoutEffect` because they're functionally equivalent in this case (with `useEffect` being marginally better for performance). I'm not sure why `mixer` is being initialized with `new AnimationMixer(undefined)` instead of `new AnimationMixer(<the passed root>)`, but that seems like a separate issue and using `useLayoutEffect` won't set the root properly in time for the first render anyway.

CodyJasonBennett: You're confusing React render with three.js/browser render. Only `useLayoutEffect` will fire before the latter because of how microtasks are scheduled in the browser. It's not a performance issue as much as whether you want to block those operations, which I'm a little unhappy about the lack of clarification in the docs for.

arcturus3: I haven't confused them, I've been referring to React render as "render" and browser render as "paint". Sorry if I wasn't clear. If you're saying this effect should happen by the next browser paint then I totally agree that we need `useLayoutEffect`.

Originally I was thinking that this does not need to happen by the next browser paint because `mixer._root` would not be referenced by anything until the next React render cycle (that is, only the user's React code would be touching `mixer`). But in thinking about it more, I suppose (1) it is quite possible that Three.js would be referencing this before browser paint for some reason, and (2) `requestAnimationFrame` could be scheduled before browser paint and expecting the updated value (e.g. to call `mixer.update`).

Considering all that I think `useLayoutEffect` may be the correct choice, do you agree?

CodyJasonBennett: I'd stick with `useLayoutEffect` so we don't get any surprises. I'm not worried about perf ramifications in a simple mutation case.

drcmda:
@arcturus3 @CodyJasonBennett i had to revert, it broke uneAnimation for most boxes i've tried it with, for instance https://codesandbox.io/s/pecl6?file=/src/Model.js actions are null in useEffect for some reason. @arcturus3 could you look at it and find the problem? then we make a unified pull again with all changes included.

arcturus3:
@drcmda Yep, I'll take a look at it.

arcturus3:
@drcmda @CodyJasonBennett See #1620.

PR #1602, opened 2023-08-05
generic type for shaderMaterial, exposing uniform props

xinaesthete:
### Why

I and at least one other person were having trouble with the type of `shaderMaterial`, as discussed in [this StackOverflow question](https://stackoverflow.com/questions/75901442/using-ref-on-shadermaterial-causes-typescript-error-the-expected-type-comes-fr/76842750#76842750).

Actually, as of this writing I'm not really entirely sure what the relationship to `useRef` is, but this seems like the right approach for the returned type of `shaderMaterial`.

### What

I've added a type parameter to `shaderMaterial`, inferred from the type of the `uniforms` argument, and union-ed that with the returned material type. As such, the type of the returned value should reflect the fact that each of the uniforms is available as a prop of appropriate type.

### Checklist

- [ ] Ready to be merged

I haven't actually got as far as using the fork, properly testing or updating documentation - but I'm using an equivalent method in my local project and it appears to work right. I'd be happy to make relevant changes to the documentation and storybook as well, but wanted to get this into a form where hopefully maintainers can review it while it's fresh in my mind.

### DIFF
--- a/src/core/shaderMaterial.tsx
+++ b/src/core/shaderMaterial.tsx
@@ -1,24 +1,26 @@
 import * as THREE from 'three'
 
-export function shaderMaterial(
-  uniforms: {
-    [name: string]:
-      | THREE.CubeTexture
-      | THREE.Texture
-      | Int32Array
-      | Float32Array
-      | THREE.Matrix4
-      | THREE.Matrix3
-      | THREE.Quaternion
-      | THREE.Vector4
-      | THREE.Vector3
-      | THREE.Vector2
-      | THREE.Color
-      | number
-      | boolean
-      | Array<any>
-      | null
-  },
+type U = {
+  [name: string]:
+  | THREE.CubeTexture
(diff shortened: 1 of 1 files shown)


xinaesthete:
So when I saw failing CI stuff before I couldn't see the actual errors and I wasn't sure if it was to do with actual problems in the actual code vs not running properly because of who I was or something.

I realised the thing that was failing was an existing use of `shaderMaterial` in `MeshReflectionMaterial`, where the actual original call to `shaderMaterial` had `envMap: null, ..., resolution: THREE.Vector2` which then were incompatible with their use...

So I still think the change I made in the PR is basically a good one - I'd certainly prefer to have this type inferred, and for that to - but there could be user-code that would get broken in a similar way if it got through to a release.

I'm also not entirely sure that the way I fixed issues with existing code in my last commit is necessarily the best - still had some niggles with getting it .

Issue #1615, opened 2023-08-17
`<ScrollControls infinite/>` current scroll position is less than `scrollThreshold`, leads to unable reset to top.


nhtoby311:
- `three` version: 0.155.0
- `@react-three/fiber` version: 8.13.6
- `@react-three/drei` version: 9.80.1

### Problem description:

I tested this with Samsung A53, on Chrome, and I notice that my scroll does not got "reset" back to top when reaching the bottom scroll bar. A bit further look into the <ScrollControls/> code, I notice that at the bottom scroll, my current scroll position `currentScrollPos` is less than `scrollThreshold` by a small amount, ~0.44, therefore, it never got reset to the top.

![image](https://github.com/pmndrs/drei/assets/52330522/8970017e-44dd-41d1-8fbb-5d50c92ad786)
 
 
### Relevant code:

```js
//ScrollControls
if (infinite) {
          if (!disableScroll) {
          //This does not get triggered in my case!
            if (current >= scrollThreshold) {
....
```

### Suggested solution:

Although it is quite specific for certain devices, and browser combination, but I think it is good if we have an option to offset this threshold, so we can adjust this to work well on those cases as well.

PR #1616, opened 2023-08-17
fix: add `thresholdOffset` for infinite scrolling

nhtoby311:
### Why

Regarding this bug #1615, on some device and browser combination, the scroll does not reach the `scrollThreshold`, therefore, it does not reset the scroll bar back to top, and results in broken infinite scroll effect. 

### What

I simply add a props name `thresholdOffset` to offset this `scrollThreshold`. 

Could not identify the reason why the calculation of current scroll position is off by ~0.44 in my test using Samsung A53 with Chrome, as it varied between device and browser. However, I think this solution at least give us the option to tackle these edge cases.

image

### Checklist

- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,8 @@ type ScrollControlsProps = {
   /** Infinite scroll, default false (experimental!) */
   infinite?: boolean
   /** Defines the length of the scroll area, each page is height:100%, default 1 */
+  thresholdOffset?: number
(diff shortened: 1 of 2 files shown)


drcmda:
but this would require knowledge around these edge cases and magic numbers deployed in user land. 😕 did you find any more info on why these platforms never reach the threshold?

PR #1621, opened 2023-08-22
Update useEnvironment.tsx

SanjayBalnad:
Throwing error when using Environment component, so I just found a flaw and added a comment how to fix it. Same link might be there in others files also

### Checklist

- [ ] Ready to be merged


drcmda:
i think your drei might just be outdated? these links were recently changed to fix an outage. the old links went to a different cdn.

PR #1631, opened 2023-08-27
fix(AccumulativeShadows): set progressivelightmap Texture to halffloattype and nearestfilter

### LINKED ISSUE #1611: AccumulativeShadows goes all black on iPad Safari
- `three` version: 0.155.0
- `@react-three/fiber` version: 8.13.6
- `@react-three/drei` version: 9.80.3
- `node` version: 18.12.1
- `npm` (or `yarn`) version: 1.22.19

### Problem description:

On iPadOS Safari, where AccumulativeShadows is used, the entire plane slowly turns black instead of a shadow appearing on the plane.
All the examples in CodeSandbox have the same problem.

I confirmed this with Safari on my iPadOS 16.6, but I don't know if this is an iOS-wide issue as I don't now have an iPhone.

I have checked and it seems to go black regardless of the option values set, such as `alphaTest`, regardless of how large or small they are. As the value of `alpha` decreases, black planes also become transparent.

### Relevant code:

The following official samples where AccumulativeShadows is used have this problem. 

https://codesandbox.io/s/hxcc1x
IMG_0003

https://codesandbox.io/s/19uq2u
IMG_0002

hitmarten:
### Why

resolves #1611 
ProgressiveLightMap is not being rendered correctly because the wrong texture format being used on iPadOS with Safari.

### What

Currently the code checks the user agent to determine what float type to use (half float or float). When the user agent is being spoofed, texture is not rendered correctly (this happens by default on iPadOS with Safari). 

To use [THREE.FloatType ](https://threejs.org/docs/#api/en/textures/DataTexture) the browser must support the 	OES_texture_float and OES_texture_half_float WebGL extension. 
I added a check for those extensions to determine what float type to use.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/AccumulativeShadows.tsx
+++ b/src/core/AccumulativeShadows.tsx
@@ -349,13 +349,13 @@ class ProgressiveLightMap {
     this.clearAlpha = 0
 
     // Create the Progressive LightMap Texture
-    const format = /(Android|iPad|iPhone|iPod)/g.test(navigator.userAgent) ? THREE.HalfFloatType : THREE.FloatType
-    this.progressiveLightMap1 = new THREE.WebGLRenderTarget(this.res, this.res, {
-      type: format,
-    })
-    this.progressiveLightMap2 = new THREE.WebGLRenderTarget(this.res, this.res, {
-      type: format,
-    })
+    const textureParams = {
+      type: THREE.HalfFloatType,
+      magFilter: THREE.NearestFilter,
+      minFilter: THREE.NearestFilter,
+    }
+    this.progressiveLightMap1 = new THREE.WebGLRenderTarget(this.res, this.res, textureParams)
+    this.progressiveLightMap2 = new THREE.WebGLRenderTarget(this.res, this.res, textureParams)
 
     // Inject some spicy new logic into a standard phong material
     this.discardMat = new DiscardMaterial()


CodyJasonBennett:
`OES_texture_float` is only available in WebGL 1; it's inbuilt into WebGL 2 as part of core and will always return `null` from a `WebGL2RenderingContext` which is a false positive -- `gl.FLOAT` is well defined. Perhaps the issue is with `OES_texture_float_linear` which is still an optional extension and enables the default filtering configuration (with mips) in three.js.

hitmarten:
I could swap the `gl.getExtension('OES_texture_float')` with `renderer.capabilities.floatFragmentTextures`. This checks for webgl2 or the `OES_texture_float` extension.

I also tested the the half floats en nearest filtering like you suggested. This also fixed the issue.

Texture with `THREE.FloatType` and `THREE.LinearFilter`
FloatLinear

Texture with `THREE.HalfFloatType` and `THREE.NearestFilter`
HalfFloatNearer

Also tried `THREE.UnsignedByteType`. but that looked visibly worse.
UnsignedLinear


I don't see any difference between the first two textures. Would you like me to change the texture to `THREE.HalfFloatType` and `THREE.NearestFilter` instead of using the check?

CodyJasonBennett:
I'd prefer `HalfFloatType` with `NearestFilter` in that case.

PR #1638, opened 2023-09-04
feat: Bounds update

STINGNAILS:
### Why
There are few problems with current Bounds implementation:
1. It is impossible to, for example, change camera position / orientation and fit view to the expected bounds (in the end of transition) simultaneously. Also it isn't possible to orient camera.up and animate it respectively.
2. The animation pattern is fixed (it is the standart `THREE.MathUtils.damp(...)`), and there is little control over it (just the `damping` value)

### What
Instead of `to` (which is deprecated, not removed), now there are functions `moveTo` and `lookAt`. Some of `fit` functionality got moved to a new `reset` function.
`refresh` only resets goal values. `reset` repositions and rotates the camera to look at the center (i.e., what `fit` did before except zooming). `moveTo` changes the camera position. `lookAt` changes the camera orientation, with the respect to `up` vector, if specified. `fit` only changes zoom value. `clip` changes near / far values, as before. Each of these functions respects previously called function's updated goals.
Now there also are: an optional `interpolationFunc` callback which should reinterpolate the animation progress value (in [0, 1] interval), and `maxDuration` which specifies the animation length (in seconds).

### Checklist
- [x] Modify the `clip` function: either remove clamping values to 0.1 and 1000, or allow passing these values as props explicitly (forgot to do...)


FarazzShaikh:
Hello @STINGNAILS @drcmda. This PR breaks the `.fit()` method.

`9.88.17`: https://codesandbox.io/s/bounds-and-makedefault-forked-z5zrcy
`9.89.0`: https://codesandbox.io/s/bounds-and-makedefault-forked-gppgq3

It seems like some changes were made to the legacy methods. The docs dont seem to reflect these changes. I am unsure how to use the bounds API now. Can we please update the docs?

STINGNAILS:
> Hello @STINGNAILS @drcmda. This PR breaks the `.fit()` method.
> 
> `9.88.17`: https://codesandbox.io/s/bounds-and-makedefault-forked-z5zrcy `9.89.0`: https://codesandbox.io/s/bounds-and-makedefault-forked-gppgq3
> 
> It seems like some changes were made to the legacy methods. The docs dont seem to reflect these changes. I am unsure how to use the bounds API now. Can we please update the docs?

With 9.92.2, it seems to work as before with fit (for non-orthographic camera): https://codesandbox.io/p/sandbox/bounds-and-makedefault-forked-rf62ks
I'll also make a PR for docs update (this week hopefully)

Issue #1385, opened 2023-04-04
<Selection> deselects selected object on second click - can't configure to behave differently


soadzoor:
- `three` version: 150.1
- `@react-three/fiber` version: 8.12.0
- `@react-three/drei` version: 9.62.2
- `node` version: 18.15.0
- `npm` (or `yarn`) version: 9.5.0

### Problem description:

When I select an object using the `<Select>` component, it works pretty nicely, but when I click on the same object again, it gets deselected. It makes sense in some cases, but for my use-case, it doesn't. Since I'm also providing the ability to drag/grab the object and move it around with the mouse, it's weird how it always gets deselected whenever the user moves it a little bit, so they have to click it again.

I couldn't find a property or option where this behavior could be modified. In fact, I can barely find any documentation on this component, except for some examples I could find on codesandbox.

### Relevant code:

https://codesandbox.io/s/ny3p4?file=/src/App.js:475-483

### Suggested solution:

Provide an attribute, something like `deselectOnSecondClick={false}`, which would prevent the above mentioned behavior. Of course, when someone clicks the object with the "shift" key being held, it would make sense for my use-case as well, to deselect the clicked object on the second click, so there might be multiple attributes/properties required to satisfy all the use-cases...

Haschtl:
I had the same issue now.
It's a very simple fix, I created a pull-request for it: https://github.com/pmndrs/drei/pull/1648

PR #1648, opened 2023-09-09
Select: Add deselectOnSecondClick prop (#1385)

Haschtl:
### Why

If you combine <Select> with <TransformControls> you will notice, that an Object is deselected after you stop dragging it. In some scenarios it's unintuitive that the object is deselected on the second click.

### What

Added the deselectOnSecondClick prop (default: true) to Select, which allows to disable the deselect-on-second-click behavior.


drcmda:
if you drag it and let the button go - imo if it deselects that is a bug, no? it seems unintuitive.

PR #1673, opened 2023-10-04
docs: capitalize spelling of <Points> in README

seantai:
### Why

https://github.com/pmndrs/drei/tree/master#pointmaterial

Assuming `<PointMaterial>` docs should use the `<Points>` Drei component (as it does in CSB example), instead of extending the native three points or something. 

- [X ] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example))
- [X ] Ready to be merged


drcmda:
i think that's not a mistake. `<points>` is valid, it's THREE.Points and the material works with it as well.

PR #1695, opened 2023-10-21
fix(Html): use transform for perspective browser issues

### LINKED ISSUE #428: <Html /> becomes misaligned when placed in a grid
- `three` version: 0.129.0
- `@react-three/fiber` version: 6.2.2
- `@react-three/drei` version: 5.3.0

### Problem description:

When I place a model inside a CSS grid, and I use the grid system to resize and place the item, the items within the Html tags are no longer aligned with the surface of the model.

When the page is resized, the items in the HTML tag also jump around.

### Relevant code:

Here is the codesandbox with the code: https://codesandbox.io/s/mixing-html-and-webgl-w-occlusion-forked-ct4fl?file

This only seems to occur when the initial grid or row is not 1.

For example, this code works

```css
.canvas {
  grid-row: 1/11;
  grid-column: 1/11;
}
```

However, this code does not

```css
.canvas {
  grid-row: 4/11;
  grid-column: 4/11;
}
```

Please let me know if you are not meant to use containers with CSS to place react-three-fiber models

CodyJasonBennett:
Fixes #428, mirrors https://github.com/mrdoob/three.js/pull/27017.

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -341,7 +341,7 @@ export const Html: ForwardRefComponent<HtmlProps, HTMLDivElement> = /* @__PURE__
             }
             el.style.width = size.width + 'px'
             el.style.height = size.height + 'px'
-            el.style.perspective = isOrthographicCamera ? '' : `${fov}px`
+            el.style.transform = isOrthographicCamera ? '' : `perspective(${fov}px)`
             if (transformOuterRef.current && transformInnerRef.current) {
               transformOuterRef.current.style.transform = `${cameraTransform}${cameraMatrix}translate(${widthHalf}px,${heightHalf}px)`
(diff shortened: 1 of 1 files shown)


src/web/Html.tsx
```diff
@@ -341,7 +341,7 @@ export const Html: ForwardRefComponent<HtmlProps, HTMLDivElement> = /* @__PURE__
             }
             el.style.width = size.width + 'px'
             el.style.height = size.height + 'px'
-            el.style.perspective = isOrthographicCamera ? '' : `${fov}px`
+            el.style.transform = isOrthographicCamera ? '' : `perspective(${fov}px)`
```
donmccurdy, line 344: This will need to be the first transform on "transformOuterRef" below, rather than a transform on "el".

CodyJasonBennett:
Hey, @drcmda, this wasn't supposed to be merged. Maybe I should have converted into a draft but I needed to investigate @donmccurdy's review further.

PR #1697, opened 2023-10-25
fix: include indirect BVH option

geoff-harper:
### Why

My use case requires that the triangles of my mesh be in their original order. Garrett's library saw this use case and has created an experimental accomodation for it but that isn't currently carried through to the `<Bvh>` component. I know it's marked as experimental but I've been using it for some time now and I've had no issues. I think as long as Drei doesn't obscure it's experimental nature it's fine to include?

### What

Includes the `indirect` option from [three-mesh-bvh](https://github.com/gkjohnson/three-mesh-bvh). I've copied his explanation of the option into the docs verbatim.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -3344,6 +3344,14 @@ export interface BVHOptions {
   maxDepth?: number
   /** The number of triangles to aim for in a leaf node, default: 10 */
   maxLeafTris?: number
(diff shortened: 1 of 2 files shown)


drcmda:
sorry for the long wait, and thanks!

PR #1698, opened 2023-10-26
feat(SpriteAnimator) `resetOnEnd` for reset with replay

netgfx:
### Why

Added support for json hash (object based spritesheet).
Added `resetOnEnd` property in order to reset the frames on end but allow replay

### What

JSON Hash spritesheets are now supported. Playback on demand and reset is now available

### Checklist

- [ x] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example))
- [ x] Ready to be merged


src/core/SpriteAnimator.tsx
```diff
@@ -17,12 +17,12 @@ export type SpriteAnimatorProps = {
   onEnd?: Function
   onLoopEnd?: Function
   onFrame?: Function
-  play?: boolean
   pause?: boolean
```
CodyJasonBennett, line 20: Why was `play` removed? I'm not sure I see any conflicts for it.

netgfx: @CodyJasonBennett It was never used, autoPlay does the same thing. It was a clean-up change.

CodyJasonBennett: I see. We're not able to remove or rename APIs as part of a feature, but we can deprecate the property so it can be cleaned up when we release v10 of Drei.

netgfx: @CodyJasonBennett ok sure I'll make the change

netgfx: I reverted the change and marked the property as deprecated on the ReadMe

CodyJasonBennett (approved):

PR #1701, opened 2023-11-03
AccumulatedShadows NPE protection

supertick:
### Why

When component is unmounted it throws.
The way its currently written
```js
    React.useLayoutEffect(() => {
      const group = gLights.current
      if (parent) parent.lights.set(group.uuid, api)
      return () => void parent.lights.delete(group.uuid)
    }, [parent, api])
  ```
  The check for valid parent is done for the lights.set, but there is no check on the
  ```js
        return () => void parent.lights.delete(group.uuid)
   ```
   so if parent is null this ^ line throws

### What
prevents AccumulatedShadows from throwing when parent is null

### DIFF
--- a/src/core/AccumulativeShadows.tsx
+++ b/src/core/AccumulativeShadows.tsx
@@ -307,8 +307,8 @@ export const RandomizedLight: ForwardRefComponent<
     React.useImperativeHandle(forwardRef, () => api, [api])
     React.useLayoutEffect(() => {
       const group = gLights.current
-      if (parent) parent?.lights.set(group.uuid, api)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by drcmda** [comment]
hi @supertick, thanks! could you do this instead?

```jsx
      if (parent) parent?.lights.set(group.uuid, api)
      return () => void parent?.lights.delete(group.uuid)
```

and w/o semicolons please due to the prettier config we use.

supertick:
Sure, updated.
* removed the first parent? because its already tested with if (parent)
* improved formatting :+1:

PR #1718, opened 2023-11-14
[fix] `Bvh` doesn't recompute every rerender

krispya:
This is a simple change that makes it `Bvh` only computes once on mount instead of recomputing every time there is a rerender in the tree.

Reactive recomputing is impractical in real time scenarios. Instead, I think we should encourage users to wrap `Bvh` arouund reusable graphs in components.


krispya:
@drcmda Could you clarify the intention here?

drcmda:
@krispya might have been simple oversight

PR #1791, opened 2024-01-23
fix: prevent successive incline or decline to trigger onFallback in PerformanceMonitor

### LINKED ISSUE #1739: PerformanceMonitor's `onFallback` gets called when it's not flip-flopping
- `three` version: 0.156.1
- `@react-three/fiber` version: 8.14.3
- `@react-three/drei` version: 9.86.3
- `node` version: 20
- `npm` (or `yarn`) version: 9.6.7

### Problem description:

This might be me misunderstanding what a "flip-flop" is - but I'd have expected it to only get called when it does a series of `onDecline` / `onIncline` calls, rather than the series of `onIncline` calls which it does in the sandbox below since it's already running at 60fps.

### Relevant code:

https://codesandbox.io/p/sandbox/amazing-borg-x3v9dr?file=%2Fsrc%2Findex.jsx%3A23%2C9-23%2C19

The cube starts at "perfLevel" 0 where it's red, then 1 where it's orange, then 2 where it's green - but then gets reset to red by the `onFallback` being called even though its only ever flipped and never flopped.

### Suggested solution:

The `onFallback` shouldn't be called if all the latest changes have all been inclines or all declines - only if it's flip-flopping between an incline and a decline

AxiomeCG:
### Why

The flipflops is triggered by the successive call of onIncline or successive call of onDecline, which leads to the call of onFallback.
In particular, if you have successive improvement in performances that leads to the call of your last resort optimization in the onFallback. For example, you ask for the Effects in the EffectComposer to be better in onIncline and to disable totally the Effects in the onFallback, you see improvements, improvements, then you default fallback with no effects which defeat the purpose of the PerformanceMonitor. 

resolves #1739 
### What

I have added a `lastChange` variable that keeps track of the last change to see if it was a decline before the incline or vice versa, which is what consist in a flipflop. It increments the value of `flipped` only if it switch back and forth between incline and decline, which is the instability we are searching in the first place. 
### Checklist

- [x] Ready to be merged


drcmda:
is it ready @AxiomeCG  ?

AxiomeCG:
> is it ready @AxiomeCG ?

I put it in draft as mentioned in the template of the issue, but to me it is done ! :)

PR #1845, opened 2024-02-24
feat(Edges): use LineSegments2

CodyJasonBennett:
Refactors `Edges` to use a mesh-based solution for wider edge widths.


drcmda:
ready to merge? @CodyJasonBennett

CodyJasonBennett:
This needs testing and I've since been traveling. If it looks good to you, then sure. I need to add a Storybook entry since Edges doesn't seem to have one.

drcmda:
tested it, it wouldn't work ootb but added a small fix in an additional commit

PR #1847, opened 2024-02-24
feat: useVideoTexture enable hls support

netgfx:
### Why

To add support for HLS in `useVideoTexture`

### What

Added hlsjs as a dependency, allowed HTML video element of `useVideoTexture` to be enhanced by hlsjs in order to allow playback of variable bit rate videos.

### Checklist

- [x] Ready to be merged


netgfx:
@CodyJasonBennett please take a look when you can. 
I have tested it out here: https://codesandbox.io/p/sandbox/white-smoke-wx5zht?file=%2Fsrc%2FApp.js%3A21%2C1
I would prefer if we could also return the HLS instance object so that people can use it to listen to hls events, but I couldn't make it return in a way that it didn't cause re-renders,let me know if you have some idea about this.

src/core/useVideoTexture.tsx
```diff
@@ -1,24 +1,41 @@
...
   const gl = useThree((state) => state.gl)
+
+  const createHLSBinding = useCallback(() => {
+    if (videoRef.current && typeof src === 'string') {
+      var video: HTMLVideoElement = videoRef.current
+      const _hls = new Hls({ ...hls })
+      _hls.attachMedia(video)
+      _hls.loadSource(src)
+    }
+  }, [])
```
CodyJasonBennett, line 37: Should this teardown at any point? I worry this will create a memory leak.

src/core/useVideoTexture.tsx
```diff
@@ -1,24 +1,41 @@
 import * as THREE from 'three'
-import { useEffect } from 'react'
+import { useCallback, useEffect, useRef } from 'react'
 import { useThree } from '@react-three/fiber'
-import { suspend, preload, clear } from 'suspend-react'
+import { suspend } from 'suspend-react'
+import Hls, { HlsConfig } from 'hls.js'
```
CodyJasonBennett, line 5: I need to check that this is published correctly since default exports are a general hazard. If so, or it doesn't tree-shake, we'll need to move it to a lazy `import` in the guard you have within `suspend`. The module reference can be memoized within this file.

src/core/useVideoTexture.tsx
netgfx: This needs to happen after the video element has been created, can we ensure the useLayoutEffect will run after the video has been created? Since the video instance is passed to a ref and not a state variable.

CodyJasonBennett: Yes, effects won't fire until suspense resolves.

netgfx: Sure I'll try it out and update the PR, thanks!

netgfx: @CodyJasonBennett I've tried it, you can check it here: https://codesandbox.io/p/sandbox/white-smoke-wx5zht?file=%2Fsrc%2FuseHLSVideoTexture.js%3A51%2C1

But it doesn't work. I believe that this:
```
 const _hls = new Hls({ ...hls })
 _hls.attachMedia(video)
 ```
 
 needs to happen before this: `const texture = new THREE.VideoTexture(video)`
 
 if that is true it has to happen inside the promise

src/core/useVideoTexture.tsx
drcmda: @CodyJasonBennett is it good to go? @netgfx

netgfx: @drcmda I think I need to make sure that HLS gets removed when the component unmounts, it is a bit tricky because when setting ref values from inside the promise those aren't available outside of it like in the useEffect in order to call them on the useEffect return

CodyJasonBennett (approved):

netgfx:
@CodyJasonBennett I think I got it right this time, it should properly teardown now.

PR #1868, opened 2024-03-09
fix: memoize result of useTexture

roelkok:
### Why

When passing the same values to `useTexture`, I would expect the same return value. However, when passing an object as the input parameter, a different object is returned each time even when the same input object is used.

Example:
```
const urlMap = useMemo(() => {
  return {
    textureA: urlA,
    textureB: urlB
  }
}, [urlA, urlB])

const textureMap = useTexture(urlMap);

useEffect(() => {
  // This will run every render. Even if urlA and urlB (and therefore urlMap) haven't changed.
}, [textureMap])

```

### What

To give a more expected result, memoize the return value when dealing with an object as input.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/useTexture.tsx
+++ b/src/core/useTexture.tsx
@@ -1,6 +1,6 @@
 import { Texture, TextureLoader } from 'three'
 import { useLoader, useThree } from '@react-three/fiber'
-import { useLayoutEffect, useEffect } from 'react'
+import { useLayoutEffect, useEffect, useMemo } from 'react'
 
 export const IsObject = (url: any): url is Record<string, string> =>
   url === Object(url) && !Array.isArray(url) && typeof url !== 'function'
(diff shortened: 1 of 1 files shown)


drcmda:
hmm, this is bad, i forgot that giving it an object wouldn't yield a consistent cache key. all useLoader hooks are supposed to be cached, not memoized, meaning two distinct components giving it the same object are supposed to yield the same output. in a way that's possible

```jsx
const textures = { map: "foo.png", metalnessMap: "bar.png" }

function Foo() {
  const config = useTexture(textures)
  
function Bar() {
  const config = useTexture(textures)
```

to memoize would fix it per component but it would yield two results for the above components Foo and Bar, which isn't correct. i think im going to still merge to at least stop it from memory leaking.

PR #1872, opened 2024-03-14
Fix path default value in useEnvironment.tsx

dipanjanpanja6:
### Why

path & files params not working in useEnvironment function

### Checklist

- [x] Ready to be merged


drcmda:
im not sure @dipanjanpanja6 by default it's supposed to resolve px.png, nx.png et al at the /public root, so `path = ''`. it does not work if these files aren't there but that's expected.

dipanjanpanja6:
If I send the `preset` params, the path is overridden by `CUBEMAP_ROOT`. It shouldn't be overridden if I provide a path

PR #1877, opened 2024-03-16
fix: use default import for Stats.js

kkocdko:
Use default import, not namespace import.


kkocdko:
```
▲ [WARNING] Constructing "StatsImpl" will crash at run-time because it's an import namespace object, not a constructor [call-import-namespace]

    node_modules/@react-three/drei/core/Stats.js:11:44:
      11 │   const stats = useEffectfulState(() => new StatsImpl(), []);
         ╵                                             ~~~~~~~~~
```

It was introduced here https://github.com/pmndrs/drei/commit/97cf006641167dc10d2034d8fbb2d8bf9b65988f#diff-49739308cbf61365d9e1da0ff995973345abe032e3445fdf766cca5c2f19cc36R3

kkocdko:
CI failed? However I have no access permission to CI so I can't view logs? It's weird.

PR #1878, opened 2024-03-17
fix: modify nested ternary operators and any type in useTexture

Gaic4o:
### Why

Since there is a nested ternary operator and an `any` type, we refactored the nested ternary operator to make it easier to understand and maintain, and modified the `any` type.

### What

- Things that were of type any were changed to type unknown.
- We refactored the ternary operator type nested in the MappedTextureType type to make it easier to understand and maintain.
- Simplify texture processing logic in useEffect to handle various input types more uniformly.
- Because the string type is not a subtype of any[], it always returns string. So I removed the extends any[] type and modified the code to make it easier to understand.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/useTexture.tsx
+++ b/src/core/useTexture.tsx
@@ -2,14 +2,17 @@ import { Texture, TextureLoader } from 'three'
 import { useLoader, useThree } from '@react-three/fiber'
 import { useLayoutEffect, useEffect, useMemo } from 'react'
 
-export const IsObject = (url: any): url is Record<string, string> =>
+export const IsObject = (url: unknown): url is Record<string, string> =>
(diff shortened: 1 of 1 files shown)


drcmda:
thanks for all these fixes @Gaic4o really appreciate it! could you write us over at the discord to discuss administrative rights? i think fixes like these wouldn't have to site there for 2 weeks until somebody with time looks at it, it could go quicker. :)

Gaic4o:
@drcmda  
drei fix Thank you for your prompt response and the opportunity to further discuss the following issues.
Next time, I will contact you through Discord.
Thank you again for contacting us.

PR #1901, opened 2024-04-03
e2e reflection Sphere snapshot

abernier:
### Why

In the e2e scene, replace the Box with a Sphere that reflects the Env

[image](https://github.com/pmndrs/drei/pull/1901/files#diff-dd2de7a7b6866512c55a8aa736756af487e676d6edcb66cb47d49617b60727d6)

### Checklist

- [x] Ready to be merged


.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   main:
-    runs-on: ubuntu-latest
+    runs-on: macos-14
```
abernier, line 18: mac m1

PR #1909, opened 2024-04-09
fix: apply env rotation and intensity when not using background

ffdead:
### Why

`environmentRotation` and `environmentIntensity` is currently only applied when `background=true`

Fix for #1910 

### What

Always apply scene props even if background is set to false.

@drcmda please take a look as I'm unsure if there are any side effects of applying background props when not using a background.


ffdead:
Seems to have been fixed in master already while this PR was open

PR #1918, opened 2024-04-12
fix: move useVideoTexture to /web

### LINKED ISSUE #1917: All Drei imports are broken on react-native because of navigator.userAgent check in video texture
"dependencies": {
    "@react-three/drei": "^9.105.3",
    "@react-three/fiber": "^8.16.1",
    "expo": "~50.0.14",
    "expo-gl": "~13.6.0",
    "expo-status-bar": "~1.11.1",
    "react": "18.2.0",
    "react-native": "0.73.6",
    "three": "^0.163.0"
  }

### Problem description:

Trying to use any import from @react-three/drei/native in react-native causes error "TypeError: Cannot read property 'toLowerCase' of undefined, js engine: hermes" referencing the hls.js library, which was recently added to videoTexture. This is because of the check `var chromeOrFirefox = /chrome|firefox/.test(navigator.userAgent.toLowerCase());`  in hls.js - React native has no navigator defined. 

References addition of hls support here: https://github.com/pmndrs/drei/commit/47d94d72f70ad23818b993f8b316c132ff04e146

### Relevant code:
Example

```js
import { StatusBar } from 'expo-status-bar';
import { StyleSheet, Text, View } from 'react-native';

import { Suspense } from 'react'
import { Canvas } from '@react-three/fiber/native'
import { Box } from '@react-three/drei/native'


  return (
    <View style={styles.container}>
      <Text>You will get a 'tolowercase' error. </Text>
      <StatusBar style="auto" />
      <Canvas width={500} height={500}>
      <ambientLight />
      <Suspense>
        <Box />
      </Suspense>
    </Canvas>
      
    </View>
  );
}

```

### Suggested solution:

Since video texture relies on dom video elements, I don't think it would work at all in react-native. The easiest fix might be to check for navigator and do an early return if it's undefined, maybe console.warn explaining that react-native doesn't currently support video-texture. 

Eventually, it might make more sense if it's located in /web rather than /core, but that is probably a breaking change, or maybe there's some other consideration I don't know about. 

Happy to create a PR if needed :)

jorjordandan:
fixes #1917.
Moves useVideoTexture to /web. Also fix reference from faceControls.

### Why

Currently, useVideo texture imports hls.js. This import triggers a check to navigator, which causes an error because navigator doesn't exist on react native. It's not possible to dynamically import it without making it into an async function which would cause other issues.

### What

I've moved useVideoTexture from core to web, since it does not work on native anyways.

### Checklist

- [x ] Ready to be merged

Also fixed one reference to useVideoTexture from faceControls.


CodyJasonBennett:
Continued in #1919. Thanks for the help!

netgfx:
@CodyJasonBennett Thanks for this, I tried something similar but keep in mind that the `hlsRef` on the useEffect is null so I think this cleanup:
```
return () => {
    if (hlsRef.current) {
      hlsRef.current.destroy()
      hlsRef.current = null
    }
  }
 ```
 
 will never run.

CodyJasonBennett:
If `hlsRef` is ever non-null before unmount, that will run. #1919 doesn't alter cleanup behavior.

PR #1932, opened 2024-04-22
fix: PivotControls - preventDefault onContext & react only on button=0 clicks

### LINKED ISSUE #1926: PivotControls - preventDefault onContext & operate controls only with button 0

I'm using PivotControls in a scene with OrbitControls. It happens that I'm close to the control but want to pan the OrbitControls with a right click. In such cases, the browser opens a default menu. I would prefer no reaction on the right (context) click.

pivotControls - onContext

Moreover, I think it is more intuitive if the PivotControls can only be operated with mouse button 0 (left button)

### Suggested implementation:

In the single components <ScalingSphere>, <AxisArrow>, <PlaneSlider>, <AxisRotator> the event needs to cancel the default behavior:
 
```
 onContextMenu={(e) => {
          e.nativeEvent.preventDefault()
  }}
```

moreover, I would add the check for button = 0 in the onPointerDown event

```
onPointerDown={(e) => {
          if (e.button !== 0) {
            return
          }
          onPointerDown(e)
        }}
```

romankoho:
### Why

closes #1926 

### What

call preventDefault onContext and operate controls only with button 0 (left mouse button)

### Checklist

- [x] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example)) -> n.a.
- [x] Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/Example.stories.tsx)) -> n.a.
- [x] Ready to be merged

### DIFF
--- a/src/web/pivotControls/AxisArrow.tsx
+++ b/src/web/pivotControls/AxisArrow.tsx
@@ -68,6 +68,10 @@ export const AxisArrow: React.FC<{ direction: THREE.Vector3; axis: 0 | 1 | 2 }> 
 
   const onPointerDown = React.useCallback(
     (e: ThreeEvent<PointerEvent>) => {
+      if (e.button !== 0) {
+        e.stopPropagation()
+        return
+      }
       if (annotations) {
         divRef.current.innerText = `${translation.current[axis].toFixed(2)}`
         divRef.current.style.display = 'block'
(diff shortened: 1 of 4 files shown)


romankoho:
@drcmda any opinion on this MR and the linked issue?

PR #1938, opened 2024-04-25
Fix: renderOrder of <Svg> meshes

cyango:
### Why

<Svg> component was showing a weird z-fighting on camera orbit

### What

Fixed the render order of the meshes.

Discussion here: https://discord.com/channels/740090768164651008/741751532592038022/1227336884255064195

### DIFF
--- a/src/core/Svg.tsx
+++ b/src/core/Svg.tsx
@@ -39,6 +39,8 @@ export const Svg: ForwardRefComponent<SvgProps, Object3D> = /* @__PURE__ */ forw
       return () => strokeGeometries.forEach((group) => group && group.map((g) => g.dispose()))
     }, [strokeGeometries])
 
+    let renderIndex = 0
+
     return (
       <object3D ref={ref} {...props}>
         <object3D scale={[1, -1, 1]}>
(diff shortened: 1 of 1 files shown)


cyango:
After testing on more use cases, this seems to not be a good solution. Using renderOrder like that makes wierd behaviour of SVG showing on top of everything in the camera planes. Also not sure how to fix this.

src/core/Svg.tsx
```diff
@@ -48,12 +48,11 @@ export const Svg: ForwardRefComponent<SvgProps, Object3D> = /* @__PURE__ */ forw
                 path.userData?.style.fill !== undefined &&
                 path.userData.style.fill !== 'none' &&
                 SVGLoader.createShapes(path).map((shape, s) => (
-                  <mesh key={s} {...fillMeshProps}>
+                  <mesh key={s} {...fillMeshProps} renderOrder={s}>
                     <shapeGeometry args={[shape]} />
                     <meshBasicMaterial
                       color={path.userData!.style.fill}
                       opacity={path.userData!.style.fillOpacity}
-                      transparent={true}
```
CodyJasonBennett, line 56: Why remove `transparent`? It does affect grouping during sorting, but is that any relevant here?

cyango: From my experiments this was the only way i found to avoid the render order issues.

CodyJasonBennett: I'll investigate merging the geometries here. Definitely an issue with sorting, not z-fighting exactly.

RodrigoHamuy: I think the reason this PR doesn't work with `transparent` is because the `renderOrder` should be applied based on all the meshes, not in the subgroups.

Here is a working example with transparent enabled: [sandbox](https://codesandbox.io/p/sandbox/svg-bug-gvjv2z?file=%2Fsrc%2FSvg.jsx%3A47%2C3-47%2C23)

If we update `renderOrder` it should be aligned again with the threejs example.

The introduction of `renderOrder` happened in this PR: https://github.com/mrdoob/three.js/issues/26110

PR #1941, opened 2024-04-29
[feat] update for R3F v9 

krispya:
This is untested. It is a quick conversion of Drei v9 to use R3F v9 without any typing errors.

PR #1945, opened 2024-05-01
useHelper RefObject (non-mutable)

abernier:
### Why

Passing a mutable `ref={}` makes TS complains, here in the story:

image

### What

so, I've naively convert `useHelper`'s 1st param `object3D` into a simple `RefObject` (rather a `MutableRefObject`)

|before|after|
|-|-|
|image|image|

### Checklist

I'm really not sure, how "safe" this is / how required this `ref` was to be mutable...

- [ ] Ready to be merged

### DIFF
--- a/.storybook/stories/useHelper.stories.tsx
+++ b/.storybook/stories/useHelper.stories.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { type ElementRef } from 'react'
 import { BoxHelper, CameraHelper } from 'three'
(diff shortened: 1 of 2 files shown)


CodyJasonBennett:
This is actually what the React 19 types do since `useRef` can lead to confusing mutable/immutable types.

CodyJasonBennett:
Yes, this will be resolved by React 19 and the next major of both R3F/Drei. #1941

PR #1981, opened 2024-06-04
fix: add treeshakable to useProgress

mayank1513:
This will ensure that only one instance of useProgess is created. For more info please see https://github.com/react18-tools/treeshakable/ and https://treeshakable.vercel.app/

### Why

Enhancement: allow importing from the component files, e.g., `./web/Loader` to avoid importing unnecessary code.

### What

Wrapped store creation inside `treeshakable`

### Checklist

- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "suspend-react": "^0.1.3",
     "three-mesh-bvh": "^0.7.0",
     "three-stdlib": "^2.29.9",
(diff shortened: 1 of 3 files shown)


mayank1513:
Not updating `KeyboardControls` as the store is created within the component and passed to children using context. Updating this will require breaking changes.

Issue #1966, opened 2024-05-29
View.tsx does not respect "manual" attribute of Orthographic camera.


talkingtab:
Line 88. 

see also #515 
see also #1562 

What a hydra!

talkingtab:
Setting a custom frustum in an OrthographicCamera is fragile. Setting {manual: true} is required. You also have to ensure that left right, top bottom are passed the right way - not args[]. Then there are issues of updating cameras and ensuring new code does not just assume that the right thing is just update on `size`. 

I can think of  strategies: 
    if {rltb} is passed in, automatically mark it as 'manual'
    invert the sense, assume "manual" but pass in "dynamic" :-)
or maybe best, require some flag to be set like {policy='manual'} 
I'm not sure these are really good idea or suggestions, but it would be nice to avoid users running into this and coders updating based on size.

PR #1991, opened 2024-06-12
fix: 1966 View respects manual ortho camera

talkingtab:
### Why
View.tsx is changed for bug #1966. A the frustum of an orthographic camera was reset to the default position even when the camera was manual. The user's manual setting (for example to do a full-frame view) was thus overwritten.

### What
After checking that the camera is Orthographic, the code now checks if state.camera.manual is set. If not the old code is executed. Otherwise just update the projection matrix. NOT SURE THIS IS NEEDED

### Checklist

- [x] Ready to be merged 

Not sure what the right next step is.


mikaelnousiainen:
Is there something preventing this from getting merged? `View` can't really be used with use-cases that require manual orthographic camera.

PR #1999, opened 2024-06-20
sb 8.1.10

abernier:
### Checklist

- [ ] Ready to be merged


.storybook/preview.tsx
```diff
@@ -8,14 +8,6 @@ const preview: Preview = {
     layout: 'fullscreen',
   },
 
-  decorators: [
-    (Story) => (
-      <React.Suspense fallback={null}>
-        <Story />
-      </React.Suspense>
```
abernier, line 15: @CodyJasonBennett | @drcmda: I think this is not required anymore since r3f now has this

package.json
```diff
@@ -137,10 +137,10 @@
     "rollup-plugin-terser": "^7.0.2",
     "semantic-release": "^21.0.6",
     "serve": "^14.2.0",
-    "storybook": "^8.1.5",
+    "storybook": "^8.1.10",
     "three": "^0.151.0",
     "ts-node": "^10.9.1",
-    "typescript": "^4.7.4",
```
abernier, line 143: bumping TS too

.storybook/preview.tsx
CodyJasonBennett: I believe this was here since Storybook would crash prior if something suspended into its component tree.

CodyJasonBennett: Let's try removing it.

abernier (commented):
lgtm

PR #2011, opened 2024-06-30
fix(PositionalAudio): Set [loop=false] by default

satelllte:
### Why

This is kinda confusing that `three` has it set to `false`:
- https://github.com/mrdoob/three.js/blob/dev/src/audio/Audio.js#L21 (current)
- https://github.com/mrdoob/three.js/blob/d7f8eb282529c29e8395753c76ffe343c7add5b8/src/audio/Audio.js#L21 (as of 2024-06-30)

As well as the JSDoc tells the same for `PositionalAudio.loop` JSX prop:
```
(property) loop?: boolean | undefined
@default

false
```

### What

As title says

### Checklist

- [x] N/A ~Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example))~
- [x] N/A ~Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/Example.stories.tsx))~
- [x] Ready to be merged


satelllte:
Hi @CodyJasonBennett 👋🏻 
Will be possible to release this small improvement?

DennisSmolek:
I personally think you should always loop and thats why it was set this way. However it should be intentional to loop and not just done by default. 

Even the examples show it as a param. 

Implemented in v11

PR #2036, opened 2024-07-17
fix: Ensure ContactShadows respects frames=1

mjurczyk:
Demo [link](https://codesandbox.io/p/sandbox/shoe-configurator-forked-x97rp4?file=%2Fsrc%2FApp.js%3A8%2C1-9%2C53)

(Uncomment `App.tsx:L9` for current drei ContactShadows, uncomment `App.tsx:L8` for this PR ContactShadows.)

### Why

Currently `ContactShadows` renders a new shadow frame whenever anything above it updates, regardless of the value passed via `frames`. Updating state of any parent component / context causes shadows to darken / sharpen overtime, as `count` of frames is reset but the RT is not cleared. Currently there's also no performance gain of setting `frames`, shadows of static objects remain dynamic.

### What

Stores the frame count along the RT to determine whether new shadow frame is necessary. Allows fully static shadows when `frames` is passed.
Counter is reset only if the entire RT is disposed.

### Checklist

- [ ] Ready to be merged


CodyJasonBennett:
Are you still working on this?

mjurczyk:
Do we want to go with `static` / `cacheKey` in the end? I can add it in this PR, replacing the frame freeze.

PR #2038, opened 2024-07-17
fix: Fix the Readme about PivotControls

billdc:
I am using the PivotControls component. Since the positions of all objects are data-driven, I need to set autoTransform to false. I found a small error in the readme that prevents the pivot's matrix from updating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -977,7 +977,7 @@ return (
     ref={ref}
     matrix={matrix}
     autoTransform={false}
-    onDrag={({ matrix: matrix_ }) => matrix.copy(matrix_)}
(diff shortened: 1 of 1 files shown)


DennisSmolek:
Props were expanded here to also include `onDragStart` and `onDragEnd` with new types as well as args in the callback. it will be a breaking change but now they all align

```jsx
const matrix = new THREE.Matrix4()
return (
  <PivotControls
    ref={ref}
    matrix={matrix}
    autoTransform={false}
    onDrag={({ local }) => matrix.copy(local)}
  />
)
```

PR #2055, opened 2024-07-26
fix: Line accepts readonly points and vertex colors

RodrigoHamuy:
### Why

Currently this will give TS errors:

```tsx
import { Line } from "@react-three/drei";
const points: readonly [number, number][] = [[0,0], [0,1]];
const MyLine = ()=> <Line points={points} />
```

### What

Updated the types so it accepts also readonly arrays.

This doesn't mean that you need to pass immutable arrays now, just means that `<Line>` promises it won't change the receiving arrays. 

### Checklist

- [x] Ready to be merged


RodrigoHamuy:
I think the image snapshot test is failing due to an update to the test docker image.
@abernier https://github.com/pmndrs/playwright/commit/3f65c314ca74c1d315295b08d2d0047350e57de1

abernier:
ah yes, missed that, will fix

abernier:
should be fixed

PR #2066, opened 2024-08-16
fix(GizmoHelper): allow GizmoHelper to be used with camera-controls

### LINKED ISSUE #1399: Gizmo helper with camera controls won't work
Awesome work here, great library!. Now to the issue:

- `three` version: "^0.149.0"
- `@react-three/fiber` version: "^8.12.0"
- `@react-three/drei` version: "^9.65.2"
- `node` version:  v18.14.2
- `npm` (or `yarn`) version:  npm 9.5.0

### Problem description:

Gizmo helper with camera controls throws error on tween camera call

### Relevant code:

```js

...

<CameraControls ref={ref} makeDefault/>

...

<GizmoHelper
  alignment="bottom-right"
  margin={[80, 80]}
>
  <GizmoViewcube  />
</GizmoHelper>

```

### Workaround solution:

A workaround is, but if this could work out of the box would be awesome or that event have the direction of the clicked face or edge

```ts
function Gizmo() {
  const { scene, controls } = useThree();

  const tweenCamera = (position: Vector3) => {
    const point = new Spherical().setFromVector3(
      new Vector3(position.x, position.y, position.z)
    );
    // @ts-ignore
    controls.rotateTo(point.theta, point.phi, true);
    // @ts-ignore
    controls.fitToSphere(getBoundingSphere(scene), true);
  };

  return (
    <GizmoHelper margin={[80, 80]} alignment={'top-right'}>
      <GizmoViewcube
        onClick={(e) => {
          e.stopPropagation();
          if (
            e.eventObject.position.x === 0 &&
            e.eventObject.position.y === 0 &&
            e.eventObject.position.z === 0
          ) {
            tweenCamera(e.face.normal);
          } else {
            tweenCamera(e.eventObject.position);
          }
          return null;
        }}
      />
    </GizmoHelper>
  );
}
```

a3ng7n:
When using camera-controls sets target via the getTarget method, and updates the control position via the setPosition method.

### Why

Allows GizmoHelper to be used alongside CameraControls/camera-controls.
resolves #1399 

### What

Fixes two issues:
1. `CameraControls` doesn't expose a `target` property directly - so when `focusPoint` is set at animation start, get the target via `CameraControls` `getTarget` instead.
2. `CameraControls` doesn't seem to like having its camera moved directly, and will continuously attempt to reset the `mainCamera`'s position - the fix is to call `CameraControls` `setPosition` method after the `mainCamera`'s position is updated, and before it has a chance to revert it.

### Checklist

- [x] Ready to be merged

I did create a storybook entry that's just a duplicate of the GizmoHelper story below but using CameraControls to test the fix out - but I didn't think it was really value added / worthy of another tutorial entry https://github.com/pmndrs/drei/blob/f4f910f76f48e16ef73c52f2d69bd5bbf96a138e/.storybook/stories/GizmoHelper.stories.tsx#L51-L86 lmk if you think otherwise and I'll update this PR with it

### DIFF
--- a/src/core/GizmoHelper.tsx
+++ b/src/core/GizmoHelper.tsx
@@ -4,6 +4,7 @@ import { Group, Matrix4, Object3D, OrthographicCamera as OrthographicCameraImpl,
 import { OrthographicCamera } from './OrthographicCamera'
 import { OrbitControls as OrbitControlsType } from 'three-stdlib'
 import { Hud } from './Hud'
+import { CameraControls as CameraControlsType } from './CameraControls'
 
 type GizmoHelperContext = {
   tweenCamera: (direction: Vector3) => void
@@ -22,7 +23,7 @@ const [q1, q2] = [/* @__PURE__ */ new Quaternion(), /* @__PURE__ */ new Quaterni
 const target = /* @__PURE__ */ new Vector3()
 const targetPosition = /* @__PURE__ */ new Vector3()
 
-type ControlsProto = { update(): void; target: Vector3 }
+type ControlsProto = { update(delta?: number): void; target: Vector3 }
 
 export type GizmoHelperProps = JSX.IntrinsicElements['group'] & {
   alignment?:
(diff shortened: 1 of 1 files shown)


CodyJasonBennett:
Thanks.

PR #2079, opened 2024-08-29
feat: add clear and preload function to useEnvironment

### LINKED ISSUE #2076: clear and preload function for useEnvironment

Since useEnvironment is based on useLoader it should be possible to provide a preload and clearing function. 
At the moment, it would be quite a bit of trouble to preload or clear environment files because some logic is involved in selecting the correct loader based on the file extension. 

### Suggested implementation:

Similar to the other hooks based on useLoader:
```tsx
useEnvironment.clear(...)
useEnvironment.preload(...)
```

I worked out a first version in a Drei fork and it works great so far, I will open a PR soon. 
The only limitation at the moment would be that you cannot preload gainmaps, because that loader requires the renderer to be set. 
This should not be an issue as long as it is documented.

AaronClaes:
Closes #2076

### Why

Pull request for the following feature: #2076 

### What

I added a `clear` and `preload` function for `useEnvironment` based on `useLoader`. 

I also restructured parts of the hook to be able to reuse them in the `clear` and `preload` functions, to prevent duplicate code. 

The only limitation is that gainmaps cannot be preloaded, but this is stated in the updated documentation. 
The reason is that their loader requires the renderer to be set, which is not possible. An error will be thrown when attempting to preload a gainmap.

### Checklist

- [x] Documentation updated 
- [x] Ready to be merged


README.md
abernier, line 1: documentation should go to docs/ folder

PR #2089, opened 2024-09-09
fix(types): shaderIntersectFunction is defined by three-mesh-bvh

Methuselah96:
### Why

There are currently type errors when using drei with `three-mesh-bvh@^0.7.7`, since they added the `shaderIntersectFunction` type definition in https://github.com/gkjohnson/three-mesh-bvh/pull/697:

```
./node_modules/@react-three/drei/materials/MeshRefractionMaterial.d.ts(3,11): error TS2451: Cannot redeclare block-scoped variable 'shaderIntersectFunction'.
./node_modules/three-mesh-bvh/src/index.d.ts(318,14): error TS2451: Cannot redeclare block-scoped variable 'shaderIntersectFunction'.
```

### What

Update `three-mesh-bvh` and remove module augmentation

### Checklist

- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "stats-gl": "^2.0.0",
     "stats.js": "^0.17.0",
     "suspend-react": "^0.1.3",
-    "three-mesh-bvh": "^0.7.0",
+    "three-mesh-bvh": "^0.7.7",
     "three-stdlib": "^2.29.9",
     "troika-three-text": "^0.49.0",
     "tunnel-rat": "^0.1.2",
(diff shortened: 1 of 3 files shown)


Methuselah96:
Hmm, seems like `three-mesh-bvh@0.7.7` also imports `BatchedMesh` which is not available in all versions of `three` that drei supports. I guess the `three-mesh-bvh` could be pinned back to `three-mesh-bvh@0.7.6`, and just do some casting instead of module augmentation, but seems like you'll want to be able to upgrade at some point...

Methuselah96:
The issue with importing `BatchedMesh` was fixed in `three-mesh-bvh@0.7.8`, this PR should be safe to merge now

PR #2095, opened 2024-09-17
Use `TTFLoader` in `useFont` for non-JSON files

### LINKED ISSUE #2094: Allow `Text3D` to use `TTFLoader`

`three-stdlib` provides a `TTFLoader` that automatically does the same type of conversion that happens when using the Facetype.js website. `Text3D` should support both loaders, depending on the content-type returned from the fetched URL. This way, especially during development where speed of loading is not a concern, developers don't have to manually convert fonts using the Facetype.js website, which slows down experimentation with different fonts / variants etc.

### Suggested implementation:

I will submit a PR soon.

mz8i:
### Why

Resolves #2094 

### What

I modified `useFont` to rely not only on `FontLoader`, but also `TTFLoader`, which can transform an `ArrayBuffer` to Typeface.json object. If a string is passed to the `Text3D.font` prop, or to `useFont` directly, and the content type of the fetched response is not `application/json`, then `TTFLoader` is used first.

NOTE: the Storybook story is draft for now in that I didn't want to add any .ttf file to the `.storybook/public/fonts` folder, so the story points to a `Placeholder.ttf` and will error. I welcome suggestions on what font file should be added to the tests when considering licensing.

Also NOTE: the current implementation doesn't allow for setting properties on the loaders. `FontLoader` doesn't really have any properties apart from the ones inherited from base `Loader`, and the original code also didn't give an option to access these. However, `TTFLoader` does have one important property: `reversed` (which is the same as the "reverse direction" checkbox on the Facetype.js website and helps deal with issues with rendering of characters such as "8", "b" etc). I haven't figured out yet what would be a nice way to allow for setting this property. Perhaps something like:
```jsx
<Text3D font="/fonts/Example.ttf" loaderArgs={ ttf: {reversed: true } } >
Text
</Text3D>
```

I also updated the documentation/examples, but any suggestions for improvements there are very welcome!

### Checklist

- [ ] Ready to be merged


mz8i:
Hi @drcmda - hope you don't mind me tagging directly but I can see you did the initial development on `useFont`. This PR is set to draft as I wasn't sure how to approach the updating of the storybook story (with regards to what font file to include in the storybook) - but otherwise I'd be grateful if someone from the maintainers could have a look at this and review!

mz8i:
I've marked this as ready for review to hopefully get some attention from the maintainers - see comment above about Storybook needing an update of the TTF font file to load.

abernier (commented):
added a ttf file for the story

Methuselah96:
Is this a breaking change? I'm loading using `Text3D` with `*.jsonfont` files where the content type is `application/octet-stream`, so now my app is crashing. I don't really have control over what the content type of the font is. Should there maybe be an option parameter on `Text3D`/`useFont` to control which loader to use instead?

BenBeckTA:
Would it also be possible to update the `Text` interface to to allow `font` to be `Parameters<typeof useFont>[0]` rather than `string`? It seems that `useFont` can only be used with `Text3D` and not `Text`?

abernier:
> Would it also be possible to update the `Text` interface to to allow `font` to be `Parameters<typeof useFont>[0]` rather than `string`? It seems that `useFont` can only be used with `Text3D` and not `Text`?

in a separate PR yes, but first we need to solve the issue raised by @Methuselah96 

@mz8i what about @Methuselah96 's suggestion about explicit loader prop?

mz8i:
All good points - perhaps this needs a bit more design and testing to ensure a good interface and no breaking changes. Also see my note above about the `reversed` argument to `TTFLoader` not being accessible. This is my first contribution here so feel free to suggest better approaches to designing this!

Re immediate fixes - yes, that sounds like a good solution @abernier to add an explicit prop to enable the new behavior, while the default will be to stick to the old one. The tricky bit is that it's not just about switching between one loader or the other. The options are actually:
- only use `FontLoader`, or 
- first use `TTFLoader`, and then `FontLoader`.

So maybe something like `ttfLoader={true}` prop to explicitly enable this additional first step?

abernier:
ok so I'm going to revert the PR, and @mz8i you'll then be able to open a new one with the cherry-picked commit applied to

We'll can then discuss the implementation, is it ok for you?

mz8i:
Yes, perfect, thanks!

abernier:
it's reverted and included in v9.117.3

@mz8i please reopen a new PR, cherry-picking https://github.com/pmndrs/drei/commit/61d3bc52b806da56c9d8df8bd5600fe5570b1f7a

Issue #2097, opened 2024-09-18
feat: prevent overlapping on DragControls component


niccolofanton:
# WHY
When using multiple DragControls, if the user drags one and others are overlapping, all of them will move

https://github.com/user-attachments/assets/7e14fd6a-a28c-4343-b386-5530d8d4a29a

PR #2098, opened 2024-09-18
Added preventOverlap flag to DragControls

niccolofanton:
# WHY
Solves https://github.com/pmndrs/drei/issues/2097

# WHAT
This PR introduces a new property `preventOverlap` on the DragControls component that, if set to `true`, prevents the Click event from propagating to other DragControls

https://github.com/user-attachments/assets/1febf41a-8914-4319-803e-dd4ca49e50cb


abernier:
wouldn't be better to pass the event object to the onDrag prop callback?

PR #2109, opened 2024-09-24
refactor: use vite.config.ts as the sole config file

hichemfantar:
### Why

the extra file seemed redundant

### What

consolidate into one config file

### Checklist

- [ ] Ready to be merged


hichemfantar:
marked as not ready because i'm not sure why things were setup like this.

perhaps @abernier can chime in?

PR #2115, opened 2024-10-04
useVariants hook to use KHR Material Variants

DennisSmolek:
image

[Live Sandbox](https://codesandbox.io/p/sandbox/variants-demo-tqnq4d)

### Why

[The KHR Material Variants](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_variants) GLTF Extension allows you to set multiple materials on multiple meshes within a single GLTF instance. This allows you to create many different versions of objects inside  a single gltf file and set a global state object to switch between them.

This is supported within threeJS core: [Here](https://threejs.org/examples/?q=var#webgl_loader_gltf_variants)

### What

This hook takes the returned GLTF object from the standard loader or loaders like `useGLTF`
```jsx
const gltf = useGLTF('model.glb');
const { activeVariant, setVariant, variants, materials } = useVariants(gltf);
```

`activeVariant` is the string name of the current variant
`setVariant` takes the string name input of the variant object
`variants` is an array of of variants that look like `{ name: 'variantName' }` ( NOTE: I was originally going to simplify this to just be the variant names, but talking with Don it would be possible to add data to variant objects that could be useful. Leaving it this way makes that data accessible)
`materials` This is less commonly used but is the actual materials array from the GLTF. apps like GLTFJSX actually struggle with gltf material lists and can accidentally merge them or mess them up. Because we are already deep into the GLTF I added this to help me pull them out and read them if needed. It will probably be ignored by most.

You can set the variant using the `setVariant` method or as an input to the hook itself.
This makes things convienent when the switching logic happens outside the component.

Using `setVariant()`

```jsx
const gltf = useGLTF('model.glb')
const { setVariant } = useVariants(gltf)

useEffect(() => {
  setVariant(props.variant)
}, [props.variant])
```

Using direct hook input:

```jsx
const gltf = useGLTF('model.glb')
useVariants(gltf, props.variant)
```

### Checklist

- [x ] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example))
- [x ] Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/Example.stories.tsx))
- [x ] Ready to be merged

### DIFF
--- a/.storybook/stories/useVariants.stories.tsx
+++ b/.storybook/stories/useVariants.stories.tsx
@@ -1,0 +1,55 @@
+import * as React from 'react'
+import { Vector3 } from 'three'
+import type { GLTF } from 'three-stdlib'
+import type { Meta, StoryObj } from '@storybook/react'
+
+import { Setup } from '../Setup'
+
+import { useAnimations, useGLTF, useMatcapTexture, useVariants } from '../../src'
+
+export default {
+  title: 'Abstractions/useVariants',
+  component: UseVariantsScene,
+  decorators: [
+    (Story) => (
+      <Setup cameraPosition={new Vector3(0, 0, 2)}>
+        <Story />
+      </Setup>
+    ),
+  ],
+} satisfies Meta<typeof UseVariantsScene>
+
+interface UseVariantsSceneProps {
+  inVariant: string
+}
+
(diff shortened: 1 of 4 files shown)


DennisSmolek:
Implemented in v11

PR #2133, opened 2024-10-22
Allow clipping planes to be used with Outlines

Archimagus:
### Why 

resolves#2132

### What

Added clipping planes to the Outlines object.

### Checklist

- [ X] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example))
- [X] Ready to be merged


Archimagus:
Is this missing anything for it to be merged?  The process seems a bit different than the last PR I made.

PR #2139, opened 2024-10-27
Fix any return on useSpriteLoader

leo-petrucci:
### Why

When using the `useSpriteLoader` hook it returns an object with a `spriteObj` property which is what we need to use in the `SpriteAnimator` component. 

The hook itself however explicitly returns `any`, which could be confusing for someone who has never used the hook before.

### What

Removed the explicit `any` return.

### Checklist

- [x] Ready to be merged

PR #2171, opened 2024-10-30
`SpriteAnimator`/`useSpriteLoader` types

netgfx:
### Why

Better typescript support for inner variables and functions for SpriteAnimator

### What

Added more robust and detailed types for SpriteAnimator

### Checklist

- [x ] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example))
- [x ] Ready to be merged


docs/misc/sprite-animator.mdx
```diff
@@ -15,7 +15,71 @@ sourcecode: src/core/SpriteAnimator.tsx
 
 ```tsx
 
-/** Common properties that can be safely applied to both sprite and instance components */
+/**
+ * Source size information for a sprite frame
+ */
+interface SourceSize {
```
abernier, line 21: Not prescriptive, but we generally prefer using `type` over `interface` in the project

docs/misc/sprite-animator.mdx
```diff
@@ -43,7 +107,10 @@ type CommonMeshProps = {
   matrixAutoUpdate?: boolean
 }
 
-type SpriteAnimatorProps = {
+/**
+ * Props for the SpriteAnimator component
+ */
+export type SpriteAnimatorProps = {
   /** The start frame of the animation */
```
abernier, line 114: please also add this type of "jsdoc comment" in the component, so Storybook will add it to the props table:
image

docs/misc/sprite-animator.mdx
netgfx: done in latest commit

docs/misc/sprite-animator.mdx
netgfx: added props comments

src/core/SpriteAnimator.tsx
```diff
@@ -1,72 +1,176 @@
...
+  /** Whether or not the animation should auto-start when all assets are loaded */
   autoPlay?: boolean
+  /** The animation names of the spritesheet (if the spritesheet -with JSON- contains more animation sequences) */
   animationNames?: Array<string>
-  onStart?: Function
-  onEnd?: Function
-  onLoopEnd?: Function
-  onFrame?: Function
+  /** Event callback when the animation starts or restarts
+   * @param data Animation state data containing currentFrameName and currentFrame
```
abernier, line 83: not sure `@param` is really useful is since it is already typed, ie: `: AnimationEventData`

src/core/SpriteAnimator.tsx
```diff
@@ -1,72 +1,176 @@
...
 
+// Type guard for SpriteData
+function isSpriteData(data: SpriteData | null): data is SpriteData {
+  return data !== null && 'meta' in data && 'frames' in data
+}
+
 const geometry = new THREE.PlaneGeometry(1, 1)
 
-export const SpriteAnimator: React.FC<SpriteAnimatorProps> = /* @__PURE__ */ React.forwardRef(
+export const SpriteAnimator: React.ForwardRefExoticComponent<SpriteAnimatorProps & React.RefAttributes<THREE.Group>> = React.forwardRef(
```
abernier, line 173: type casting is not the greatest, you can let TS infer using `forwardRef` generics `<Ref, Props>`:

```ts
const SpriteAnimator = forwardRef<Ref, Props>((props, fref) => {...})
```

see: https://github.com/pmndrs/drei/blob/master/src/core/Example.tsx#L28

src/core/SpriteAnimator.tsx
```diff
@@ -1,72 +1,176 @@
...
+   */
   offset?: number
+  /** Allows the sprite animation to start from the end towards the start */
   playBackwards?: boolean
+  /** Allows the animation to be paused after it ended so it can be restarted on demand via autoPlay */
   resetOnEnd?: boolean
+  /** Array of Vector3-like positions for creating multiple instances of the sprite */
+  instanceItems?: (THREE.Vector3 | [number, number, number])[]
+  /** The maximum number of instances to render (for buffer size calculation)
+   * @default 1
```
abernier, line 125: to avoid this `@default` manual jsdoc annotation, you could rely on TS which will infer it as long as you really use default value in the props:

```ts
const SpriteAnimator = ...
(
    {
      startFrame,
      ...
      maxItems = 1, // default value !
      ...
    },
    fref
  ) => {
```

if using a real default value like this, Storybook will have it also in its doc

abernier: image

abernier: + same for other props with default values

src/core/SpriteAnimator.tsx
```diff
@@ -102,23 +206,23 @@ export const SpriteAnimator: React.FC<SpriteAnimatorProps> = /* @__PURE__ */ Rea
     },
     fref
   ) => {
-    const ref = React.useRef<any>()
-    const spriteData = React.useRef<any>(null)
-    const matRef = React.useRef<any>()
-    const spriteRef = React.useRef<any>()
+    const ref = React.useRef<THREE.Group>(new THREE.Group())
```
abernier, line 209:
```suggestion
    const ref = React.useRef(new THREE.Group())
```
since you pass `new THREE.Group()` TS will infer the returned ref type

src/core/SpriteAnimator.tsx
netgfx: done in latest commit

src/core/SpriteAnimator.tsx
netgfx: I removed these where applicable

src/core/SpriteAnimator.tsx
```diff
@@ -1,72 +1,163 @@
 import * as React from 'react'
```
abernier, line 1: as in `src/core/Example.tsx`, it is good practice to enable exhaustive-deps hook:
```suggestion
/* eslint react-hooks/exhaustive-deps: 1 */
import * as React from 'react'
```

src/core/SpriteAnimator.tsx
```diff
@@ -472,7 +571,7 @@ export const SpriteAnimator: React.FC<SpriteAnimatorProps> = /* @__PURE__ */ Rea
     })
 
     // utils
-    const getFirstItem = (param: any): any => {
+    const getFirstItem = (param: object | null | []): any => {
```
abernier, line 574: why type-casting the returned value with `: any`?

netgfx: fixed on latest commit

src/core/SpriteAnimator.tsx
netgfx: added the instruction

src/core/SpriteAnimator.tsx
```diff
@@ -242,13 +333,16 @@ export const SpriteAnimator: React.FC<SpriteAnimatorProps> = /* @__PURE__ */ Rea
         }
       } else {
         spriteData.current = frameData
-        totalFrames.current = spriteData.current.frames.length
-
+        if (spriteData.current) {
+          totalFrames.current = spriteData.current.frames.length as number
```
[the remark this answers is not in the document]
netgfx: it needed some refactoring but I think it is ok now

abernier:
I can see your `src/core/SpriteAnimator.tsx` is not prettified, did you disable pre-commit hook?

netgfx:
> I can see your `src/core/SpriteAnimator.tsx` is not prettified, did you disable pre-commit hook?

I haven't disabled anything (at least not intentionally), also since there is the `.prettierrc` on the repo I think my vscode prettier picks it up
On vscode I get:
```
["INFO" - 5:49:54 PM] Detected local configuration (i.e. .prettierrc or .editorconfig), VS Code configuration will not be used
["INFO" - 5:49:54 PM] Prettier Options:
{
  "filepath": "c:\\xampp\\htdocs\\drei\\drei\\src\\core\\SpriteAnimator.tsx",
  "parser": "typescript",
  "semi": false,
  "trailingComma": "es5",
  "singleQuote": true,
  "tabWidth": 2,
  "printWidth": 120,
  "useTabs": false,
  "endOfLine": "auto"
}
["INFO" - 5:49:54 PM] Formatting completed in 449ms.
```

src/core/SpriteAnimator.tsx
netgfx: Yes now I see them (I had disabled eslint on this workspace for some reason), I tried to quickly fix them but it isn't as straightforward, some dependencies if added need quite the rework and extensive testing because they will or might cause unnecessary re-renders. I'll see if I can work on them during the weekend

abernier:
I removed some more type-assertions, but be aware that using `any` of `as XXX` of `: XXX` disables TS

abernier:
same apply https://github.com/pmndrs/drei/blame/master/src/core/useSpriteLoader.tsx where so many type-assertions are used

docs/misc/sprite-animator.mdx
```diff
@@ -15,7 +15,71 @@ sourcecode: src/core/SpriteAnimator.tsx
...
+  /** Width of the source frame */
+  w: number
+  /** Height of the source frame */
+  h: number
+}
+
+/**
+ * Data structure for a single frame in the sprite sheet
+ */
+interface FrameData {
```
abernier, line 31: I don't think users have to know about all those type, only the Props one

netgfx: cleaned them up and reduced the comments to the bare minimum on the documentation

netgfx:
> same apply https://github.com/pmndrs/drei/blame/master/src/core/useSpriteLoader.tsx where so many type-assertions are used

I'll open another PR for `useSpriteLoader`

src/core/SpriteAnimator.tsx
```diff
@@ -1,98 +1,177 @@
...
-      asSprite,
+      asSprite = false,
       offset,
       playBackwards,
       resetOnEnd,
-      maxItems,
+      maxItems = 1,
       instanceItems,
       spriteDataset,
       canvasRenderingContext2DSettings,
```
abernier, line 177: you should set as many default as possible since it will reduce the complexity of handling with `undefined` values into the code

src/core/SpriteAnimator.tsx
```diff
@@ -1,98 +1,177 @@
...
 
+type FrameData = {
+  frame: {
+    x: number
+    y: number
+  }
+  sourceSize: Size
+}
+
+type Frames = Record<string, FrameData[]> | FrameData[]
```
abernier, line 23: I think the 2nd form `FrameData[]` does not bring anything, but brings complexity into the code to deal with this alternate format

i think you should remove it

src/core/SpriteAnimator.tsx
```diff
@@ -127,20 +209,144 @@ export const SpriteAnimator: React.FC<SpriteAnimatorProps> = /* @__PURE__ */ Rea
...
+    }, [flipOffset, frameName, onStart])
+
+    // Move these into refs to prevent them from triggering re-renders
+    const loadAssetsRef = React.useRef((imageUrl: string, dataUrl: string) => {
+      loadJsonAndTexture(imageUrl, dataUrl)
+    })
+
+    const parseDataRef = React.useRef((texture: THREE.Texture, data: SpriteData) => {
+      parseSpriteDataLite(texture, data)
+    })
```
abernier, line 339: putting those 2 funcs into refs is really weird to me.

netgfx: I did so because if added as dependencies (even if they are useCallback) they still cause un-intentional re-renders, one solution was to omit them from the dependency array but that won't pass from eslint, this was the other way, perhaps there is another way more common to deal with this scenario?

src/core/SpriteAnimator.tsx
netgfx: The JSON format that this component supports could be either an array (JSON Array) or an object (JSON Hash), this is how some software export them (TexturePacker for example) so we need to support  both options, the code deals with this so the end-user doesn't have to pick between the two

abernier: makes sense then ;)

src/core/SpriteAnimator.tsx
abernier: why those 2 funcs would change "unintentional"ly?

netgfx: not sure, I'll have to dig deeper to see which of their own dependencies trigger the change

abernier: I think https://github.com/pmndrs/drei/pull/2139 should be fixed with this PR, as both are coupled

netgfx: Added type fixing on useSpriteLoader and amended the documentation examples to showcase the replaced properties like scale with meshProps

abernier: ok but still using your wrapped version of your 2 funcs?

src/core/useSpriteLoader.tsx
```diff
@@ -1,19 +1,66 @@
...
+
+type SpriteData = {
+  frames: Record<string, FrameData[]> | FrameData[]
+  meta: MetaData
+}
+
+type SpriteMetaDimension = {
+  row: number
+  col: number
+}
```
abernier, line 52: can't reuse any type from SpriteAnimator?

src/core/useSpriteLoader.tsx
```diff
@@ -1,19 +1,66 @@
...
     return param[0]
   } else if (typeof param === 'object' && param !== null) {
     const keys = Object.keys(param)
 
     return param[keys[0]][0]
   } else {
-    return { w: 0, h: 0 }
+    return { w: 0, h: 0, sourceSize: { w: 0, h: 0 } }
   }
 }
```
abernier, line 65: can't reuse getFirstFrame from SpriteAnimator?

src/core/SpriteAnimator.tsx
netgfx: not using them anymore, I removed them now (not sure why eslint didn't warn about unused vars)

src/core/useSpriteLoader.tsx
netgfx: should I export from spriteanimator? I thought it would be a good idea to have them not rely on each other, after all useSpriteLoader can be used as a standalone hook to load up textures from a sprite, not necessarily used with SpriteAnimator. 
But I do understand that we have duplicate types now and that could be an issue, perhaps if SpriteAnimator could pull types from useSpriteLoader since the latter is more generic? What do you think?

netgfx: agreed, I'll see to it, cheers!

netgfx: I've added these changes, SpriteAnimator pulls types when possible from useSpriteLoader

netgfx:
@abernier I have made all the amendments you pointed out, is there anything else needed here?

src/core/SpriteAnimator.tsx
```diff
@@ -102,23 +155,22 @@ export const SpriteAnimator: React.FC<SpriteAnimatorProps> = /* @__PURE__ */ Rea
...
-    const [aspect, setAspect] = React.useState<number[] | null>([1, 1, 1])
+    const currentFrame = React.useRef(startFrame)
+    const currentFrameName = React.useRef(frameName)
+    const fpsInterval = fps > 0 ? 1000 / fps : 0
+    const [spriteTexture, setSpriteTexture] = React.useState(new THREE.Texture())
+    const totalFrames = React.useRef(0)
+    const [aspect, setAspect] = React.useState(new THREE.Vector3(1, 1, 1))
     const flipOffset = flipX ? -1 : 1
-    const [displayAsSprite, setDisplayAsSprite] = React.useState(asSprite ?? true)
+    const [displayAsSprite, setDisplayAsSprite] = React.useState(asSprite)
```
abernier, line 170: you can remove that state, which is just syncing its value with `asSprite`: instead you can just use `asSprite` value direclty

src/core/SpriteAnimator.tsx
```diff
@@ -127,20 +179,139 @@ export const SpriteAnimator: React.FC<SpriteAnimatorProps> = /* @__PURE__ */ Rea
...
+                size: { w: width, h: height },
+                scale: '1',
+                rows: 0,
+                columns: 0,
+                frameWidth: 0,
+                frameHeight: 0,
+              },
+            }
+
+            spriteData.current = data
```
abernier, line 224: it's like this instruction is overwriting the previous:
```tsx
            spriteData.current = {
              frames: [],
              meta: {
                version: '1.0',
                size: { w: width, h: height },
                scale: '1',
                rows: 0,
                columns: 0,
                frameWidth: 0,
                frameHeight: 0,
              },
            }
```

src/core/useSpriteLoader.tsx
```diff
@@ -26,355 +75,411 @@ export const checkIfFrameIsEmpty = (frameData: Uint8ClampedArray) => {
...
-    for (const frame of frames) {
-      const { w, h } = frame.frame
-      const area = w * h
-      if (area === largestFrame?.area) {
-        frame.scaleRatio = 1
+      new Promise<THREE.Texture>((resolve) => {
+        textureLoader.load(validUrl, resolve)
+      }).then((texture) => {
+        parseSpriteData(null, texture)
+      })
```
abernier, line 449: you don't use the promise object you create, so why not just a callback:

```suggestion
      textureLoader.load(validUrl, (tex) => {parseSpriteData(null, tex)})
```

netgfx: fixed in the latest commit

abernier:
@netgfx there was a problem with eslint: `react-hooks/exhaustive-deps` was temporarily not applied

can you merge with master? thank you

abernier:
@netgfx there was an issue with eslint where `react-hooks/exhaustive-deps` was temporarily not applied

Can you merge with master?

netgfx:
> @netgfx there was an issue with eslint v9(recently upgraded) where `react-hooks/exhaustive-deps` was temporarily not applied
> 
> Can you merge with master?

merged and fixed 2 issues that came up.

src/core/useSpriteLoader.tsx
```diff
@@ -26,355 +75,409 @@ export const checkIfFrameIsEmpty = (frameData: Uint8ClampedArray) => {
...
+  /** The URL of the sprite sheet. */
+  input: Url | null,
+  /** The JSON data of the sprite sheet. */
   json?: string | null,
+  /** The names of the animations in the sprite sheet. */
   animationNames?: string[] | null,
+  /** The number of frames in the sprite sheet. */
   numberOfFrames?: number | null,
+  /** A callback that is called when the sprite sheet is loaded. */
   onLoad?: (texture: Texture, textureData?: any) => void,
```
abernier, line 88: can you remove any here?

src/core/useSpriteLoader.tsx
```diff
@@ -26,355 +75,409 @@ export const checkIfFrameIsEmpty = (frameData: Uint8ClampedArray) => {
...
+        result[key] = processFrameArray(frames[key])
       }
+      return result
+    }
+  }, [])
 
-      canvas.width = texture.image.width
-      canvas.height = texture.image.height
+  // for frame based JSON Hash sprite data
+  const parseFrames = React.useCallback((): any => {
```
abernier, line 230: can you remove the any returned type?

src/core/useSpriteLoader.tsx
```diff
@@ -26,355 +75,409 @@ export const checkIfFrameIsEmpty = (frameData: Uint8ClampedArray) => {
...
+    const sprites: Record<string, any> = {}
```
abernier, line 231: also here

netgfx: done, it needed some typescript gymnastics, but it's more streamlined now

src/core/SpriteAnimator.tsx
```diff
@@ -1,99 +1,152 @@
...
-  current: number | undefined
-  /** The 0-1 normalised and damped current goal position along curve */
-  offset: number | undefined
-  hasEnded: boolean | undefined
-  ref: React.MutableRefObject<any> | undefined | null | ((instance: any) => void)
+  current?: number
+  offset?: number
+  imageUrl?: string
+  hasEnded: boolean
+  ref: React.RefObject<THREE.Group> | null | ((instance: THREE.Group) => void)
```
abernier, line 104:
```suggestion
  ref: React.Ref<THREE.Group>
```

src/core/useSpriteLoader.tsx
abernier: i did in https://github.com/pmndrs/drei/pull/2171/commits/a15183296dc433425992b19dfc33fcc0c4aa89a7

abernier (commented):
in https://github.com/pmndrs/drei/pull/2171/commits/c097c0df4367ba41127251631917a2cb356cb345 to reduce dependencies, I've moved out utilities functions out of the hook

abernier (commented):
in https://github.com/pmndrs/drei/pull/2171/commits/9deb57f8309721004979ef86a0f921bee6028576 I've just simplified a lot useSpriteLoader

abernier:
imo, SpriteAnimator could also be simplified a lot

netgfx:
> then in [9deb57f](https://github.com/pmndrs/drei/commit/9deb57f8309721004979ef86a0f921bee6028576) I've just simplified a lot useSpriteLoader

I'll check it out and run some tests, I think one of the latest changes on useSpriteLoader is causing re-renders

netgfx:
> @netgfx are we good to merge?

I'll perform some tests and let you know as soon as possible

netgfx:
@abernier I think we are good to go

PR #1902, opened 2024-04-05
fix: correct calcultation for trackRect is offscreen

ddrdragon:
### Why

When `<View />` is non-fullscreen, the calculation of isOffscreen doesn't consider the position of left and top. 

[...]

PR #2191, opened 2024-11-12
fix(view): properly calculate offscreen state

0xsven:
### Why

The `View` component is supposed to work in a `Canvas` that is not fullscreen. 

Quote from the [docs](https://drei.docs.pmnd.rs/portals/view#view):
> Note that @react-three/fiber newer than ^8.1.0 is required for View to work correctly if the canvas/react three fiber root is not fullscreen.

I noticed that it didn't work properly. It still checked if the View is offscreen, when it should really check if it's outside the canvas.

Here is a codesandbox: https://codesandbox.io/p/sandbox/view-skissor-forked-w2tds3?file=%2Fsrc%2Fstyles.css%3A38%2C1 
Notice how the View (red border) is not always running when inside the canvas (blue border).

### What

I have kept the old logic for legacy CanvasSize, and rewrote the logic for r3f < 8.1.0. 

### Checklist

- [x] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/README.md#example)) --> not needed
- [ ] Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/Example.stories.tsx)) --> Haven't found a View story. I could create one, if that's needed...
- [x] Ready to be merged

### DIFF
--- a/src/web/View.tsx
+++ b/src/web/View.tsx
@@ -67,14 +67,20 @@ export type ViewProps = {
 
 function computeContainerPosition(canvasSize: LegacyCanvasSize | CanvasSize, trackRect: DOMRect) {
   const { right, top, left: trackLeft, bottom: trackBottom, width, height } = trackRect
-  const isOffscreen = trackRect.bottom < 0 || top > canvasSize.height || right < 0 || trackRect.left > canvasSize.width
+
   if (isNonLegacyCanvasSize(canvasSize)) {
+    const isOffscreen =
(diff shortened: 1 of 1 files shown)


abernier:
for reference https://discord.com/channels/740090768164651008/741751532592038022/1306519825560109108

DennisSmolek:
Duplicate of #1902 and implemented in v11

PR #2216, opened 2024-11-18
fix: `<StatsGl />` props

devloop01:
### Why
Provide proper typing for stats-gl constructor parameters while maintaining existing custom props

### What
Updated Props type to include stats-gl constructor parameters

### Checklist

- [x] Ready to be merged


DennisSmolek:
This might be old but we're handling the constructor and props pretty well in v11

PR #2217, opened 2024-11-18
refactor: add option to set default shader material to transparent in…

Abhishek-Singh02:
… Html component

### Checklist

- [ ] Ready to be merged


DennisSmolek:
With the new legacy/webgpu split and it not needing a material 90% of the time I've added a HtmlMaterial in both legacy and webgpu paths. you can add the transparency there although its not needed

PR #2225, opened 2024-11-20
pkg-size action

abernier:
### Checklist

- [ ] Ready to be merged

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,9 @@ jobs:
           cache: 'yarn'
       - run: yarn install
       - run: yarn build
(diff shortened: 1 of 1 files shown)


DennisSmolek:
A slightly different version of this (as its own action, different config) was added to v11: https://github.com/pmndrs/drei/pull/2679

abernier:
what v11 does not have? 😅

DennisSmolek:
> what v11 does not have? 😅

A lot unfortunately.. I'm trying to get all open PR's and issues into the v11 setup. That doesn't mean 100% and working but on that pattern. 
Any PR or issue I'm treating as a todo, I'm almost there!

PR #2250, opened 2024-12-04
Make looking for parent not crash

i-am-the-slime:
### Why

I have trouble with this [codesandbox](https://codesandbox.io/p/sandbox/scene-transitions-reveal-forked-nnj7py?file=%2Fpackage.json%3A19%2C8):

It uses the latest versions of react and fiber. Every frame it fails to find a parent and logs an error. I suppose this is because [this logic](https://github.com/pmndrs/drei/blob/master/src/core/RenderTexture.tsx#L60-L66) doesn't work anymore. 

### What

I'm proposing a safer variant here checking if __r3f exists, but I'm not sure if internals have changed and there's a better way.

### Checklist

- [x] Ready to be merged


i-am-the-slime:
Yeah, this didn't help, sorry.

PR #2252, opened 2024-12-05
chore(deps): Use `useId()` instead of `uuid` module

nikeee:
### Why
The `crypto` global is available in all supported Node.js versions as well as [in all major browsers](https://caniuse.com/mdn-api_crypto_randomuuid). The uuid module is rendered unnecessary for uuidv4.

`@react-three/drei` is currently downloaded about 253000 times per week, so about 13409000 times per year. `uuid@9` is about 3.7 kB in size (gzipped+minified). Removing this dependency saves about 50GB traffic per year.

### Checklist
- [x] Ready to be merged


nikeee:
Note: the global `crypto` object is `undefined` in non-secure contexts. If it is desired that this library also in [non-secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts), just close this PR without any comment.

abernier:
what about https://react.dev/reference/react/useId

nikeee:
I think that would be even better in the context of react. Should I change this PR?

nikeee:
Done

PR #2268, opened 2024-12-12
fix: calculate instanced mesh bounds after update

davidboy:
### Background

We are using the [&lt;Merged /&gt;](https://drei.docs.pmnd.rs/performances/merged) component for a visual that has a number of different possible shapes. A single `<Merged />` is rendered to provide instanced meshes for each of these shapes.

We were encountering a bug where, after changing from one shape to another, a visual would begin disappearing at certain camera zoom levels. I suspected this was caused by frustum culling. See the video below for an example. I have not yet been able to reproduce this behavior in a code sandbox.

https://github.com/user-attachments/assets/80c0d81f-ed95-408e-9a6a-139852e2ec63

### What

The ["How to update things"](https://threejs.org/docs/index.html#manual/en/introduction/How-to-update-things) article from the three.js documentation, when talking about `InstancedMesh`, states:

> Certain library features like view frustum culling or ray casting rely on up-to-date bounding volumes (bounding sphere and bounding box). Because of the way how InstancedMesh works, the class has its own [boundingBox](https://threejs.org/docs/index.html#api/en/objects/InstancedMesh.boundingBox) and [boundingSphere](https://threejs.org/docs/index.html#api/en/objects/InstancedMesh.boundingSphere) properties that supersede the bounding volumes on geometry level... Similar to geometries you have to recompute the bounding box and sphere whenever you change the underlying data.

My interpretation of the above paragraph is that we need to re-calculate the bounding sphere of an instanced mesh whenever the instance matrix is changed. This pull request modifies `<Instances />` to check whether an instanced mesh has a bounding sphere or bounding box when updating its instance matrix, and if it does, recalculates them.

This change fixes the issue that we were experiencing, and makes sense to me based on my reading of the documentation, so I have not continued to investigate further. If you suspect I am misunderstanding the root cause of the bug or this proposed fix, or if it would be helpful for me to continue attempting to reproduce the issue in a code sandbox, please let me know. Thanks.

- [x] Ready to be merged


DennisSmolek:
This has the right idea but this would trigger the bounds check on every frame. I've done a version of this but only when an update moves something and this would be needed

PR #2276, opened 2024-12-18
fix DREI to work with latest three-sdlib

RodrigoHamuy:
### Why

After releasing https://github.com/pmndrs/three-stdlib/pull/392 we need to do this small change in DREI to fix some TS issues.

### What

`OrbitControlsChangeEvent` is now more specific.

### Checklist

- [ ] Ready to be merged

Keeping it in draft until https://github.com/pmndrs/three-stdlib/pull/392 is merged.

### DIFF
--- a/src/core/OrbitControls.tsx
+++ b/src/core/OrbitControls.tsx
@@ -1,12 +1,14 @@
 import { EventManager, ReactThreeFiber, useFrame, useThree } from '@react-three/fiber'
 import * as React from 'react'
-import type { Camera, Event, OrthographicCamera, PerspectiveCamera } from 'three'
+import { Camera, Event, OrthographicCamera, PerspectiveCamera } from 'three'
 import { OrbitControls as OrbitControlsImpl } from 'three-stdlib'
 import { ForwardRefComponent } from '../helpers/ts-utils'
 
-export type OrbitControlsChangeEvent = Event & {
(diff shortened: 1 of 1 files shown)


CodyJasonBennett:
Picked this into #2283.

RodrigoHamuy:
@CodyJasonBennett thanks ! 
This one is in draft because it needs this first https://github.com/pmndrs/three-stdlib/pull/392
After that you won't need to ts ignore and things should work as usual. 
Just to understand where we are going with this issue, base on #2283 code comments, will it be preferable instead to loosen back the types directly in https://github.com/three-types/three-ts-types/pull/1398 ? Just asking to close the alternative solutions and avoid noise.
Thanks!

CodyJasonBennett:
No, it would not be preferable, as that is equally breaking and on the side of incorrect WRT to the types themselves. I think if remaining type issues only affect R3F usage, we loosen types there. Otherwise, we need to be creative to minimize breakage to vanilla users and collaborate with three types on this if possible.

RodrigoHamuy:
Okidok, in that case if you are happy @CodyJasonBennett I will close https://github.com/three-types/three-ts-types/pull/1398 +  https://github.com/pmndrs/react-three-fiber/issues/3416 in favour of https://github.com/pmndrs/three-stdlib/pull/392 + https://github.com/pmndrs/drei/pull/2276

Issue #1949, opened 2024-05-02
useMask hook and Mask component doesn't work with the current version of three.js


Dooqp:
- `three` version: 0.164.0
- `@react-three/fiber` version: 8.16.3
- `@react-three/drei` version: 9.105.6
- `node` version: v20.9.0
- `npm` (or `yarn`) version: 1.22.19

### Problem description:
I've used useMask hook and Mask component without a problem for a while. But today I've upgraded my packages and they don't work properly anymore. They still work with the older version of three.js (0.155.0)

### Relevant code:
I've tried to reproduce this issue with the drei's own useMask example and it doesn't work either so I'll just leave the examples link below
https://www.npmjs.com/package/@react-three/drei

### Suggested solution:
I don't really have any suggested solution but I tried the older version of three.js (0.155.0) and it works as expected with the updated drei package. Maybe this helps.

Jarodwr:
as of threejs 163 you need to toggle the stencil flag on your canvas gl property:
![image](https://github.com/pmndrs/drei/assets/12545583/f3d66c4f-fd49-48fb-9f2a-3fa5293639c0)

laurentlahmy:
Really worth updating the documentation! I just wasted hours on this.

PR #2287, opened 2025-01-02
docs(Mask): mention enabling stencil in the Canvas component

bolex222:
### Why

Since three 0.163.x `stencil` must be set to true in the `gl` props of the `Canvas` component to use the useMask hook without depthWrite but the documentation doesn't mention it yet.
see the following issue: https://github.com/pmndrs/drei/issues/1949

### What

I'm adding a mention about the `stencil: true` value in the `gl` props of the `<Canvas/>`component.

### Checklist

- [x] Documentation updated ([Mask](https://drei.docs.pmnd.rs/portals/mask))
- [ ] Storybook entry added
- [x] Ready to be merged

### DIFF
--- a/docs/portals/mask.mdx
+++ b/docs/portals/mask.mdx
@@ -44,6 +44,14 @@ return (
     <meshStandardMaterial {...stencil} />
 ```
 
+You must unable stencil in your `<Canvas>` component to use the mask without `depthWrite` set to `true`.
+
+```jsx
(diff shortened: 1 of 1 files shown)


sngazm:
Why this PR not merged yet?

bolex222:
@sngazm since the ticket has been closed I believe it's not in their plan to add it to the doc, keeping it open for now cause we never know.

PR #2303, opened 2025-01-16
Allow using `TTFLoader` in `useFont`

### LINKED ISSUE #2094: Allow `Text3D` to use `TTFLoader`

mz8i:
Resolves #2094
(this is a second PR after #2095 got reverted to avoid a breaking change).

In comparison to the previous PR, the code here does not try to automatically determine if `TTFLoader` should be used.
It is only enabled manually through the second argument to `useFont`.

Additional considerations:
- currently there's no option to pass the `reversed` option to `TTFLoader`, which is required to correctly parse some font files. Suggestions for the cleanest API here are welcome.
- I wasn't 100% sure about the correct setup for the `preload` and `clear` functions for `useFont` in the case where `useFont` has more than one argument.

### Checklist

- [ ] Ready to be merged


mz8i:
@abernier creating this new PR as you suggested in #2095 - any comments to the above points will be appreciated!

CodyJasonBennett:
Why wasn't feature detection in #2095 done by looking at the file extension?

mz8i:
Originally I thought detecting based on MIME type would be more generic. But I removed the detection anyway - correct me if I'm wrong, but I think any sort of detection would potentially be a breaking change for users who rely on the hook to always apply `FontLoader` regardless of extension , MIME type etc (and avoiding a breaking change was why the original PR was reverted)

mz8i:
@abernier @CodyJasonBennett would you have any more comments about this PR?

PR #2318, opened 2025-01-20
v10

CodyJasonBennett:
React 19 support. See [R3F v9 changelog](https://github.com/pmndrs/react-three-fiber/releases/tag/v9.0.0).

PR #2335, opened 2025-02-04
Update package.json

ShayokhShorfuddin:
three-mesh-bvh 0.7.8 is incompatible with ThreeJS. Changed the package to be ^0.8.0.

### Why
I get a deprecation warning for three-mesh-bvh@0.7.8 when I do `npm install`. This is not a big deal locally but when I upload my codebase to MelloHost CPanel and do `npm install`, it shows the warning in a popup and refuses to start the app unless the warning is resolved. That's why, I have updated the dependency to ^0.8.0

image

### What
I have updated the three-mesh-bvh dependency from  ^0.7.8  to ^0.8.0

### Checklist
- [x] Ready to be merged


CodyJasonBennett:
This is a breaking change to do, unfortunately, but will be done in the next major with React 19 support #2318.

PR #2354, opened 2025-02-16
feat: add MotionPathControls control via ref + add two properties

Riment:
Adding control of the motion state of MotionPathControls via the reference.
Adding `debugColor` property to change debug helpers color.
Adding `loop` property to control if the target object loop back to the start when reaching the end of the curve.

### Why

Avoid using only the `useMotion` hook inside MotionPathControls component to control the target object.

### What

Add `motion` property to the MotionPathControls ref.
Handle `useMotion` hook error when used outside of MotionPathControls component.
Update doc.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/MotionPathControls.tsx
+++ b/src/core/MotionPathControls.tsx
@@ -8,10 +8,14 @@ type MotionPathProps = JSX.IntrinsicElements['group'] & {
   curves?: THREE.Curve<THREE.Vector3>[]
   /** Show debug helpers */
   debug?: boolean
+  /** Color of debug helpers */
+  debugColor?: THREE.ColorRepresentation
   /** The target object that is moved, default: null (the default camera) */
   object?: React.MutableRefObject<THREE.Object3D>
   /** An object where the target looks towards, can also be a vector, default: null */
   focus?: [x: number, y: number, z: number] | React.MutableRefObject<THREE.Object3D>
+  /** Should the target object loop back to the start when reaching the end, default: true */
+  loop?: boolean
   /** Position between 0 (start) and end (1), if this is not set useMotion().current must be used, default: null */
   offset?: number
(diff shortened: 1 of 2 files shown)


src/core/MotionPathControls.tsx
```diff
@@ -45,46 +49,49 @@ type MotionState = {
   next: THREE.Vector3
 }
 
-const isObject3DRef = (ref: any): ref is React.MutableRefObject<THREE.Object3D> =>
-  ref?.current instanceof THREE.Object3D
+export type MotionPathRef = THREE.Group & { motion: MotionState }
 
-const context = /* @__PURE__ */ React.createContext<MotionState>(null!)
+const isObject3DRef = (ref: any): ref is React.MutableRefObject<THREE.Object3D> => ref?.current instanceof THREE.Object3D
```
netgfx, line 54: please replace `any` with proper type

src/core/MotionPathControls.tsx
```diff
@@ -114,10 +123,10 @@ export const MotionPathControls = /* @__PURE__ */ React.forwardRef<THREE.Group,
 
     React.useLayoutEffect(() => {
       path.curves = []
-      const _curves = curves.length > 0 ? curves : ref.current?.__r3f.objects
-      for (var i = 0; i < _curves.length; i++) path.add(_curves[i])
+      const _curves = curves.length > 0 ? curves : (ref.current as any)?.__r3f?.objects ?? []
```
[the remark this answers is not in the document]
CodyJasonBennett: There is no proper type for this yet.

src/core/MotionPathControls.tsx
CodyJasonBennett: This is okay as a simplification with unknown.

PR #2372, opened 2025-02-24
fix(MeshReflectorMaterial): avoid updates when possible

davcri:
### Why

These changes avoids unnecessary updates for MeshReflectorMaterial. 
Previously it was updating even if `visible=false` was set on the material or on the parent mesh.

A couple of notes:

1. I had to update `three` and `@types/three` package because `onBeforeRender()` method was not typed correctly up til https://github.com/three-types/three-ts-types/pull/1169 and it was causing errors
2. sorry I disabled Husky locally as I was getting a lot of errors that were not related to these changes


src/core/MeshReflectorMaterial.tsx
```diff
@@ -194,12 +192,12 @@ export const MeshReflectorMaterial: ForwardRefComponent<MeshReflectorMaterialPro
 
       useFrame(() => {
         const parent = (materialRef.current as any).parent || (materialRef.current as any)?.__r3f.parent?.object
-        if (!parent) return
+        if (!parent || !materialRef.current.visible || !parent.visible || !shouldUpdate.current) return
```
davcri, line 195: I used `shouldUpdate` to avoid unnecessary instructions in `useFrame` which  is called unconditionally.

Three's `onBeforeRender` instead is called only when the renderer needs it (EG: it shouldn't be called if the mesh isn't part of the scene being rendered).

PR #2409, opened 2025-04-05
fix: MeshReflectorMaterial creating FBO every re-render

Nickardson:
### Why

The `<MeshReflectorMaterial/>` had a useMemo referencing a blur array which would be a different object reference every React render, causing the FBO and other expensive things to be recreated every render.

### What

Changed the useMemo to check the values of the blur array rather than the array itself.

### Checklist

- [x] Ready to be merged


src/core/MeshReflectorMaterial.tsx
```diff
@@ -57,6 +57,8 @@ export const MeshReflectorMaterial: ForwardRefComponent<MeshReflectorMaterialPro
       const scene = useThree(({ scene }) => scene)
       blur = Array.isArray(blur) ? blur : [blur, blur]
       const hasBlur = blur[0] + blur[1] > 0
+      const blurX = blur[0];
+      const blurY = blur[1];
```
CodyJasonBennett, line 61:
```suggestion
      const blurX = blur[0]
      const blurY = blur[1]
```

PR #2427, opened 2025-04-20
Updated useCursor to support cursor styles updating

ivoilic:
Adds `onPointerOver` and `onPointerOut` as dependencies of the useEffect hook inside of useCursor.

### Why

The cursor styles currently don't update once they have been set.

### What

I added `onPointerOver` and `onPointerOut` as dependencies of the useEffect hook inside of useCursor.

### Checklist

- [X ] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/docs/misc/example.mdx?plain=1))
- [X ] Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/Example.stories.tsx))
- [X] Ready to be merged

### DIFF
--- a/src/web/useCursor.tsx
+++ b/src/web/useCursor.tsx
@@ -11,5 +11,5 @@ export function useCursor(
       container.style.cursor = onPointerOver
       return () => void (container.style.cursor = onPointerOut)
     }
-  }, [hovered])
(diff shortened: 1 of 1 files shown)


DennisSmolek:
The point of the component is you set the "hovered" state outside of it. It wont need anything but that

PR #2428, opened 2025-04-21
fix: typos

s-rigaud:
### What

Multiple typos fix in various files


DennisSmolek:
implemented most but there are some that were legit spellings (Skissor etc)

abernier:
@DennisSmolek can you reference where it is merged so @s-rigaud can eventually review?

Issue #1529, opened 2023-06-21
PresentationControls doesn't trigger invalidate() onChange


stefanwagnerdev:
- `three` version: 0.153.0
- `@react-three/fiber` version: 8.13.3
- `@react-three/drei` version: 9.74.16
- `node` version: v18.16.0
- `npm` (or `yarn`) version: 9.5.1

### Problem description:

Camera does not render the scene if presentation controls are used

### Relevant code:

https://codesandbox.io/s/cocky-dust-ql5qnq?file=/src/index.jsx:0-539

```js
<Canvas frameloop="demand" camera={{ position: [2, 2, 3], fov: 25 }}>
    <PresentationControls
        config={{ mass: 1, tension: 170, friction: 26 }}
        polar={[0, 0]}
        azimuth={[-Math.PI / 10, Math.PI / 10]}
    >
        <group position={[-.4, 0, 0.6]}>
           <mesh castShadow rotation={[0, 0, 0]}>
              <boxGeometry args={[0.1, 0.1, 0.1]} />
              <meshStandardMaterial color="#FFFFFF" />
          </mesh>
        </group>
    </PresentationControls>
    <ContactShadows opacity={0.5} blur={1.5} far={0.9} />
</Canvas>
```

### Suggested solution:

Is it possible to add an invalidate() if the PresentationControls are animating? Maybe in onChange?

feresr:
I am having the same issue. Have you found a workaround? (other than setting frameloop="always")

DennisSmolek:
implemented in v11

PR #2432, opened 2025-04-23
fix/1529-invalidate-presentation-controls

nielsvdpol:
### Why
Resolves: https://github.com/pmndrs/drei/issues/1529
When using `<Canvas frameloop="demand" />`, the `PresentationControls` component was non-functional because interactions and easing animations did not automatically trigger re-renders. Since `frameloop="demand"` disables continuous rendering, it requires manual invalidation of the render loop to update the scene. Without this, changes to rotation or scale weren't reflected visually.

This PR fixes that issue by ensuring the component properly invalidates the render loop on interaction and animation updates, restoring expected behavior when `frameloop="demand"` is in use.

### What

- Fixed broken behavior under `frameloop="demand"` by manually triggering re-renders with `invalidate()` from `useThree`.
- Called `invalidate()`:
   - Inside `onDrag` and `onDragEnd` gesture handlers to ensure updates apply during/after interactions.
   - Inside the `useFrame` loop when easing causes a change to rotation or scale.
- This ensures the component works correctly in both default and `frameloop="demand"` render modes.

### Checklist
- [x] Ready to be merged

### DIFF
--- a/src/web/PresentationControls.tsx
+++ b/src/web/PresentationControls.tsx
@@ -35,6 +35,7 @@ export function PresentationControls({
 }: PresentationControlProps) {
   const events = useThree((state) => state.events)
   const gl = useThree((state) => state.gl)
+  const invalidate = useThree((state) => state.invalidate);
   const explDomElement = domElement || events.connected || gl.domElement
 
   const { size } = useThree()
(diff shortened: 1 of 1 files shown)


DennisSmolek:
implemented in v11

PR #2449, opened 2025-05-26
Fix decal debug util to update its transforms when the props change / add scene object to keep track of decal transforms

itsdouges:
### Why

A [bug report was made to Triplex](https://github.com/trytriplex/triplex/issues/336) noting that transform controls weren't working well with decal. After debugging it's because Triplex doesn't have easy access to the current decal transforms in the scene graph.

While I'm here I've also fixed the decal helper not updating when transforms change.

### What

I've declared transforms on a sibling scene object that Triplex picks up when "selecting" the decal. I've also moved the decal helper transforms to the jsx element so it updates when they change.

I'm going to explore an alternate fix in Triplex where it doesn't need to read from the scene. So if we're uncomfortable with this change I can remove everything except for the debug transform fix.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/Decal.tsx
+++ b/src/core/Decal.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react'
 import * as THREE from 'three'
 import * as FIBER from '@react-three/fiber'
-import { applyProps } from '@react-three/fiber'
+import { applyProps, useFrame } from '@react-three/fiber'
 import { DecalGeometry } from 'three-stdlib'
 import { ForwardRefComponent } from '../helpers/ts-utils'
 
@@ -71,7 +71,6 @@ export const Decal: ForwardRefComponent<DecalProps, THREE.Mesh> = /* @__PURE__ *
         if (parent.geometry.attributes.normal === undefined) parent.geometry.computeVertexNormals()
         const normal = parent.geometry.attributes.normal.array
         let distance = Infinity
-        let closest = new THREE.Vector3()
         let closestNormal = new THREE.Vector3()
         const ox = o.position.x
         const oy = o.position.y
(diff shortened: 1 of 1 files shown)


src/core/Decal.tsx
```diff
@@ -71,7 +71,6 @@ export const Decal: ForwardRefComponent<DecalProps, THREE.Mesh> = /* @__PURE__ *
         if (parent.geometry.attributes.normal === undefined) parent.geometry.computeVertexNormals()
         const normal = parent.geometry.attributes.normal.array
         let distance = Infinity
-        let closest = new THREE.Vector3()
```
itsdouges, line 74: This wasn't used so I removed it.

src/core/Decal.tsx
```diff
@@ -116,31 +115,33 @@ export const Decal: ForwardRefComponent<DecalProps, THREE.Mesh> = /* @__PURE__ *
 
   React.useLayoutEffect(() => {
     if (helper.current) {
-      applyProps(helper.current as any, state.current)
```
itsdouges, line 119: This has been moved to the decal helper itself.

src/core/Decal.tsx
```diff
@@ -116,31 +115,33 @@ export const Decal: ForwardRefComponent<DecalProps, THREE.Mesh> = /* @__PURE__ *
...
-        <mesh ref={helper}>
-          <boxGeometry />
-          <meshNormalMaterial wireframe />
-          <axesHelper />
-        </mesh>
-      )}
-    </mesh>
+    <>
+      {/* We add the transforms to another object3d to keep them in the scene graph so editors like triplex.dev can pick them up.  */}
+      <group position={position} rotation={rotation} scale={scale} />
```
itsdouges, line 126: This lets visual editors that need to have the transform data in the scene graph get access to it. I couldn't add this to the root `mesh` as it breaks the decal geometry expectations.

itsdouges:
I'm able to improve https://triplex.dev instead of changing this 😄.

Issue #2005, opened 2024-06-21
Rendering starts too late with frameloop demand and CameraControls transitions


AaronClaes:
- `three` version: 0.0165
- `@react-three/fiber` version: 8.16.8
- `@react-three/drei` version: 9.107.0

### Problem description:

When setting the canvas frameloop to `demand`, the `<CameraControls />` functions trigger `invalidate()` too late. 

When the canvas is not being rendered and you do an action like this, you will notice the problem:
```tsx 
controls.dolly(1, true) // true for the transition to be smooth
```

When the transition is set to `false`, it is not noticeable. But with the transition, the first few frames are skipped. 
When de Canvas is already rendering, this problem does not happen, it is only the initial trigger.

### Relevant code:

In the codesandbox below you will notice that if you give the Canvas time to stop rendering, and you then click the box, the initial transition of the camera controls is not smooth. If you keep clicking it, the following transitions (when the canvas is still running) are smooth. 

https://codesandbox.io/p/sandbox/use-controls-bug-2mtn6l

### Suggested solution:

I am unsure if this is a @react-three/fiber problem or something in the CameraControls component. It might even have something to do with how the camera-controls library works, but it would be nice to see this fixed, or have a workaround.

I have tried to use `invalidate` to start rendering earlier but did not have any success with this.

Thanks!

vmakela:
The CameraControls update code looks pretty suspicious to me. It only calls `controls.update` for the implementation on every frame rendered, and it only forces renders on camera events. It does kind of work for animated transitions because the `transitionstart` event gets fired when starting the transition, but it does come with a delay.

Something like this would fix it, but I'm not sure if it is actually wanted here.

```diff
diff --git a/src/core/CameraControls.tsx b/src/core/CameraControls.tsx
index 082b967..b2dcab7 100644
--- a/src/core/CameraControls.tsx
+++ b/src/core/CameraControls.tsx
@@ -1,5 +1,6 @@
 import {
   Box3,
+  Clock,
   EventDispatcher,
   MathUtils,
   Matrix4,
@@ -16,7 +17,7 @@ import {
 
 import * as React from 'react'
 import { forwardRef, useMemo, useEffect } from 'react'
-import { extend, useFrame, useThree, ReactThreeFiber, EventManager } from '@react-three/fiber'
+import { extend, useThree, ReactThreeFiber, EventManager } from '@react-three/fiber'
 
 import CameraControlsImpl from 'camera-controls'
[...]

PR #2451, opened 2025-05-28
feat: CameraControls support for 1:1 events `onX` callback props

abernier:
supports all `onX` https://github.com/yomotsu/camera-controls/tree/dev?tab=readme-ov-file#events

| Event name          | Timing |
| ------------------- | ------ |
| `'controlstart'`    | When the user starts to control the camera via mouse / touches. ¹ |
| `'control'`         | When the user controls the camera (dragging). |
| `'controlend'`      | When the user ends to control the camera. ¹ |
| `'transitionstart'` | When any kind of transition starts, either user control or using a method with `enableTransition = true` |
| `'update'`          | When the camera position is updated. |
| `'wake'`            | When the camera starts moving. |
| `'rest'`            | When the camera movement is below `.restThreshold` ². |
| `'sleep'`           | When the camera end moving. |

+ backward compat for `onStart`/`onEnd`/`onChange`

- [x] Ready to be merged

---

# 🐛 it also fixes a bug with `frameloop="demand"`

where calling programmatic CameraControls methods caused a jerky animation

Before:

https://github.com/user-attachments/assets/9ab0fe2f-c523-4af1-b212-f20cbecad427

After:

https://github.com/user-attachments/assets/10d85755-288c-46b9-a6b5-d03299298cb5

### DIFF
--- a/.storybook/stories/CameraControls.stories.tsx
+++ b/.storybook/stories/CameraControls.stories.tsx
@@ -9,22 +9,17 @@ import { Box, CameraControls, PerspectiveCamera, Plane, useFBO } from '../../src
 export default {
   title: 'Controls/CameraControls',
   component: CameraControls,
-  decorators: [
-    (Story) => (
-      <Setup controls={false}>
-        <Story />
-      </Setup>
-    ),
-  ],
 } satisfies Meta<typeof CameraControls>
 
 type Story = StoryObj<typeof CameraControls>
+
+//
 
 function CameraControlsScene1(props: ComponentProps<typeof CameraControls>) {
   const cameraControlRef = useRef<CameraControls>(null)
 
   return (
-    <>
+    <Setup controls={false}>
       <CameraControls ref={cameraControlRef} {...props} />
       <Box
         onClick={() => {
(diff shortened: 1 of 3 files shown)


abernier:
> Still, the camera teleports big distances because of [#2005 (comment)](https://github.com/pmndrs/drei/issues/2005#issuecomment-2220002815) in functions `setLookAt()` and similar functions.

does https://r3f.docs.pmnd.rs/advanced/scaling-performance#sync-animations-with-on-demand-rendering-and-invalidate prevents the issue?

if yes it's probably good enough, right?

FrostKiwi:
> does userland's https://r3f.docs.pmnd.rs/advanced/scaling-performance#sync-animations-with-on-demand-rendering-and-invalidate prevent the issue?
> 
> if yes it's probably good enough, right?

Yes and no.
Indeed, following the doc I can go through all my code and patch every usage of stuff that animates from `controls.setLookAt(-2, 1.5, 3, 0.470086, 1.16085, 0, true)` to `invalidate(); requestAnimationFrame(() => controls.setLookAt(-2, 1.5, 3, 0.470086, 1.16085, 0, true));`.

But that doesn't cover all cases and AFAIK, that leaves one special case out: Scrollwheel, which is handled specially. Scrollwheel always results in small teleportation.
So... am I missunderstanding something?: We can't use `CameraControls` events like `wake` as they fire too late. You have to go lower level, below React Three Fiber and hook into canvas to catch scroll events Events and issue an invalidate + requestAnimationFrame combo? I'm smelling a race condition here.

PR #2461, opened 2025-06-09
fix: (caustics) use halfFloatType as fallback when FloatType is not supported

vis-prime:
### Why

The FBOs in Caustics currently use `FloatType` by default, which causes issues on devices that do not support the `OES_texture_float_linear` WebGL extension. While this does not crash the renderer, caustics are not rendered and warnings appear in the console.

### What

This PR updates the code to check if the extension is supported using the `renderer.extensions.get('xxxx')` method. If not supported, it falls back to `HalfFloatType`.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/Caustics.tsx
+++ b/src/core/Caustics.tsx
@@ -270,7 +270,12 @@ const NORMALPROPS = {
   type: THREE.UnsignedByteType,
 }
 
-const CAUSTICPROPS = {
+const CAUSTICPROPS: {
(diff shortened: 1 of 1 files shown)


vis-prime:
@abernier is this safe/ready to be merged ?

DennisSmolek:
This is another crazy material/shader. This legacy fix has been applied

PR #2465, opened 2025-06-13
fix: (RenderTexture): Add optional chaining to prevent TypeError

gaievskyi:
### Why

RenderTexture.js:39 Uncaught TypeError: Cannot read properties of undefined (reading 'parent')
    at RenderTexture.useCallback[uvCompute] [as compute] (RenderTexture.js:39:29)
    at handleRaycast (events-dc44c1b8.esm.js:541:62)
    at Array.flatMap (<anonymous>)
    at intersect (events-dc44c1b8.esm.js:553:6)
    at HTMLDivElement.handleEvent (events-dc44c1b8.esm.js:788:20)

### What

Add optional chaining to prevent runtime Uncaught TypeError when obtaining the parent. If parent is undefined, we fall into the next line with truthy check and early return instead of runtime error. 
```ts
if (!parent) return false
```

### DIFF
--- a/src/core/RenderTexture.tsx
+++ b/src/core/RenderTexture.tsx
@@ -60,9 +60,9 @@ export const RenderTexture: ForwardRefComponent<RenderTextureProps, THREE.Textur
       // Since this is only a texture it does not have an easy way to obtain the parent, which we
       // need to transform event coordinates to local coordinates. We use r3f internals to find the
       // next Object3D.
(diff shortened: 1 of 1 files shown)


DennisSmolek:
im not sure when but the first conditional  in the `let parent` line is already in the renderTexture handler. 
The second conditional within the while is redundant as the first check of the while is that it exists

PR #2472, opened 2025-06-24
html-custom-camera-support

calcifer-3118:
feat(Html): support rendering Html with a specified custom camera

### Why

Currently, the `<Html />` component always renders using the active camera from the R3F context (`useThree().camera`). In multi-camera setups, this causes HTML overlays to shift or misalign when the active camera changes.

This PR adds support for a `camera` prop, allowing developers to lock `<Html />` rendering to a specific camera — improving control and reliability in HUDs, fixed screens, and persistent overlays.

### What

- Adds a `camera` prop to `<Html />`
- Defaults to the current active R3F camera (`useThree().camera`) if not provided (non-breaking)
- Enables stable Html rendering in multi-camera scenes
- Fully backward compatible

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -1,5 +1,5 @@
-import * as React from 'react'
-import * as ReactDOM from 'react-dom/client'
+import * as React from "react";
+import * as ReactDOM from "react-dom/client";
 import {
   Vector2,
   Vector3,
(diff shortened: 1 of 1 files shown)


DennisSmolek:
I'm hesitant to do this... Views, Portals, and connected Canvases all allow to switch the contextual camera and html has a lot of vodoo to bind it. 
If you can provide an example that would need this where Hud, View, Portal, or a second canvas wont work I'll reopen

PR #2498, opened 2025-07-23
Update MeshRefractionMaterial.tsx

jflatt-gia:
geometry.toNonIndexed() can produce warning "Geometry is already non-indexed".  Avoid excessive warnings by adding a check

### Checklist

- [x] Ready to be merged


DennisSmolek:
this material is on my hate list, but this was added. We'll likely need to refactor this in general but I've added it and started on a webgpu version for v11

PR #2527, opened 2025-09-02
docs: Fix PivotControls onDrag function parameter destructuring

vis-prime:
Destructuring not needed, first parameter is the matrix4

### Why
Typescript printed this error and saw that there is no object to destructure

### What
use the first matrix4 directly

### Checklist

- [x] Ready to be merged


DennisSmolek:
We added more to this and some other handlers, destructuring now needed

PR #2529, opened 2025-09-07
fix(type): Update useHelper to accept nullable Object3D ref

Cybermb:
Fix type error `Type 'null' is not assignable to type 'Object3D<Object3DEventMap>'` when using useHelper(ref, ...) as ref can be null

### Why

When using useHelper(ref, ...) you get type error

### What

Added union type null because react refs can be null

### Checklist

- [x] Ready to be merged


DennisSmolek:
implemented in v11

PR #2538, opened 2025-09-16
feat(DragControls): Pass through useGesture handlers

astralarya:
### Why

useGesture, the underlying library for DragControls, supports passing handlers such as `onClick` to be called seamlessly during gesture handling.  This enables us to call things like `event.stopPropagation()` to occlude raycast events from React Three Fiber.

### What

Add a `handlers` prop, which is spread into the useGesture call.

### Checklist

- [x] Ready to be merged


DennisSmolek:
DragControls no longer uses useGesture because of some nesting/events issues. so this setup doesnt quite match with v11. We may readress later though!

PR #2541, opened 2025-09-20
feat(`ScrollControls`&`Html`): add pixelPerfect prop to prevent subpixel blur

### LINKED ISSUE #859: Html component render blurry when using transform attribute on firefox
- `three` version: 0.139.2
- `@react-three/fiber` version: 8.0.10
- `@react-three/drei` version: 9.4.2
- `node` version: 16.13.2
- `yarn` version: 1.22.17
- `Firefox Developer Edition` version: 100.0b4 (64-bit)

### Problem description:

Elements placed into the `<Html>` component render blurry when using the `transform` attribute.

## Firefox:
image

## Edge:
image

### Relevant code:

```jsx
<Html transform>
```

### LINKED ISSUE #2380: Html: Blurry text caused by subpixel translate3d() positioning
When using drei/html, the rendered text sometimes appears blurry. This issue occurs because the transform: translate3d(x, y, z) applied by drei contains decimal values for x and y, which can cause subpixel rendering artifacts.

Image

A possible improve could be round the position to current device's native pixel ratio:
```ts
function roundToPixelRatio(value) {
  const ratio = window.devicePixelRatio;
  return Math.round(value * ratio) / ratio;
}
```

Apply this to 
```ts
            el.style.transform = `translate3d(${roundToPixelRatio(vec[0])}px,${roundToPixelRatio(vec[1])}px,0) scale(${scale})`
```

Not sure if this conflicts with other use cases, if so may we add a new option  `roundToPixelRatio: boolean` so that we can enable / disable this feature ?

Thanks

artokun:
## Summary

Adds `pixelPerfect` prop to both `ScrollControls` and `Html` components to prevent subpixel rendering blur caused by CSS `transform3d` values with decimal pixels.

## Problem

HTML elements rendered with CSS transforms in 3D scenes often appear blurry due to subpixel positioning, especially during animations. This affects text readability and overall visual quality, as discussed in issues #859 and #2380.

## Solution

Implements a `pixelPerfect` prop that rounds transform values to whole pixels using `devicePixelRatio`-based calculation for accurate rendering across all display types.

### Key Features

- **ScrollControls**: `pixelPerfect` prop for `<Scroll html>` components
- **Html**: `pixelPerfect` prop for transform mode rendering
- **DevicePixelRatio Support**: Proper high-DPI display handling via `window.devicePixelRatio`
- **Performance Optimized**: Velocity-based conditional application (only when motion < 0.001)
- **Backward Compatible**: Default behavior unchanged, opt-in via prop

### Usage

```jsx
// ScrollControls
<ScrollControls pages={3}>
  <Scroll html pixelPerfect>
    <div>Crisp text during scroll animations</div>
  </Scroll>
</ScrollControls>

// Html component
<Html transform pixelPerfect>
  <div>Crisp text during 3D transforms</div>
</Html>
```

## Implementation Details

- **Helper Function**: `roundToPixelRatio()` uses `window.devicePixelRatio` for accurate rounding
- **Velocity Detection**: Only applies rounding when velocity < 0.001 to maintain smooth animation
- **Cross-Component**: Consistent implementation across ScrollControls and Html components

## Documentation & Examples

- ✅ Updated component documentation with usage examples
- ✅ Comprehensive Storybook stories with before/after comparisons
- ✅ Interactive demos showing the difference in text crispness

## Credit

Special thanks to [@chris-xinhai-li](https://github.com/chris-xinhai-li) for suggesting the `devicePixelRatio` approach in issue #2380.

## Fixes

- Closes #859
- Closes #2380

## Testing

The feature has been thoroughly tested with:
- Various device pixel ratios (1x, 1.5x, 2x, etc.)
- Different animation scenarios (scroll, transform, etc.)
- Performance impact validation
- Cross-browser compatibility

### DIFF
--- a/.storybook/stories/HTML.stories.tsx
+++ b/.storybook/stories/HTML.stories.tsx
@@ -12,6 +12,9 @@ import { Meta, StoryObj } from '@storybook/react-vite'
 export default {
   title: 'Misc/Html',
   component: Html,
+  args: {
+    pixelPerfect: false, // Round transforms to whole pixels for crisp rendering (default: false)
+  },
   decorators: [
     (Story) => (
       <Setup cameraPosition={new THREE.Vector3(-20, 20, -20)}>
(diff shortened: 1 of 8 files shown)


DennisSmolek:
Went a slightly different way with this. I've set the system to be pixelPerfect at all times and not do any velocity or whatever handling. I'm not sure it would even be visible. but you can disable it if it is noticeable.

PR #2557, opened 2025-10-13
fix(ScrollControls): cache ReactDOM root for React 19 compatibility

iamadagostino:
## Description
Prevent duplicate `createRoot()` calls by caching the root instance on the DOM element, fixing React 19 error on component remount.

## Why
React 19 throws an error when `ReactDOM.createRoot()` is called multiple times on the same container. `ScrollHtml` was creating a new root on each remount, breaking compatibility with React 19.

## What
- Cache ReactDOM root instance on the DOM element using `__r3_root` property
- Reuse cached root on component remount
- Fixes React 19 compatibility while maintaining backwards compatibility

## Checklist
- [x] Ready to be merged

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,10 @@ on:
       - 'rc'
   pull_request: {}
 
+permissions:
+  contents: write
+  id-token: write
+
 # Cancel any previous run (see: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency)
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -70,4 +74,3 @@ jobs:
       - run: yarn release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/docs/getting-started/introduction.mdx
+++ b/docs/getting-started/introduction.mdx
@@ -12,6 +12,7 @@ nav: -1
 [![logo](../logo.jpg)](https://codesandbox.io/s/bfplr)
 
 [![Storybook](https://img.shields.io/static/v1?message=Storybook&style=flat&colorA=000000&colorB=000000&label=&logo=storybook&logoColor=ffffff)](https://drei.pmnd.rs/)
+[![](https://img.shields.io/badge/chromatic-171c23.svg?style=flat&colorA=000000&colorB=000000&logo=chromatic&logoColor=ffffff)](https://www.chromatic.com/library?appId=64a019f36ecd3751d0ada612&branch=master)
 [![Version](https://img.shields.io/npm/v/@react-three/drei?style=flat&colorA=000000&colorB=000000)](https://www.npmjs.com/package/@react-three/drei)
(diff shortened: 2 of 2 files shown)


src/web/ScrollControls.tsx
```diff
@@ -236,7 +236,32 @@ const ScrollHtml: ForwardRefComponent<{ children?: React.ReactNode; style?: Reac
       React.useImperativeHandle(ref, () => group.current, [])
       const { width, height } = useThree((state) => state.size)
```
artokun: can you update useThree types to include state.fixed and tag me for review again?

src/web/ScrollControls.tsx
```diff
@@ -236,7 +236,32 @@ const ScrollHtml: ForwardRefComponent<{ children?: React.ReactNode; style?: Reac
...
+        try {
+          // If a root was previously created on this element, reuse it.
+          if (state.fixed && (state.fixed as any).__r3_root) return (state.fixed as any).__r3_root
+          const r = ReactDOM.createRoot(state.fixed)
+          try {
+            Object.defineProperty(state.fixed, '__r3_root', {
+              value: r,
+              configurable: true,
+              writable: false,
+            })
```
artokun: verify the return type of the useThree lambda, if it's a valtio proxy, you'll need to properly define it in the store. we don't want to be setting properties this way.

src/web/ScrollControls.tsx
iamadagostino: - The `state` variable used in the root caching logic (lines 245-264) comes from `useScroll()` on line 234, which returns `ScrollControlsState` (already typed with `fixed: HTMLDivElement` on line 37).
- This `useThree()` call only extracts `size` for layout calculations and is unrelated to the `state.fixed` property being accessed below.

src/web/ScrollControls.tsx
iamadagostino: The `state.fixed` here is from `useScroll()` (line 234), which returns `ScrollControlsState` - not from R3F's store. It's a plain `HTMLDivElement` created via `useState` (line 68), not a Valtio proxy.
Setting `__r3_root` on this DOM element doesn't mutate any reactive store state.
The property is defined non-enumerable and is used solely to cache the ReactDOM root instance to prevent React 19's "createRoot already called" error on remount.

src/web/ScrollControls.tsx
```diff
@@ -236,7 +236,32 @@ const ScrollHtml: ForwardRefComponent<{ children?: React.ReactNode; style?: Reac
...
-      const root = React.useMemo(() => ReactDOM.createRoot(state.fixed), [state.fixed])
+      // Cache a root instance on the fixed DOM element to avoid calling
+      // ReactDOM.createRoot multiple times on the same container which
+      // triggers an error in React 19. We store the created root in a
+      // non-enumerable property on the element so repeated mounts or
+      // remounts re-use the same root instance.
+      const root = React.useMemo(() => {
+        try {
+          // If a root was previously created on this element, reuse it.
+          if (state.fixed && (state.fixed as any).__r3_root) return (state.fixed as any).__r3_root
```
artokun: What I am trying to avoid is this. why is there a `__r3_root` property on a HTMLDIvElement? Seems odd that we need to typecast this to any.

PR #2570, opened 2025-11-12
Fix: update e2e tests

DennisSmolek:
### Why
Changes in[ three 181](https://github.com/mrdoob/three.js/pull/32054) slightly modify how envmaps reflect off surfaces. That is what we use as a visual check for the test environment.

### What

- Pinned Tests to current three.js minimum
- Added minimum version to test file name
- Created method to generate new test image easily when we move the minimum
- Added test to confirm that a new image was generated
- Added set browsers to generation
- Tests output results to folders to be replayed/viewed
- Because of replayability, added console logging to three vals/colorspace to quickly compare issues. 
- Extended GH Action to upload artifacts/results onto tests run so we can see failing tests directly on PRs. (30 day retention)
- Minor adjustments related to line endings
- Minor adjustments to dev container
- Minor general test adjustments/logging

### Remaining
This PR might trigger the action change so we may see it, but that part remains untested.
Would like a non-windows dev to test the PR and run tests locally to ensure I didn't break anything


DennisSmolek:
Ok I thought it was me messing up the line endings on a few files but its actually that some files in the repo were stored as CRLF.

I spent forever trying to figure out what I did wrong..

DennisSmolek:
image

Test report now shows all tests.

Added min and latest results but failing image test doesnt fail the run

PR #2579, opened 2025-11-25
fix: Ensure `KeyboardControls` rebinds `onChange` on prop change

### LINKED ISSUE #2578: `KeyboardControls` `onChange` callback retains stale closure when dependencies change
### Problem description:

The `onChange` callback passed to the `KeyboardControls` component becomes stale and continues to refer to the initial function closure.

This occurs because the internal event listener registration within `KeyboardControls` is missing the `onChange` prop from its `useEffect` dependency array.

See https://github.com/pmndrs/drei/blob/master/src/web/KeyboardControls.tsx#L111

### Environment:
- `three` version: 0.180.0
- `@react-three/fiber` version: 10.7.6
- `@react-three/drei` version: 9.3.0
- `node` version: v22.18.0
- `npm` version: 10.9.3

### Relevant code:
See https://codesandbox.io/p/sandbox/quizzical-water-6yy5tn

### Suggested solution:
Add `onChange` to the above `usEffect` deps array.

dammafra:
### Why
Resolves #2578 

This PR fixes a bug where the `onChange` callback provided to the `KeyboardControls` component retains a stale closure if the function depends on reactive state or props from the parent component. This leads to the control handler executing with outdated values.

### What

I have updated the internal `useEffect` hook within `KeyboardControls` that manages the event listener registration. The solution is to explicitly include the `onChange` prop in the dependency array of that `useEffect`.

### Checklist
- [x] Ready to be merged

### DIFF
--- a/src/web/KeyboardControls.tsx
+++ b/src/web/KeyboardControls.tsx
@@ -108,7 +108,7 @@ export function KeyboardControls({ map, children, onChange, domElement }: Keyboa
       source.removeEventListener('keydown', downHandler as EventListenerOrEventListenerObject)
       source.removeEventListener('keyup', upHandler as EventListenerOrEventListenerObject)
     }
-  }, [domElement, key])
+  }, [domElement, key, onChange])
 
   return <context.Provider value={api} children={children} />
 }


DennisSmolek:
implemented in v11

PR #2581, opened 2025-12-02
fix: fixed select behaving erratically when mouse goes over an html element

woodward54:
### Why

Fixed a bug in the `Select` component's box selection feature where `prepareRay` was using `offsetX/offsetY` which fails when the pointer is over UI elements outside the canvas.

Bug:
https://github.com/user-attachments/assets/484b5a04-cd4f-42af-a157-8ad0b5a33b52

### What

- Changed coordinate calculation from `offsetX/offsetY` to `clientX/clientY` relative to the canvas bounding rect, ensuring correct selection even when the pointer is over UI elements
- Added TypeScript typing to function parameters

Fixed:
https://github.com/user-attachments/assets/b891366a-527c-4574-9fe9-5e32a96d6fe4

### Checklist

- [x] Ready to be merged


DennisSmolek:
implemented in v11. I left of the types for now as I think this whole component needs a cleanup pass

PR #2618, opened 2026-01-21
feat: add side prop to the ContactShadows component

diegomacario:
### Why and what

I recently used the `ContactShadows` component to try to render the shadow of a pinball table. Here is what the shadow looks like with blur disabled:

image (8)

As you can see, the shadow is incorrect. The floor of the pinball machine is missing.

This component renders shadows using an orthographic camera that looks upwards along the Y axis.

It replaces all the materials in the scene with a `MeshDepthMaterial`, renders the shadows, and then restores the original materials.

The `side` prop of the `MeshDepthMaterial` it uses is `FrontSide` by default. That explains why the floor of the pinball machine is missing.

In this PR I'm adding a `side` prop to the `ContactShadows` component that allows users to make the `MeshDepthMaterial` double-sided if they want.

Here is what the shadows look like with my change:

image (9)

### Checklist

- [x] Ready to be merged

### DIFF
--- a/docs/staging/contact-shadows.mdx
+++ b/docs/staging/contact-shadows.mdx
@@ -22,3 +22,13 @@ Since this is a rather expensive effect you can limit the amount of frames it re
 ```jsx
 <ContactShadows frames={1} />
 ```
+
+This component renders shadows using an orthographic camera that looks upwards along the Y axis.
+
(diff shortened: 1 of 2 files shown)


DennisSmolek:
This seems like a problem with your mesh and not the shadow. Why is the bottom showing the backside? seems inverted no?

PR #2696, opened 2026-02-17
fix: pass mapped textures to onLoad callback in useTexture

MichaelangJason:
### Why

`onLoad` callback for `useTexture` is typed to accept `MappedTextureType<Url>`, which includes `TextureRecord<Url>`. However, `onLoad` is called with the raw output of `useLoader`, which is cast to `MappedTextureType<Url>` but will never actually be a `TextureRecord<Url>` (only `_Texture[]` or `_Texture`, respectively, `TextureArray<Url>` and `SingleTexture<Url>`). This means an `onLoad` callback expecting a `TextureRecord<Url>` would receive an array instead of a keyed object.

### What

This PR moves the `useMemo` call before the `useLayoutEffect`, and passes the correctly structured `mappedTextures` to the `onLoad` call. 

I noticed that the original `useLayoutEffect` didn't include textures in its dependency list. I assumed this was intentional to avoid re-firing `onLoad` unnecessarily, so I kept the same pattern and didn't include `mappedTextures` in the dependency list either.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/Texture.tsx
+++ b/src/core/Texture.tsx
@@ -22,8 +22,19 @@ export function useTexture<Url extends string[] | string | Record<string, string
   const gl = useThree((state) => state.gl)
   const textures = useLoader(TextureLoader, IsObject(input) ? Object.values(input) : input) as MappedTextureType<Url>
 
+  const mappedTextures = useMemo(() => {
+    if (IsObject(input)) {
+      const keyed = {} as MappedTextureType<Url>
+      let i = 0
+      for (const key in input) keyed[key] = textures[i++]
+      return keyed
+    } else {
+      return textures
+    }
+  }, [input, textures])
+
   useLayoutEffect(() => {
-    onLoad?.(textures)
+    onLoad?.(mappedTextures)
   }, [onLoad])
 
   // https://github.com/mrdoob/three.js/issues/22696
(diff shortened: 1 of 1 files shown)


travisbreaks (commented):
The fix is correct — `onLoad` was receiving raw loader output instead of the mapped texture record. Moving `useMemo` above `useLayoutEffect` ensures `mappedTextures` is computed before the callback fires.

One note: the `useLayoutEffect` dependency array is `[onLoad]` but doesn't include `mappedTextures`. If the textures change (e.g., dynamic URL swap), `onLoad` won't re-fire with the new mapped textures unless the callback reference also changes. This matches the existing behavior before this PR, so not a regression — but worth noting for anyone expecting `onLoad` to fire on texture updates.

The fix itself is clean and minimal. LGTM on the core change.

MichaelangJason:
> The fix is correct — `onLoad` was receiving raw loader output instead of the mapped texture record. Moving `useMemo` above `useLayoutEffect` ensures `mappedTextures` is computed before the callback fires.
> 
> One note: the `useLayoutEffect` dependency array is `[onLoad]` but doesn't include `mappedTextures`. If the textures change (e.g., dynamic URL swap), `onLoad` won't re-fire with the new mapped textures unless the callback reference also changes. This matches the existing behavior before this PR, so not a regression — but worth noting for anyone expecting `onLoad` to fire on texture updates.
> 
> The fix itself is clean and minimal. LGTM on the core change.

@travisbreaks Thank you so much for the code review! Sorry for the late reply.

MichaelangJason:
Hi @DennisSmolek !

Just following up on this PR whenever you have a chance to take a look. I’d really appreciate any feedback on whether these changes address the issue correctly, and specifically whether ⁠`mappedTextures` should be included in the ⁠`useLayoutEffect` dependency lists.

I noticed the project seems to be focused on v11, so I understand if this PR is lower priority at the moment. I took a quick look at the v11 code, and it seems these changes could likely be applied there as well without modification. If that would be more useful, I’d be happy to open a separate PR for v11 or retarget this one to it.

Since this is my first contribution to an open-source project, I’d also appreciate any guidance on the best next step. 

Thanks for your time!

PR #2699, opened 2026-03-03
docs: add Storybook stories for Backdrop, Edges, and GradientTexture

travisbreaks:
## Summary

Adds Storybook stories for three components that were missing them, as tracked in #2683.

### Stories added

**`Backdrop.stories.tsx`** (`Staging/Backdrop`)
- Shows a curved studio backdrop receiving shadows, with a sphere mesh casting onto it
- Controls: `floor`, `segments` args exposed in Storybook
- Uses `ContactShadows` for ambient occlusion, `OrbitControls` for interaction

**`Edges.stories.tsx`** (`Abstractions/Edges`)  
- Wraps `Edges` as a child of a `torusKnotGeometry` mesh to show edge detection on a complex shape
- Controls: `threshold`, `color`, `lineWidth` args exposed
- `threshold` at 15° shows the edges clearly without noise

**`GradientTexture.stories.tsx`** (`Misc/GradientTexture`)
- Two stories: Linear (3-stop warm-to-cool gradient) and Radial (blue-to-purple)
- Renders on a plane with `meshBasicMaterial`
- Both `GradientType.Linear` and `GradientType.Radial` demonstrated

Closes #2683

Issue #2683, opened 2026-02-11
[v11] Audit components for stories


DennisSmolek:
Most components have stories correctly, but some are internal components or no one ever got around to it.

A script I wrote added test files to every folder in a early version of the alpha so one way is to see if a folder has a a test file still (delete once the story is made)

wbrunovieira:
Opened #2749 adding a Storybook story for `GradientTexture` (Linear + Radial) — one of the abstractions that didn't have an entry yet. Happy to pick off a few more from the list if that's helpful.

DennisSmolek:
This is now derived rather than guessed at. `yarn audit:components` writes `component-status.json`, and `node scripts/audit-components.js --summary` prints:

```
143 components
  stories     96 / 143
  test files  55 / 143
  REAL tests   0 / 143   (the rest assert nothing)
```

On your heuristic — "a folder still having a test file means nobody got around to the story" — it turned out to be the reverse of a useful signal, because **all 55 of those test files are the placeholders your script generated and none of them assert anything.** They exist and, separately, they do not even run: no executed Vitest project matches `src/**/*.test.ts`. That is #2802.

The scoping rule that came out of this, now applied to #2801: **a story for a component that was never really converted is not a story task — it is part of that component's conversion.** Writing one produces a red story and no information we did not already have. So #2801 covers only components whose WebGPU implementation is finished, and story work for the rest has been moved into the per-component conversion issues (#2811, #2663, #2665, #2660, #2588, #2654, #2659, #2661, #2662, #2664, #2786) as an acceptance criterion.

What is left for *this* issue is the same question on the `core`/`legacy`/`external` side: 47 of 143 components have no story, and unlike the WebGPU ones most of those are shipping and presumed working. `--summary` lists them.

PR #2700, opened 2026-03-06
docs: add Storybook stories for Bounds, Clone, and Mask

travisbreaks:
## Summary

Follow-up to #2699 (Backdrop, Edges, GradientTexture). Adds Storybook stories for 3 more high-value visual components, continuing the effort in #2683.

### New stories

- **Bounds** (`Staging/Bounds`) — Auto-fits camera to enclose children. Two variants: Default (with args for fit/clip/observe/margin) and Margin (demonstrates zoomed-out view).
- **Clone** (`Abstractions/Clone`) — Creates copies of an existing mesh at different positions. Two variants: Default and Deep + Inject (each clone gets its own material/color).
- **Mask** (`Portals/Mask`) — Stencil buffer masking that reveals content only through a mask shape. Two variants: Default and Inverse (content visible everywhere except through the mask).

All stories follow the existing conventions: `Setup` decorator, `Meta`/`StoryObj` types, `satisfies` pattern, imports from `../../src`.

References #2683.

Co-authored-by: Egger <egger@horny-toad.com>

### DIFF
--- a/.storybook/stories/Bounds.stories.tsx
+++ b/.storybook/stories/Bounds.stories.tsx
@@ -1,0 +1,77 @@
+import * as React from 'react'
+import * as THREE from 'three'
+import { Meta, StoryObj } from '@storybook/react-vite'
+
+import { Setup } from '../Setup'
+
+import { Bounds, Box, Sphere, Torus } from '../../src'
+
+export default {
+  title: 'Staging/Bounds',
+  component: Bounds,
+  decorators: [
+    (Story) => (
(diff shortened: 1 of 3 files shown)


travisbreaks:
Friendly ping — just checking if this is on the review radar. Happy to adjust if the approach needs changes. Thanks for maintaining such a great ecosystem!

travisbreaks:
Closing this out. The repo looks focused on v11 internal work right now, and these Storybook stories aren't a priority. Happy to revisit if coverage becomes a focus later. No point adding to the open PR backlog.

PR #2701, opened 2026-03-11
feat(PivotControls): allow disabling elements of pivotControl per axis

romankoho:
### Why

When using the PivotControl with an orthographic camera certain elements of the control can collapse / project on the same plane.

In my screenshot below you see how the pink arch of the AxisRotator collapses with the arrow of the AxisArrow. The user tries to move the object along the axis but instead he/she interacts with the AxisRotator and rotates the element unintentionally.

In this case I would want to be able to control the visible elements per axis

image

With the solution implemented I would be able to hide the arch for the x and z axes if the camera forward vector is (almost) perfectly aligned with the world-up vector.

image

### What

**1. The disable props of the PivotControls now either accept a boolean to disable the element for all axes or an array of three booleans to turn on/off the element per axis.**

```
  disableAxes?: boolean | [boolean, boolean, boolean]
  disableSliders?: boolean | [boolean, boolean, boolean]
  disableRotations?: boolean | [boolean, boolean, boolean]
  disableScaling?: boolean | [boolean, boolean, boolean]
```

the flag is then considered when rendering

```jsx
    const disableAxesFinal = Array.isArray(disableAxes) ? disableAxes : [disableAxes, disableAxes, disableAxes]

    return (
          {!disableAxesFinal[0] && activeAxes[0] && <AxisArrow axis={0} direction={xDir} />}
          {!disableAxesFinal[1] && activeAxes[1] && <AxisArrow axis={1} direction={yDir} />}
          {!disableAxesFinal[2] && activeAxes[2] && <AxisArrow axis={2} direction={zDir} />}
    )
```

**2. fix typo in PivotControls index.tsx (noticable to noticeable)**

### Checklist

- [x] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/docs/gizmos/pivot-controls.mdx?plain=1))
- [x] Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/PivotControls.stories.tsx))
- [x] Ready to be merged


travisbreaks (changes requested):
Good motivation here. The orthographic camera overlap problem is real, and per-axis control over disable flags is a reasonable extension of the existing API. A few things to address:

## Bug: `disableScalingFinal` ternary is inverted

The normalization for `disableScaling` has its branches flipped compared to the other three:

```ts
const disableScalingFinal = Array.isArray(disableScaling)
  ? [disableScaling, disableScaling, disableScaling]  // array case wraps the whole array again
  : disableScaling                                      // boolean case passes raw boolean
```

Every other prop does `Array.isArray(x) ? x : [x, x, x]`. This one does the opposite. Passing `disableScaling={[true, false, true]}` would produce `[[true,false,true], [true,false,true], [true,false,true]]`, which is always truthy. Passing `disableScaling={true}` would produce `true`, and `true[0]` is `undefined`, which is falsy, so nothing gets disabled. Both paths are broken.

## Semantic concern: slider/rotator index mapping

For `disableAxes` and `disableScaling`, the tuple index maps directly to the axis: index 0 = X, 1 = Y, 2 = Z. Straightforward.

But for `disableSliders` and `disableRotations`, the tuple index maps to the **position in the render list**, not the axis. Looking at the mapping:

- `disableSlidersFinal[0]` controls `<PlaneSlider axis={2}>` (XY plane, normal is Z)
- `disableSlidersFinal[1]` controls `<PlaneSlider axis={1}>` (XZ plane, normal is Y)
- `disableSlidersFinal[2]` controls `<PlaneSlider axis={0}>` (YZ plane, normal is X)

Same pattern for rotators. So `disableRotations={[true, false, false]}` disables rotation around Z, not around X. A consumer would reasonably expect index 0 to mean "the X-related thing," especially since that is how `activeAxes`, `disableAxes`, and `disableScaling` all work.

I'd suggest indexing by the `axis` prop value instead (the normal axis), so the semantics are consistent: index 0 = X, 1 = Y, 2 = Z across all four props. That would mean:

```tsx
{!disableSlidersFinal[2] && activeAxes[0] && activeAxes[1] && <PlaneSlider axis={2} dir1={xDir} dir2={yDir} />}
{!disableSlidersFinal[1] && activeAxes[0] && activeAxes[2] && <PlaneSlider axis={1} dir1={zDir} dir2={xDir} />}
{!disableSlidersFinal[0] && activeAxes[2] && activeAxes[1] && <PlaneSlider axis={0} dir1={yDir} dir2={zDir} />}
```

(Same reindexing for `disableRotationsFinal`.)

Worth documenting what the index means in the docstring too, since "disable rotation 0" meaning "rotation around X" (which visually is the YZ ring) isn't immediately obvious.

## Minor

The typo fix (`noticable` to `noticeable`) is a nice catch, but consider splitting it into its own commit or noting it in the PR description so it doesn't get lost in review noise.

Storybook story and docs update look good. The union type `boolean | [boolean, boolean, boolean]` is clean and backward-compatible.

romankoho:
thanks @travisbreaks for your review and sorry for being heedless about the raised issues - I've fixed them and have mentioned the fix of the typo on the PR

DennisSmolek:
> Good motivation here. The orthographic camera overlap problem is real, and per-axis control over disable flags is a reasonable extension of the existing API. A few things to address:

This is a very thorough review, did you use a tool for this or write it out manually?

PR #2703, opened 2026-03-11
fix(Html): update occlusion when occlude prop changes

### LINKED ISSUE #2702: Html does not update visibility when occlude prop changes
## Description

When using `<Html occlude={...}>`, changing the `occlude` prop at runtime does not immediately update the visibility of the Html element.

The visibility only updates after the camera moves or zoom changes.

Calling `invalidate()` does not help because the occlusion logic inside `useFrame` is only executed when certain camera or position changes occur.

## Reproduction

Example usage:

```tsx
const refs = // some hook 
const enableOcclusion =  // some hook

// we expect visual changes after refs changes or enableOcclusion setting changes
<Html occlude={enableOcclusion ? refs : undfined} sprite>
  ...
</Html>
```

When toggling `enableOcclusion`:

* the `occlude` prop changes
* the component re-renders
* but the visibility state does not update

Only after moving the camera slightly does the occlusion update.

## Root Cause

Inside `Html.tsx`, the occlusion visibility check is gated behind this condition in `useFrame`:

```ts
if (
  transform ||
  Math.abs(oldZoom.current - camera.zoom) > eps ||
  Math.abs(oldPosition.current[0] - vec[0]) > eps ||
  Math.abs(oldPosition.current[1] - vec[1]) > eps
)
```

Because of this, occlusion is only recalculated when:

* camera zoom changes
* projected screen position changes
* `transform` is enabled

Changes to the `occlude` prop itself do not trigger a recalculation.

## Proposed Fix

Track changes to `occlude` and include them in the recalculation condition.

Example minimal patch:

```ts
const prevOcclude = React.useRef<typeof occlude>(undefined)

useFrame(() => {
  const occludeChanged = prevOcclude.current !== occlude
  prevOcclude.current = occlude

  if (
    transform ||
    occludeChanged ||
    Math.abs(oldZoom.current - camera.zoom) > eps ||
    Math.abs(oldPosition.current[0] - vec[0]) > eps ||
    Math.abs(oldPosition.current[1] - vec[1]) > eps
  ) {
    // existing occlusion logic
  }
})
```

This ensures visibility is recalculated immediately when `occlude` changes.

I have opened a PR for the proposed change.

chris-xinhai-li:
fixes #2702 

### Problem

Changing the `occlude` prop dynamically does not update Html visibility
until the camera moves.

This happens because the occlusion logic inside `useFrame` only runs
when camera zoom or screen position changes.

### Fix

Track changes to the `occlude` prop and force recalculation when it changes.

### Result

Occlusion updates immediately when the prop changes without requiring camera movement.

### Checklist

- [  ] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/docs/misc/example.mdx?plain=1))
- [  ] Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/Example.stories.tsx))
- [x] Ready to be merged

### DIFF
--- a/src/web/Html.tsx
+++ b/src/web/Html.tsx
@@ -280,6 +280,7 @@ export const Html: ForwardRefComponent<HtmlProps, HTMLDivElement> = /* @__PURE__
     })
 
     const visible = React.useRef(true)
+    const prevOcclude = React.useRef<typeof occlude>(undefined);
 
     useFrame((gl) => {
       if (group.current) {
@@ -287,8 +288,12 @@ export const Html: ForwardRefComponent<HtmlProps, HTMLDivElement> = /* @__PURE__
         group.current.updateWorldMatrix(true, false)
         const vec = transform ? oldPosition.current : calculatePosition(group.current, camera, size)
 
+        const occludeChanged = prevOcclude.current !== occlude
+        prevOcclude.current = occlude
+
         if (
           transform ||
+          occludeChanged ||
           Math.abs(oldZoom.current - camera.zoom) > eps ||
           Math.abs(oldPosition.current[0] - vec[0]) > eps ||
           Math.abs(oldPosition.current[1] - vec[1]) > eps


travisbreaks (commented):
Good catch on the root cause. The `useFrame` gate skipping recalc when only `occlude` changes is a real bug, and the ref-based tracking pattern is the right approach for `useFrame` callbacks (no deps array to work with).

One concern with the reference equality check: when `occlude` is passed as an inline array (which is the common pattern from the issue), e.g.:

```tsx
<Html occlude={enableOcclusion ? refs : undefined}>
```

...where `refs` is `[meshRef1, meshRef2]` created inline or via a hook that returns a new array each render, `prevOcclude.current !== occlude` will be `true` on every single frame. That bypasses the entire early-exit optimization and forces full position + visibility recalculation every frame, even when nothing actually changed.

For primitive values (`true`, `"blending"`, `"raycast"`, `undefined`), reference equality works fine. But array occluders need something like a length + identity check, or serializing to a stable key. A simple approach:

```ts
function occludeChanged(prev: typeof occlude, next: typeof occlude): boolean {
  if (prev === next) return false
  if (Array.isArray(prev) && Array.isArray(next)) {
    return prev.length !== next.length || prev.some((ref, i) => ref !== next[i])
  }
  return true
}
```

This keeps the fix correct for the toggle case (the actual bug) while avoiding a per-frame penalty for stable array occluders that happen to get a new array reference each render.

Minor: initializing `prevOcclude` with `occlude` means the very first frame won't detect a "change" from the initial prop. That's fine for this use case since the existing occlusion logic handles the initial state already.

chris-xinhai-li:
Hi @travisbreaks:

My thinking when proposing the simpler reference check was based on the typical React convention of treating props as immutable. In most React patterns (useState, useMemo, useEffect, etc.), reference equality is used as the signal for change, and developers usually ensure immutability / stability when passing arrays or objects.

(e.g., via useMemo, in my projects I normally use a custome state management bases on immer.js) 

From that perspective, assuming a immutable reference for occlude felt consistent with common React practices, and it keeps the patch minimal. If users generate a new array every render, they can usually handle it at a higher level according to their application level state management / change detection mechanisms.

So I agree with the concern (which is not specific to this Html component),  it mostly comes down to where this should be handled. Personally I lean toward letting the upper-level logic deal with it, since that’s the usual pattern in React.

I totally agree with this one:
"Minor: initializing prevOcclude with occlude means the very first frame won't detect a "change" from the initial prop. That's fine for this use case since the existing occlusion logic handles the initial state already."

I will fix this, thanks a lot for pointing out. 

Cheers

DennisSmolek:
@chris-xinhai-li can you check this against the v11 version? https://github.com/pmndrs/drei/tree/v11-working/src/core/UI/Html

There was some changes and has the new useFrame logic.

chris-xinhai-li:
Hi @DennisSmolek, I checked the v11 branch and reviewed the recent changes to the Html component. It looks like:

1. The same issue still exists
2. The relevant logic hasn’t changed
3. This PR should apply to v11 without modification

Is there anything I should do for v11? Should I open a separate PR targeting the v11 branch?

DennisSmolek:
> Hi @DennisSmolek, I checked the v11 branch and reviewed the recent changes to the Html component. It looks like:
> 
> 1. The same issue still exists
> 2. The relevant logic hasn’t changed
> 3. This PR should apply to v11 without modification
> 
> Is there anything I should do for v11? Should I open a separate PR targeting the v11 branch?

I would move this to the v11 branch. I can merge it faster and we are likely merging those over faster than smaller v10 changes.

PR #2706, opened 2026-03-16
allow disabling controls per axis

romankoho:
### Why

When using the PivotControl with an orthographic camera certain elements of the control can collapse / project on the same plane.

In my screenshot below you see how the pink arch of the AxisRotator collapses with the arrow of the AxisArrow. The user tries to move the object along the axis but instead he/she interacts with the AxisRotator and rotates the element unintentionally.

In this case I would want to be able to control the visible elements per axis

image

With the solution implemented I would be able to hide the arch for the x and z axes if the camera forward vector is (almost) perfectly aligned with the world-up vector.

image

### What

**1. The disable props of the PivotControls now either accept a boolean to disable the element for all axes or an array of three booleans to turn on/off the element per axis.**

```
  disableAxes?: boolean | [boolean, boolean, boolean]
  disableSliders?: boolean | [boolean, boolean, boolean]
  disableRotations?: boolean | [boolean, boolean, boolean]
  disableScaling?: boolean | [boolean, boolean, boolean]
```

the flag is then considered when rendering

```jsx
    const disableAxesFinal = Array.isArray(disableAxes) ? disableAxes : [disableAxes, disableAxes, disableAxes]

    return (
          {!disableAxesFinal[0] && activeAxes[0] && <AxisArrow axis={0} direction={xDir} />}
          {!disableAxesFinal[1] && activeAxes[1] && <AxisArrow axis={1} direction={yDir} />}
          {!disableAxesFinal[2] && activeAxes[2] && <AxisArrow axis={2} direction={zDir} />}
    )
```

**2. fix typo in PivotControls index.tsx (noticable to noticeable)**

### Checklist

- [x] Documentation updated ([example](https://github.com/pmndrs/drei/blob/master/docs/gizmos/pivot-controls.mdx?plain=1))
- [x] Storybook entry added ([example](https://github.com/pmndrs/drei/blob/master/.storybook/stories/PivotControls.stories.tsx))
- [x] Ready to be merged

I had PR #2701 open but messed up the git history (a noob was doing noob stuff)

### DIFF
--- a/docs/gizmos/pivot-controls.mdx
+++ b/docs/gizmos/pivot-controls.mdx
@@ -35,14 +35,14 @@ type PivotControlsProps = {
   autoTransform?: boolean
   /** Allows you to switch individual axes off */
   activeAxes?: [boolean, boolean, boolean]
-  /** Allows you to disable translation via axes arrows */
-  disableAxes?: boolean
-  /** Allows you to disable translation via axes planes */
-  disableSliders?: boolean
-  /** Allows you to disable rotation */
-  disableRotations?: boolean
-  /** Allows you to disable scaling */
-  disableScaling?: boolean
+  /** Allows you to disable translation via axes arrows either for all axes or for individual ones */
+  disableAxes?: boolean | [boolean, boolean, boolean]
+  /** Allows you to disable translation via axes planes either for all axes or for individual ones */
+  disableSliders?: boolean | [boolean, boolean, boolean]
+  /** Allows you to disable rotation either for all axes or for individual ones */
+  disableRotations?: boolean | [boolean, boolean, boolean]
+  /** Allows you to disable scaling either for all axes or for individual ones */
+  disableScaling?: boolean | [boolean, boolean, boolean]
   /** RGB colors */
   axisColors?: [string | number, string | number, string | number]
   /** Color of the hovered item */
@@ -84,4 +84,6 @@ return (
     matrix={matrix}
     autoTransform={false}
     onDrag={({ matrix: matrix_ }) => matrix.copy(matrix_)}
+  />
+)
 ```
--- a/src/core/Gizmos/PivotControls/index.tsx
+++ b/src/core/Gizmos/PivotControls/index.tsx
@@ -61,11 +61,11 @@ export type PivotControlsProps = {
   /** Allows you to switch individual axes off */
   activeAxes?: [boolean, boolean, boolean]
 
-  /** Allows you to switch individual transformations off */
-  disableAxes?: boolean
-  disableSliders?: boolean
-  disableRotations?: boolean
-  disableScaling?: boolean
+  /** Allows you to switch individual transformations off. Either across all axes, or individually. */
+  disableAxes?: boolean | [boolean, boolean, boolean]
+  disableSliders?: boolean | [boolean, boolean, boolean]
+  disableRotations?: boolean | [boolean, boolean, boolean]
+  disableScaling?: boolean | [boolean, boolean, boolean]
 
   /** Limits */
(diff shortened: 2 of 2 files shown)


DennisSmolek:
Why did you close the last PR for this? It removes the review/discussion threads and with bots/security being issues for libs it raises hairs.

DennisSmolek:
oh I see the git history note...

src/web/pivotControls/index.tsx
```diff
@@ -258,37 +258,49 @@ export const PivotControls: ForwardRefComponent<PivotControlsProps, THREE.Group>
 
     React.useImperativeHandle(fRef, () => ref.current, [])
 
+    const disableAxesFinal = Array.isArray(disableAxes) ? disableAxes : [disableAxes, disableAxes, disableAxes]
```
DennisSmolek: I would avoid using "final" in place of slightly more descriptive. something like `disableTranslationAxis` so it reads:
`!disableTranslationAxis[0]`

src/web/pivotControls/index.tsx
```diff
@@ -258,37 +258,49 @@ export const PivotControls: ForwardRefComponent<PivotControlsProps, THREE.Group>
...
+    const disableSlidersFinal = Array.isArray(disableSliders)
+      ? disableSliders
+      : [disableSliders, disableSliders, disableSliders]
+    const disableRotationsFinal = Array.isArray(disableRotations)
+      ? disableRotations
+      : [disableRotations, disableRotations, disableRotations]
+    const disableScalingFinal = Array.isArray(disableScaling)
```
DennisSmolek: Same here, slight improvement would be `disableScalingAxis[]` or `disableScalingDirection`

DennisSmolek (commented):
All in all this seems good. 

It doesnt solve your ultimate problem though and requires users to straight disable them instead if the rotation to be off when the camera at the extreme angle. 

These props are useful so I support merging it, however I made some semantic recommendations. 

I would also be interested in an expanded version that takes the camera normal and potentially disables the rotation arcs which solves your actual problem. 

The next thing is we are avoiding v10 merges unless fixing breaking/major things, so I'm not sure this would go in soon on v10, however, if you PR this against v11 with those minor changes I will merge it in immediately.

travisbreaks (approved):
Both issues from my review on #2701 are resolved:

1. **Inverted ternary**: `disableScalingFinal` now follows the same `Array.isArray(x) ? x : [x, x, x]` pattern as the other three props. Good.

2. **Axis index mapping**: Sliders and rotators now index by the `axis` prop value (the normal axis) instead of render order. `disableSlidersFinal[2]` guards `axis={2}`, `[1]` guards `axis={1}`, `[0]` guards `axis={0}`. Consistent with `disableAxes` and `disableScaling`. This is exactly what I suggested.

Clean diff overall. The typo fix and storybook entry are nice touches. LGTM.

romankoho:
thanks @travisbreaks and @DennisSmolek for your review!
I've changed the target branch to [v11-working](https://github.com/pmndrs/drei/tree/v11-working)

@travisbreaks I'm also curios about your thorough review. Did you you write it yourself or use a tool?
@DennisSmolek yes I'm aware that the PR doesn't solve my initial issue but at least it enables me to solve it on my end. I'd be open to actually fix it within the PivotControls component in a future PR.

romankoho:
hey @DennisSmolek any more changes needed or is it good to merge?

romankoho:
you guys can also check out #2708
As previously discussed this branch would directly hide the control elements depending on the perspective

DennisSmolek:
> hey @DennisSmolek any more changes needed or is it good to merge?

Ill spend time this weekend on drei and likely merge this to v11. Sorry for the delay. Starting a new job ate up some time. Will be catching up though

PR #2714, opened 2026-03-28
fix(useVideoTexture): clean up video element and cache on unmount

### LINKED ISSUE #2712: useVideoTexture: cached video element not cleaned up on unmount
- `three` version: ^0.159.0
- `@react-three/fiber` version: ^9.0.0
- `@react-three/drei` version: latest (confirmed by reading current `master` source)
- `node` version: N/A (reproducible across versions)
- `npm` (or `yarn`) version: N/A

### Problem description:

`useVideoTexture` uses `suspend-react` to cache the `<video>` element creation, keyed by `[srcOrSrcObject]`. When a component using `useVideoTexture` unmounts (e.g. by changing a `key` prop), the cached video element **persists and continues playing** in the background.

The current cleanup effect only destroys the HLS instance but does not pause or dispose the underlying video element, nor does it evict the `suspend-react` cache entry:

```ts
useEffect(() => {
  start && texture.image.play()

  return () => {
    if (hlsRef.current) {
      hlsRef.current.destroy()
      hlsRef.current = null
    }
    // no video.pause(), no video.src = "", no cache eviction
  }
}, [texture, start])
```

The `VideoTexture` component does call `texture.dispose()` on unmount, but that only disposes the Three.js texture object - the HTML `<video>` element that `suspend-react` keeps in its cache remains alive and playing.

**Expected behavior:** when a component using `useVideoTexture` unmounts, the underlying `<video>` element should be paused and cleaned up, and the `suspend-react` cache entry should be evicted.

**Steps to reproduce:**

1. Render a component that uses `useVideoTexture` with a video source
2. Switch to a different video by changing a `key` prop (causing unmount + remount with a new source)
3. Unmute audio - you can hear **both** videos playing simultaneously
4. Switch back - the timestamp of the previous video is preserved (it never stopped)

### Relevant code:

Minimal reproduction repo: https://github.com/mohamedtahaguelzim/drei-video-bug

- Clone, `npm install`, `npm run dev`
- Use the buttons to switch videos and unmute
- Open the console to observe mount/unmount logs - the cached video persists despite unmounting

### Suggested solution:

The cleanup effect in `useVideoTexture` should:

1. Pause the video element (`video.pause()`)
2. Clear the video source (`video.removeAttribute('src'); video.load()`)
3. Dispose the texture
4. Evict the `suspend-react` cache entry using `clear()` with the matching key (`[srcOrSrcObject]`)

This follows the same cleanup pattern already used by `WebcamVideoTexture` and `ScreenVideoTexture`.

Fix submitted in #2714.

mohamedtahaguelzim:
### Why

Resolves #2712

`useVideoTexture` uses `suspend-react` to cache the `<video>` element, but never cleans it up on unmount. The cached video persists and continues playing in the background, leaking resources.

### What

Added proper cleanup to the `useEffect` in `useVideoTexture`:

1. Pause the video element
2. Clear its source attribute and call `load()` to release network resources
3. Dispose the Three.js texture
4. Evict the `suspend-react` cache entry via `clear([srcOrSrcObject])`

This follows the same pattern already used by `WebcamVideoTexture` and `ScreenVideoTexture` in this codebase.

Reproduction repo: https://github.com/mohamedtahaguelzim/drei-video-bug

### Checklist

- [ ] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/src/core/VideoTexture.tsx
+++ b/src/core/VideoTexture.tsx
@@ -98,10 +98,14 @@ export function useVideoTexture(
   useEffect(() => {
     start && texture.image.play()
 
+    // NOTE: This cleanup assumes a single consumer per srcOrSrcObject. If multiple
+    // components share the same source, the first to unmount will release the video
+    // and evict the cache, leaving others with a dead texture.
     return () => {
       const video = texture.image as HTMLVideoElement
       video.pause()
       video.removeAttribute('src')
+      video.srcObject = null
       video.load()
 
       if (hlsRef.current) {
@@ -132,10 +136,6 @@ export type VideoTextureProps = {
 export const VideoTexture = /* @__PURE__ */ forwardRef<VideoTexture, VideoTextureProps>(
   ({ children, src, ...config }, fref) => {
     const texture = useVideoTexture(src, config)
-
-    useEffect(() => {
-      return () => void texture.dispose()
-    }, [texture])
 
     useImperativeHandle(fref, () => texture, [texture]) // expose texture through ref
 
--- a/src/core/VideoTexture.tsx
+++ b/src/core/VideoTexture.tsx
@@ -3,7 +3,7 @@ import * as React from 'react'
 import * as THREE from 'three'
 import { forwardRef, useEffect, useImperativeHandle, useRef } from 'react'
 import { useThree } from '@react-three/fiber'
-import { suspend } from 'suspend-react'
+import { suspend, clear } from 'suspend-react'
 import { type default as Hls, Events } from 'hls.js'
 
 const IS_BROWSER = /* @__PURE__ */ (() =>
@@ -99,12 +99,20 @@ export function useVideoTexture(
     start && texture.image.play()
 
     return () => {
+      const video = texture.image as HTMLVideoElement
+      video.pause()
+      video.removeAttribute('src')
+      video.load()
+
       if (hlsRef.current) {
         hlsRef.current.destroy()
         hlsRef.current = null
       }
+
+      texture.dispose()
+      clear([srcOrSrcObject])
     }
-  }, [texture, start])
+  }, [texture, start, srcOrSrcObject])
 
   return texture
 }


travisbreaks (commented):
Nice fix. The root cause analysis is solid, and the cleanup pattern aligns with what `WebcamVideoTexture` and `ScreenVideoTexture` already do. A few observations:

**The fix looks correct for the single-consumer case.** The cleanup order (pause, remove src, load, destroy HLS, dispose texture, clear cache) is the right sequence. Removing the `src` attribute and calling `load()` releases any ongoing network fetch, which matters for large video files. Good.

**Cache key alignment is correct.** `suspend([srcOrSrcObject])` and `clear([srcOrSrcObject])` use the same key shape, so the eviction will hit the right entry.

**One concern: shared cache entries across multiple consumers.** If two components both call `useVideoTexture("same-video.mp4")`, they share the same suspend-react cache entry (keyed by `[srcOrSrcObject]`). When the first component unmounts, the cleanup will pause the video, strip the src, dispose the texture, and evict the cache. The second component now holds a reference to a dead texture with a stopped, src-less video element. This is probably an uncommon pattern, but worth noting. A ref-counting approach (only clear/dispose when the last consumer unmounts) would handle this, though that might be out of scope for this PR. Worth a comment in the code or docs at minimum?

**Double dispose when using the `VideoTexture` component.** The `VideoTexture` component wrapper (same file, around line 115) has its own cleanup:

```tsx
useEffect(() => {
  return () => void texture.dispose()
}, [texture])
```

With this PR, `useVideoTexture` now also calls `texture.dispose()` in its cleanup. When someone uses the `<VideoTexture>` component (which calls `useVideoTexture` internally), dispose fires twice on unmount. Three.js handles this gracefully (the dispose event just fires again), so it is not a bug, but it is redundant. Might be worth removing the dispose from the `VideoTexture` component wrapper since the hook now owns that responsibility.

**Adding `srcOrSrcObject` to the dependency array is the right call.** Without it, swapping the video source would not trigger cleanup of the old video element.

**Minor edge case: unmount during initial load.** If the component unmounts while still suspended (before the `unsuspend` event fires), the useEffect cleanup never runs because the component never fully mounted. The video element created inside the Promise callback and the pending cache entry would both leak. This is a pre-existing issue, not introduced by this PR, but worth a follow-up issue if someone wants to pursue it.

Overall this is a clean, well-scoped fix. The shared-consumer edge case and the double-dispose in the component wrapper are the main items I would flag.

mohamedtahaguelzim:
Thanks for the review. Pushed a follow up commit addressing the feedback.
I removed the `texture.dispose()` from the `VideoTexture`. I think it's the right move because one may want to use this hook outside the context of `VideoTexture`.
I added an inline comment regarding the single-consumer assumption. Didn't touch larger docs, nor ref-counting. I also cleared `video.srcObject = null`, so sources set via `srcObject` (e.g. from `WebcamVideoTexture`) are properly released. Regarding unmount-during-suspension, I can open a follow-up issue for that if you'd like.

travisbreaks (approved):
Follow-up commit looks good. The double-dispose is properly resolved by centralizing cleanup in the hook, and the `srcObject = null` addition is the right call for MediaStream cleanup.

Two items for consideration (neither blocks approval):

1. **Shared cache**: The inline comment documenting the single-consumer assumption is honest, but consumers won't see it. Consider a `console.warn` in dev mode if `clear()` is called on an entry with active consumers, or a note in the JSDoc/docs. Video walls (multiple meshes, same source) are a real use case that will silently break.

2. **Unmount during suspension**: You mentioned filing a follow-up issue for the case where unmount happens while `suspend-react` is still suspended. Worth tracking before this merges so it doesn't get lost.

Overall this is a solid fix for the dominant single-consumer case. The cleanup sequence matches `WebcamVideoTexture` and `ScreenVideoTexture` patterns in the codebase.

Issue #1588, opened 2023-07-26
TrackballControls don't work with frameloop="demand"


romarybi:
The title says it all. TrackballControls only work with frameloop set to "always". No such issue found in similar OrbitControls. 
Sandbox: [https://codesandbox.io/s/dreamy-frog-8tvnkn?file=/src/App.js](https://codesandbox.io/s/dreamy-frog-8tvnkn?file=/src/App.js)

EliCDavis-Vinci:
Still a bug...

Issue #708, opened 2021-12-29
TrackballControls do not work with frameloop="demand"


nermolov:
- `three` version: 0.136.0
- `@react-three/fiber` version: 7.0.24
- `@react-three/drei` version: 8.3.1

### Problem description:

Unlike OrbitControls, TrackballControls doesn't seem to invalidate the frame and thus doesn't trigger rendering with `frameloop="demand"`

### Relevant code:

https://codesandbox.io/s/agitated-pascal-nvw2m?file=/src/App.js

### Suggested solution:

I tried using the experimental CameraControls.TrackballControls as #171 suggested that solves this issue, but couldn't get that working either.

surdu:
This is [documented in Three Fiber](https://docs.pmnd.rs/react-three-fiber/advanced/scaling-performance#triggering-manual-frames)

PR #2746, opened 2026-06-11
fix(TrackballControls): invalidate on user input with demand frameloops

### LINKED ISSUE #2745: TrackballControls do not work with frameloop="demand"
three version: 0.176.0
@react-three/fiber version: 9.1.2
@react-three/drei version: 10.0.8

## Problem description:
Unlike OrbitControls, TrackballControls doesn't seem to invalidate the frame and thus doesn't trigger rendering with `frameloop="demand"`

## Relevant code:
https://codesandbox.io/s/agitated-pascal-nvw2m?file=/src/App.js

## Suggested solution:
I tried using the experimental CameraControls.TrackballControls as https://github.com/pmndrs/drei/issues/171 suggested that solves this issue, but couldn't get that working either.

## Extra Details

This has been an on going issue for while that's just been unaddressed. What's the route to putting this in a road map?

https://github.com/pmndrs/drei/issues/1588
https://github.com/pmndrs/drei/issues/708

DevTomUK:
Fixes #2745 (same problem as #1588 and #708)

TrackballControls is dead on arrival with `frameloop="demand"`. The three-stdlib impl only applies pointer input inside `update()`, which drei runs in `useFrame` — but in demand mode no frame ever gets requested, so `update()` never runs and nothing you do produces a render. OrbitControls doesn't have this issue because its impl updates straight from the pointer handlers.

Fix: invalidate on the controls' `start` event, and on pointermove/touchmove while interacting. Wheel is covered too since it fires `start` per event.

Tested in a demand-frameloop sandbox: before, dragging renders nothing and the camera never moves; after, drag and wheel both work and damping settles on its own.

### DIFF
--- a/src/core/TrackballControls.tsx
+++ b/src/core/TrackballControls.tsx
@@ -47,15 +47,30 @@ export const TrackballControls: ForwardRefComponent<TrackballControlsProps, Trac
           if (regress) performance.regress()
           if (onChange) onChange(e)
         }
-        controls.addEventListener('change', callback)
-        if (onStart) controls.addEventListener('start', onStart)
-        if (onEnd) controls.addEventListener('end', onEnd)
-        return () => {
-          if (onStart) controls.removeEventListener('start', onStart)
-          if (onEnd) controls.removeEventListener('end', onEnd)
-          controls.removeEventListener('change', callback)
+
+        const onInput = () => invalidate()
+        const startCallback = (e: THREE.Event) => {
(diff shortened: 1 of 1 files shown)


EliCDavis-Vinci:
How do we get this merged

DevTomUK:
@EliCDavis-Vinci It needs someone from Poimandres to approve it before it can be merged

PR #2763, opened 2026-08-08
fix(GizmoHelper): land tween exactly on the target axis

smeng9:
`rotateTowards` only brings q1 within the 0.01 rad tolerance of q2 and the animation then stops, so the camera keeps a small residual offset off the requested axis.

For the four side views that residual is harmless, but for the top/bottom views it sits exactly on the orbit pole, which is where OrbitControls re-derives its spherical coordinates from the camera position:

  theta = atan2(offset.x, offset.z)

With offset ~= (0, r, 0) the azimuth ends up determined entirely by that residual noise, which varies with wherever the camera happened to start. `makeSafe()` then clamps phi off the pole and `lookAt(target)` bakes the bogus azimuth in, so clicking the +Y/-Y gizmo produces a top-down view spun to a seemingly random angle (measured up to ~70 degrees).

Snap q1 onto q2 on the completion frame and push it through the same path the animation frames use, extracted as `applyTweenRotation`. The camera then lands exactly on the axis, theta resolves to 0 deterministically and the controls agree with the tween instead of overriding it. This also adds the missing `invalidate()` on the final frame, which matters under frameloop="demand".

Verified in a browser against Canvas + GizmoHelper + GizmoViewport + OrbitControls over 6 axes x 8 starting positions: every axis now lands on an identical orientation regardless of where the camera came from, and the side axes land exactly on-axis instead of ~0.4 degrees off. CameraControls keeps its own spherical state and never drifted, and stays correct.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/src/core/GizmoHelper.tsx
+++ b/src/core/GizmoHelper.tsx
@@ -104,12 +104,35 @@ export const GizmoHelper = ({
     [defaultControls, mainCamera, onTarget, invalidate]
   )
 
+  // Push the current tween rotation (q1) onto the main camera and the controls
+  const applyTweenRotation = React.useCallback(
+    (delta: number) => {
+      mainCamera.position.set(0, 0, 1).applyQuaternion(q1).multiplyScalar(radius.current).add(focusPoint.current)
+      mainCamera.up.set(0, 1, 0).applyQuaternion(q1).normalize()
+      mainCamera.quaternion.copy(q1)
+
+      if (isCameraControls(defaultControls))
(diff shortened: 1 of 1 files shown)


smeng9:
You can reproduce the effect by visiting https://drei.pmnd.rs/?path=/story/gizmos-gizmohelper--gizmo-helper-st-1 and click the top side of the gizmo

smeng9:
Also happing in https://drei.pmnd.rs/?path=/story/gizmos-gizmohelper--gizmo-helper-st-2 (with some chance) if you click between x and y in the gizmo

PR #2787, opened 2026-08-31
fix(ci): use semantic-release 25 so alpha publishing works

DennisSmolek:
The alpha.6 release job failed after #2785 merged, so `11.0.0-alpha.6` was never published. `alpha` is still on `11.0.0-alpha.5` on npm.

Two errors, one root cause:

```
ENONPMTOKEN No npm token specified.
Resource not accessible by integration (403, issues#create-an-issue)
```

#2760 fixed trusted publishing on `master` by doing **two** things:

1. moving the release job's permissions to job level, adding `issues`, `pull-requests` and `id-token`
2. bumping `semantic-release` **24 → 25**, which is the version that actually supports OIDC trusted publishing

The v11 line kept its own `release.yml` and its own `semantic-release@^24.2.0`, so it inherited neither. semantic-release 24 has no trusted-publishing support, so it falls back to looking for `NPM_TOKEN` — which no longer exists now that the org publishes via OIDC.

This PR applies both halves to the v11 workflow.

Note the content diff here is only these three files: everything else in `v11-working` is already on `alpha` by content from the #2785 squash.

Once merged, the push to `alpha` should publish `11.0.0-alpha.6`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DennisSmolek:
Superseded by #2788.

The conflict here was mechanical, not a real content clash. `v11-working`'s merge-base with `alpha` is `a095f7d5` (February) — the #2785 squash left `alpha` with no shared history — so git three-way merged six months of divergence on `package.json` and `yarn.lock`. The two branches actually differ by only three files.

#2788 is `alpha` plus the same cherry-picked commit, so it merges cleanly.

PR #2831, opened 2026-09-14
fix: use scene camera instead of gl.camera in Html occlusion

### LINKED ISSUE #2819: Html: useFrame((gl) => …) shadows the renderer — not a crash, a hazard
## `Html.tsx:455` reads `gl.camera`, which does not exist on either renderer

```ts
occlusionMeshRef.current.lookAt(gl.camera.position)
```

`gl` is the renderer. **Neither renderer has a `camera` property.** Checked three ways against three 0.185:

- no `this.camera =` or `get camera()` in `three/src/renderers/common/Renderer.js`
- none in `three/src/renderers/WebGLRenderer.js`
- `'camera' in WebGPURenderer.prototype` → `false`

So `gl.camera` is `undefined` and `.position` throws `TypeError: Cannot read properties of undefined`.

## Not a WebGPU issue

This fails on **WebGL too**. It was found during a WebGPU audit only because the sweep was looking for renderer members that differ between the two — this one exists on neither.

## The fix looks like a one-word typo

`camera` is already destructured in the same component, `Html.tsx:197`:

```ts
const { gl, camera, scene, size, raycaster, events, viewport } = useThree()
```

So `gl.camera.position` should almost certainly be `camera.position`. Worth confirming against the intent — the surrounding block is the transform-mode occlusion mesh being sized and oriented — rather than assuming.

## Why it has not been reported

The line sits in the `else` branch of the transform-mode occlusion path, reached only with `occlude` + `transform` and a non-custom occlusion geometry. Narrow enough that nobody has hit it, deterministic once they do.

No test covers it, and `Html` has no story exercising that combination.

a1stok:
### Why

`Html.tsx` reads `gl.camera.position` in the non-transform occlusion path. `gl` is the renderer -- neither `WebGLRenderer` nor `WebGPURenderer` has a `camera` property, so this throws `TypeError: Cannot read properties of undefined (reading 'position')`.

resolves #2819

### What

replaced `gl.camera.position` with `camera.position`, using the camera already destructured from `useThree()` at the top of the component.

### Checklist

- [x] Documentation updated -- not needed, no public api change
- [x] Storybook entry added -- not needed, bug fix with no visual change
- [x] Ready to be merged


a1stok:
@krispya can you check pls?

PR #2834, opened 2026-09-21
fix(e2e): pin the scaffolded apps to a react line fiber accepts

krispya:
## Problem

react 19.3.0 was published on 2026-09-09. `test/e2e/e2e.sh` scaffolds fresh Vite and Next apps, which now resolve react to 19.3.0, and the following `npm i $TGZ` fails with `ERESOLVE`: drei's `@react-three/fiber@^9` peer picks fiber 9.7.0, whose react peer is `>=19 <19.3`. Every PR opened since (e.g. #2833) fails all three `test (w/ three@...)` jobs before any drei code runs; master's last run predates the react release.

## Fix

Pin `react` and `react-dom` in the scaffolded app to `~19.2.0`, using the same jq helper pattern the script already has for three, in both the Vite and the Next scaffolds.

## Verification

Reproduced the Vite path locally: scaffold, apply both pins, `npm i`, then install `@react-three/drei@10.7.8` (same peer graph as the tarball). Resolves to react 19.2.8, fiber 9.7.0, drei 10.7.8 with no peer conflict.

Once fiber publishes a release that admits react 19.3 the pin can be raised or dropped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


krispya:
We will fix this in R3F instead.

